PR #19, opened 2015-11-20
Fix local examples path in readme

pascalberger:
Fixes the path to the local examples in the readme to the current path of the extension


daviwil:
Hi Pascal, thank you for fixing this!

PR #32, opened 2015-11-26
Add configurationAttributes to package.json

isidorn:
Fixes microsoft/vscode#546


daviwil:
Thank you so much for sending this Isidor!

PR #38, opened 2015-12-01
Update PowerShell.json

pcgeek86:
Added Splatting example to VS Code snippets

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

### DIFF
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -1,4 +1,15 @@
 {
+	"Splatting": {
+		"prefix": "splat",
+		"body": [
+			"$Params = @{", 
+			"\tModule = '*'",
+			"\tVerb = 'Get'",
+			"}",
+			"Get-Command"
+			],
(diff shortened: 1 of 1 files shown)


daviwil:
Thanks Trevor!  Interesting, I hadn't considered using snippets for instructive code examples.  Generally they're meant for actual blocks of code that a developer would want to expand quickly.  I'll accept your PR since the snippet is helpful but in the future we may need to find a different way to surface these examples.  Maybe an action in the command palette for "PowerShell Usage Examples" which expands into a searchable list of examples that can be inserted into the code?  Not very different than what the snippets do but a little more clear on the usage pattern.

Anyway, thanks again!

pcgeek86:
@daviwil Yeah, I agree there might be a better way to do this. If nothing else, it serves as a basic construct that a developer can use, much like the other snippet constructs. I sent a new PR, because I forgot to add @Params. Whoops.

daviwil:
No problem, merged it too :)

PR #41, opened 2015-12-03
Add "Expand Alias" command

dfinke:
Adds Expand Alias to the PowerShell extension


package.json
```diff
@@ -38,19 +38,39 @@
...
 		],
-		"commands": [
+		"commands": [			
+			{
+				"command": "PowerShell.ExpandAlias",
+				"title": "Expand Alias",
+				"category": "PowerShell"
+			},		
+			{
+				"command": "PowerShell.GetInstalledModule",
```
daviwil: Looks like these two commands got left in by accident.  Can you make another commit that removes them?

PR #51, opened 2015-12-18
package.json: use ${workspaceRoot}

isidorn:
At the moment we magically append the workspace root to some properties of the launch.json on the vscode side.
This will go away in the future since it is clunky. Thus we want to educate users to use the ${workspaceRoot} variable for all paths, since we will no longer support relative paths in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
 						"name": "PowerShell",
 						"type": "PowerShell",
 						"request": "launch",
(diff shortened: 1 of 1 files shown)


isidorn:
Also the package.json in the examples folder should be updated.

daviwil:
Makes sense, thanks for sending the PR!

PR #61, opened 2015-12-30
Enhance code snippets

mad-mike:
I've enhanced the snippet collection with variable definitions and tab stops. Snippets are not only a reference, but they should speed up the coding by preparing the boilerplate code.

### DIFF
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -13,14 +13,14 @@
 	"Class": {
 		"prefix": "class",
 		"body": [
-			"class MyClass {",
+			"class ${classname:MyClass} {",
 		    	"\t# Property: Holds name",
 		   		"\t[String] $Name;",
 		
 				"",
 			    "\t# Constructor: Creates a new MyClass object, with the specified name",
-			    "\tMyClass([String] $NewName) {",
-			        "\t\t# Set name for MyClass",
+			    "\t${classname:MyClass}([String] $NewName) {",
+			        "\t\t# Set name for ${classname:MyClass}",
 			        "\t\t$this.Name = $NewName;",
 			    "\t}",
 				"",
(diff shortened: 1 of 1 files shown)


rkeithhill:
I wonder why the diff on commit 89f6488 is bonkers?  The previous commit looks fine but 89f6488 acts as if everything changed.

mad-mike:
Hi @rkeithhill, try this link: https://github.com/PowerShell/vscode-powershell/pull/61/files?w=1
I must have made whitespace changes by mistake - maybe that'd be worth a feature request for the changes view in VS Code.

rkeithhill:
@mad-mike Very nice!!  I like these enhancements.  

BTW did you find a reference on the various features of snippet completion like `$${variable:i}` and `$1` (believe that is where the cursor is left after changing snippet fields, right?)

mad-mike:
Just search for "Snippet" in the command box in VS Code and pick PowerShell. This file comes with a short introduction.
```
{
/*
	 // Place your snippets for PowerShell here. Each snippet is defined under a snippet name and has a prefix, body and 
	 // description. The prefix is what is used to trigger the snippet and the body will be expanded and inserted. Possible variables are:
	 // $1, $2 for tab stops, ${id} and ${id:label} and ${1:label} for variables. Variables with the same id are connected.
	 // Example:
	 "Print to console": {
		"prefix": "log",
		"body": [
			"console.log('$1');",
			"$2"
		],
		"description": "Log output to console"
	}
*/
}
```

rkeithhill:
Thanks!  

BTW before @daviwil accepts this, would you mind adding a snippet for `try-catch`?  I use try/catch far more often than try/finally and even try/catch/finally.  You know, while you're in there, under the hood.  :-)

daviwil:
Great changes, thanks @mad-mike!  The whitespace changes are probably caused by line ending changes in that file.  I probably need to set up a .gitattributes file to solve this problem for everyone.

Would be great if you could add the try/catch snippet too but if you don't have time it's no problem.

mad-mike:
@rkeithhill, @daviwil That's true, try-catch should definitely be in the list.
I've added it with https://github.com/mad-mike/vscode-powershell/commit/db4745c2d98a86243018e587ba07967d67f81dfe.

daviwil:
Excellent, thanks a lot for your contribution @mad-mike!  I think users will enjoy these improvements.

Issue #58, opened 2015-12-27
Extension install path in docs needs to include version


rkeithhill:
Both the README.md and the contents of the VSCode Marketplace page for this extension (https://marketplace.visualstudio.com/items/ms-vscode.PowerShell) should include a placeholder for the version number e.g. `c:\Users\<yourusername>\.vscode\extensions\ms-vscode.PowerShell-<ext_version>\examples`.

daviwil:
Is this a new folder convention that they're introducing?  I haven't seen that before

rkeithhill:
That's what I see when I've installed later versions of the extension on both 0.10.5 and 0.10.6 e.g.:
```
C:\Users\hillr\.vscode\extensions\ms-vscode.PowerShell-0.3.1
```
Makes sense in that it would solve the issue of installing a newer version of an extension while the current version is loaded.

daviwil:
Hah, so that's how they decided to solve that problem. Works for me!

PR #65, opened 2016-01-02
Fixed issue #58, extension install path needs to specify version in path

rkeithhill:


README.md
```diff
@@ -1,7 +1,7 @@
 # PowerShell Language Support for Visual Studio Code
 
 This extension provides rich PowerShell language support for Visual Studio Code.
-Now you write and debug PowerShell scripts using the excellent IDE-like interface
+Now you can write and debug PowerShell scripts using the excellent IDE-like interface
```
daviwil: hah, whoops!

daviwil:
Thanks for fixing this!  Do you mind also updating examples\README.md since it also contains a reference to the path?

rkeithhill:
Done.  Hope you don't mind me substituting `$Home` for `$env:USERPROFILE`.  `$Home` is shorter and should just as well when run from PowerShell.  And for that matter, `$env:USERPROFILE` only works in PowerShell as well.  

FYI, someone may be tempted to use `~` as an even shorter form of `$home` but `~` doesn't work when you are cd'd into a psdrive other than the filesystem.

daviwil:
Nope, that's a great change, didn't know `$Home` was available!  Thanks for letting me know about the tilde context, another little PowerShell catch I didn't know about :)

Issue #63, opened 2016-01-01
Snippets discussion - productivity feature or teaching feature


rkeithhill:
I guess I've gotten used to C# snippets that are "functional/productive".  I find snippets like this in the PowerShell extension to be an awkward hybrid:
```powershell
switch ($x)
{
    'value1' {  }
    {$_ -in 'A','B','C'} {}
    'value3' {}
    Default {}
}
```
Seriously, what are the odds that I'll have `value1` as a case or a second or third case as shown above.  I believe these were written to **show how to use a feature** i.e. they are teaching.  That is admirable but as soon as the user **learns** the features of switch, this snippet becomes cumbersome because now it is a toss-up WRT how much code they have to "delete" to make the snippet conform to their intent versus how much they would have typed in the first place without using the snippet.  

I propose that at the very least the "common" language features lean more towards being productivity snippets and not "teaching" snippets.  I recommend changing the switch statement snippet to something like:
```powershell
switch ($variable) {
    condition { action; break }
    Default {}
}
```
Where the cursor is put on a `condtion placeholder` second after `$variable` and then the `action placeholder` would be third.

BTW here are a few more examples of what I'm talking about:
```powershell
try
{
    1/0
}
catch [DivideByZeroException]
{
    Write-Host "Divide by zero exception"
}
catch [System.Net.WebException],[System.Exception]
{
    Write-Host "Other exception"
}
finally
{
    Write-Host "cleaning up ..."
}

for ($i = 1; $i -lt 99; $i++)
{
    
}
```
Everybody who uses the try/catch/finally snippet is going to have to remove the DivideByZeroException and likely have to change the last catch to use an exception they want to catch (or not specify one at all).  I believe there should a single exception there and it should be a placeholder.  Likewise **everybody** is going to have to replace `99` in the for snippet.  Might as well make it a placeholder.

daviwil:
Agree completely, snippets should be functions to speed up coding.  If we want to have educational snippets, it would be better as a different feature that doesn't overload the expectations of the snippets feature.  Feel free to fix up any existing snippets that have extraneous stuff that isn't immediately useful!

rkeithhill:
OK, I've retained the "example" style snippets but they all begin with `ex-` in the auto-completion list.  All of the rest of the non `ex-` examples are focused on `productivity`, not on `teaching`.  I've modified the organization of the snippets file to put all the example snippets at the top of the file.

[...]

Issue #62, opened 2016-01-01
Snippets discussion - bracing style


rkeithhill:
I (reluctantly) would like to open a discussion regarding which bracing style to use for snippets.  Right now, the latest snippets file uses Allman style bracing e.g.:
```powershell
function foo
{
    ...
}
```
But the bulk of the PowerShell code I've seen in the wild uses K&R style e.g.:
```powershell
function foo {
    ...
}
```
I know there are details of bracing style the community at large doesn't agree upon like whether or not to use cuddled else but I do believe K&R is fairly well accepted at the "top level".

Would anyone object strenuously if I changed the snippets to use K&R?

daviwil:
I have no objection to that since it seems to be the prevalent coding style in PowerShell.  I'm not sure who wrote the original snippets but we should definitely make them conform to generally expected PowerShell coding style.

rkeithhill:
OK we are going with K&R style which I see more often out in the wild in PowerShell scripts.

Jaykul:
I vote K&R 
With cuddled else!
Because ... code folding.

Oh, right. VS Code doesn't do that yet. :-P

rkeithhill:
With the current snippets for `if`, `elseif` and `else`, you can control whether you get cuddling or not by where you invoke the snippet.  The following will give you a cuddled else:
```powershell
if ($foo) {
    ...
} else{tab}
```
And if you want standard K&R, invoke the snippet below the `if` statement's closing brace e.g.:
```powershell
if ($foo) {
    ...
} 
else{tab}
```

rkeithhill:
This has been implemented for the upcoming 0.4.0 release.

thegraffix:
In the future, would it be possible to introduce 2x different PowerShell.json files for each bracing style (K&R and Allman) so the user can choose an appropriate style?  Some organizations don't enforce a bracing style standard for PowerShell so it would be really useful to be able to switch between the two on the fly without the need to manually edit the snippets in the PowerShell.json file.

PR #68, opened 2016-01-03
Addresses issues #62 and #63 in snippets file.

rkeithhill:
We have switched to K&R bracing stlye and have made all 'example' style
be prefixed with 'ex-' so you can still use them but know when you are
getting an example-style snippet.

Also note that again there are lots of changes.  When I diff with `git diff` I see that my check in added a lot of `^M` chars (CRs).  I edited the file in VSCode so I can easily tell what the line endings are.  Also I have VSCode to trim trailing whitespace and I had to clean up some hard-coded tabs (replaced with 4 spaces).


rkeithhill:
BTW I really wish we could put comments in the snippets JSON file (among others).  :-(

daviwil:
Yeah, I know, I hate that the JSON files can't have comments...  You can thank Node.js for that in the case of package.json.

I think I need to add a .gitattributes file to fix the line ending and tab character mess.  Not sure how to do that yet, will have to investigate.

The changes look great!  I do expect some gripes from people who are using the old snippet names that are now prefixed with `ex-` but I do prefer the new composable approach.  We may just have to educate folks on that if they complain :)

PR #70, opened 2016-01-11
Open in PowerShell ISE

janegilring:
Added a new extension for vscode-powershell called OpenInISE - a feature to open PowerShell-files in PowerShell ISE by using the keybinding Ctrl+Shift+I.


daviwil:
Thanks for the great contribution, Jan!

Issue #76, opened 2016-01-28
Update default launch.json to use $file


daviwil:
The VS Code team has fixed the $file variable so that it always gets updated to the most recently opened file path when the debugger is started.  We should use this in the default launch configuration so that it's easier for users to get started debugging their scripts.

https://github.com/Microsoft/vscode/issues/1667

rkeithhill:
I can make this update.  Any word on when the insiders update for the next VSCode release will drop?  I thought they were shooting for the 26th?

daviwil:
Thanks Keith!  I think they're doing their final validation now.  I don't know when it's going to happen, but all of their testing work items just got finished up today:

https://github.com/Microsoft/vscode/issues?page=1&q=label%3Atestplan-item+milestone%3A%22Jan+2016%22

I'm guessing we'll see the preview drop in the next few days.

PR #78, opened 2016-01-30
Fix for #76 - launch.json now uses  as default launch path

rkeithhill:


rkeithhill:
I went ahead and submitted this fix because ${file} "works" in the current VSCode drop (Dec 2015). It just doesn't update the value as the user changes from one file in the editor to another.  But that was fixed as indicated by this VSCode closed issue - https://github.com/Microsoft/vscode/issues/1667  The milestone target for the fix is the January drop of VSCode.

daviwil:
Awesome, thanks!

PR #79, opened 2016-01-31
Dfinke cvttoclass

dfinke:
Copy from clipboard and convert JSON|CSV to PS|C# class syntax

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,12 @@
                 "command": "PowerShell.PowerShellFindModule",
                 "key": "ctrl+K ctrl+f",
                 "when": "editorTextFocus && editorLangId == 'powershell'"
+            },
+            {
+                "command": "PowerShell.PowerShellPasteAsClass",
+                "key": "ctrl+K ctrl+p",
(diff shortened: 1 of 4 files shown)


daviwil:
Usage of the copy-paste module is tricky.  Technically we can't bring in any new open-source library without getting legal approval due to the need for having proper copyright attribution.  In the world of npm, it's really difficult to stay on top of that due to the magniture of micro-dependencies that each project includes.

Is there any way to do the copy/paste behavior without the user of that library?  I guess VS Code doesn't provide API methods for clipboard management?

dfinke:
I'll look into it.

daviwil:
Discussed with Doug that we'll hold off on integrating this PR until we get the new command extensibility model going.  This feature would be a great validation of the new extensibility approach.  I'm keeping this PR open as a reminder to do that :)

daviwil:
Doug just shipped an update of his ConvertToClass module which leverages the $psEditor API! https://www.powershellgallery.com/packages/ConvertToClass/1.5.  Closing this PR now.

Issue #55, opened 2015-12-24
Add host bitness selection from VSCode UI.


daviwil:
As of the following pull request, PowerShell Editor Services now has both x86 and x64 hosts:

https://github.com/PowerShell/PowerShellEditorServices/pull/92

We should add a setting in the VS Code extension so that the user can select their desired host bitness.

rkeithhill:
While the actual support for 64bit and 32bit debug hosts will be in 0.4.0, we will require VSCode changes to support the "selection" of host via the VSCode UI.  That support won't be in the upcoming 0.10.7 drop of VSCode.  So the package.json update to support this will have to wait until the 0.5.0 drop (presumably by then VSCode will have UI support for this selection).

daviwil:
:thumbsup:

daviwil:
I guess the only thing left to do here is have a setting for the language service so that it can launch the x86 host if that's what the user wants?

rkeithhill:
One way to do this (and it is something I experimented with) is to modify the package.json initial configuration like so:
```javascript
				"initialConfigurations": [
					{
						"name": "PowerShell",
						"type": "PowerShell",
						"request": "launch",
						"program": "${file}",
						"args": [],
						"cwd": "${file}"
					},
					{
						"name": "PowerShell x86",
						"type": "PowerShell x86",
						"request": "launch",
						"program": "${file}",
						"args": [],
						"cwd": "${file}"
					}
				]
```
This has to be done for both debugger configurations.  This will give you a choice of which debugger in the Debug viewlet's dropdown.  

The problem I ran into is that when you initially select a debugger - either PowerShell or PowerShell x86 - the initially selected debugger in the dropdown didn't seem to be deterministic and not always the right one.  We could go this route, as the mitigation is simple - just select the right debugger from the drop down.  Thoughts?

daviwil:
Yeah having the additional x86 configuration included in the 64-bit initialConfigurations list might be nice to simplify things for the user.  For the x86 default configuration I think we'd still want the single configuration.

My question in this case is more about having a setting for the language service to be run with the x86 if necessary.  I think this may be necessary for the editing experience in the cases where some module or configuration is only available in x86 mode.

rkeithhill:
So the language service can cause modules to get imported to support completion text?  If so, then yeah, users would need to specify which language service host to use (32-bit or 64-bit).

daviwil:
[...]

PR #82, opened 2016-02-04
Addresses issue #55 for 0.4.0 release to select debug host bitness.

rkeithhill:
This change adds a new debugger configuration to the package.json file.  The user can simply change the debug type from 'PowerShell' to 'PowerShell x86' in their launch.json file to do 32-bit debugging.  Also, when they open a folder that doesn't have a .vscode\launch.json file, when they go to create that file, they can pick between 'PowerShell' and 'PowerShell x86'.  Of course on 32-bit OSs the 'PowerShell x86' selection is redundant with 'PowerShell'.  Oh well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,6 +134,36 @@
                         "program": "${file}"
                     }
                 ]
+            },
+            {
+                "type": "PowerShell x86",
+                "enableBreakpointsFor": {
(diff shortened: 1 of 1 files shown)


daviwil:
Awesome! I think this will make some folks happy.

PR #83, opened 2016-02-05
Updated Examples launch.json to use ${file}

rkeithhill:
This will make the example launch.json be consistent with the one that VSCode will create by default (from our package.json info).


daviwil:
Cool! Didn't even think about that.

PR #89, opened 2016-02-07
Fixes #88 by adding cwd initialConfig and configProps to package.json

### LINKED ISSUE #88: Allow user to customize the working dir that the debugger starts in
Technically you can do this without any change to the package.json file but without `cwd` being defined in the `configurationProperties`, VSCode will whine when you edit a launch.json file and add the property.

rkeithhill:
Also updated examples\.vscode\launch.json file. Sorry, couldn't resist replacing tabs with spaces. :-)

### DIFF
--- a/examples/.vscode/launch.json
+++ b/examples/.vscode/launch.json
@@ -1,11 +1,12 @@
 {
-	"version": "0.2.0",
-	"configurations": [
-		{
-			"name": "PowerShell",
-			"type": "PowerShell",
-			"request": "launch",
-			"program": "${file}"
-		}
-	]
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "PowerShell",
+            "type": "PowerShell",
+            "request": "launch",
+            "program": "${file}",
+            "cwd": "${file}"
+        }
+    ]
 }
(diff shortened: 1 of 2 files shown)


daviwil:
No problem on the conversion, I set my VS Code settings to auto convert now, so I've been converting every file I touch ;)

rkeithhill:
BTW what's up with the vscode-powershell\.vscode\launch.json file? Does it need to be updated.

rkeithhill:
I should have added in the comment that in this release of VSCode most of the debugger parameter description have been modified to indicate that you should pass only an absolute path.  No more workspace relative paths (use ${workspaceRoot} instead).

daviwil:
That's supposed to be for debugging the extension itself. Apparently it isn't working lately because someone tried to use it recently and said it didn't work :) probably should fix it at some point as the TypeScript we have here is getting larger.

daviwil:
That's OK, we may need to start writing some documentation to capture this stuff anyway.

rkeithhill:
Ah, sounds like a definite postpone then.  :-)

daviwil:
Yep!

PR #90, opened 2016-02-07
Fixes #24 - you can now use the launch.json args prop to send args to…

### LINKED ISSUE #24: args not passed to script being debugged from launch.json
I have tried several combinations of "args": [] and none of them produce a $args in the debugger.  Should the "args" parameter for a configuration work?  Looking through the source but not finding anything yet.
{
	"version": "0.1.0",
	"configurations": [
		{
			"name": "PowerShell",
			"type": "PowerShell",
			"program": "PowerShell Scripts/output.ps1",
			"args": ["Verbose:$true"],
			"request": "launch"
		}
	]
}

The debugger launches fine, but attempts using "-Verbose",  "-Verbose:$true",  or "/Verbose:$true" do not get passed to $args.  The "output.ps1" file is very simple:

   Write-Output $args	 
   Get-Process
   Write-Verbose -Message "Debugging PS in VSCode!"

rkeithhill:
… the script.

### DIFF
--- a/examples/.vscode/launch.json
+++ b/examples/.vscode/launch.json
@@ -6,6 +6,7 @@
             "type": "PowerShell",
             "request": "launch",
             "program": "${file}",
+            "args": [],
             "cwd": "${file}"
         }
     ]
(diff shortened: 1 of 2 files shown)


daviwil:
Awesome!

PR #91, opened 2016-02-08
Improvements to path processing snippets, added tests for these. Udpa…

rkeithhill:
…ted examples readme.md

Hopefully you don't mind a few more PS1 files in the "examples" dir.  I'm kind of using these as (Pester) tests for the path processing snippets.


daviwil:
That's a great idea actually. And thanks for updating the README! I think know well have to move that into a real documentation file soon.

PR #100, opened 2016-02-17
Prepare version 0.4.1

daviwil:


daviwil:
Will release this as soon as PowerShell Editor Services binaries are signed.

Issue #106, opened 2016-03-01
Extension should not register IntelliSense provider on OSX/Linux


daviwil:
It seems that we should not be setting up the LanguageClient when running on OSX and Linux because it disables the alternative completion behavior where symbols from the file get automatically completed.  Since the language service doesn't work on these platforms it would be good if the default IntelliSense worked there.

PR #110, opened 2016-03-10
Minor bug fixes for 0.5.0

daviwil:
Addressing issues #55 and #106.  Still looking into a couple more issues.


daviwil:
@rkeithhill I'm not auto-resolving #55 with that commit, will wait for your change to the default launch config before I resolve it.

PR #111, opened 2016-03-10
Update 'PowerShell' debug init config to offer both any & x86 dbgrs.

rkeithhill:
Also decided to use the new OS specifiers for 'windows' and 'winx86'.  I *think* this
will disallow folks on Mac and Linux to launch a debugger that won't work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,12 @@
                         "powershell"
                     ]
                 },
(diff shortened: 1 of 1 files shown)


daviwil:
Great!  Looks good to me.

PR #112, opened 2016-03-10
Modified Examples dir to provide a preview of PSake based module publ…

rkeithhill:
…ishing.  OK, take this for a spin and see what you think. FYI, PSake has not been published to the PSGallery yet.  They are shooting for end of the month.  For now, just grab it from here (https://github.com/psake/psake/releases) and install it manually.  Then press Ctrl+Shift+B to create the Release directory.  You can CD to that run ipmo SampleModule and play with the sample module.  When ready to publish, Ctrl+P then "task publish".  First time you should get a credential prompt as you to enter your NuGet API key.  I wonder if we should modify the prompt text for now to say "Enter anything - doesn't matter"?  The only other thing I'm worried about is that running Publish-Module could trigger bootstrapping which requires the user to respond.  Unfortunately the way the task is run, that "prompt" doesn't become visible.  However once you've bootstrapped publish-module once, there shouldn't be any more prompts.  Probably need to test this on a virgin system.


nightroman:
@rkeithhill, I wonder if it will be possible to use an alternative tool similar to PSake. I mean [Invoke-Build](https://github.com/nightroman/Invoke-Build).

Please do not get me wrong, I am not asking to use it. But it would be nice if in VSCode I can choose or at least to be able to configure for another build engine. Will this be possible?

daviwil:
@rkeithhill Awesome!  Excellent work on the documentation and everything.  I think it might be time for me to clean up or get rid of `DebugTest.ps1`, `StopTest.ps1`, and `Stop-Process2.ps1` so that the module example is a bit more pristine.  Maybe there's a way we can integrate the same concepts into your new module script files so that the same things can be demonstrated?  We don't need to do it for this release, thoug.

@nightroman You can definitely use an alternative build system by hooking up your own build script to VS Code's task interface.  Check out the `examples/.vscode/tasks.json` file that Keith changed as a part of this pull request.  You can just change that file to invoke whatever PowerShell build script you like.

rkeithhill:
Good part of what we are trying to figure out is what is exposed and what is done 'black box' by this extension. I lean towards having the extension rely upon existing infrastructure but help tie it together. By tie together I mean scaffold the project files required to accomplish build and publish. The two files that are of interest here are the build.ps1 file and the tasks.json file that defines how to invoke the build script. 

I could imagine providing alternatives during the scaffolding process.  That said, I'd like to steer folks towards the best solution for things like build and test. For test, Pester is the clear winner. For build, I liked that Psake had lots of contributors. But as @daviwil says, if we go with something similar to this approach you can easily plumb in invoke-build for your builds.

daviwil:
@rkeithhill Regarding the Publish-Module concerns, this is definitely going to be a problem with invoking those cmdlets directly through the task interface since the PowerShell engine won't be able to prompt the user.  Maybe we can open a new PowerShell.exe prompt to run those commands so that the console host can show the prompts?

For the NuGet key prompt we could say "Enter your PSGallery API key (or nothing, this isn't used yet in the preview)"

daviwil:
Forgot to ask, is this ready to be merged?  Looks good to me.

rkeithhill:
Not quite yet.  Few more tweaks.

daviwil:
No problem man, I'm not in a rush :)

rkeithhill:
OK, should be ready now.  BTW found a bug I think.  Peek definition doesn't seem to work for module-based functions.

daviwil:
You mean a function defined in another file in a module?  If so, I still need to implement "find references" support in general for functions defined in modules.  Only references we're traversing right now are dot-source references.

rkeithhill:
Yes.  I had thought about replacing the Stop/Stop-Processing examples by importing SampleModule.psd1 in DebugTest and then using the New-File command.  But peek def did not work on New-File.

daviwil:
Hmmm, interesting, will need to look into that.

nightroman:
@daviwil and @rkeithhill, thank you. Configurable `tasks.json` is the complete answer to my question.

PR #116, opened 2016-03-11
Adding foo[1].txt to .gitignore for when I'm building in the vscode-p…

rkeithhill:
…oweshell\examples dir


rkeithhill:
Actually .gitignore did not like `examples/Tests/foo[1].txt` but it was OK with `examples/Tests/foo*.txt`.

daviwil:
:+1:

PR #127, opened 2016-03-26
Resaved this file as UTF8 and tweaked...

rkeithhill:
to eliminate the "infinite" loop and replace with a delay in the while loop.  This will result in the script finishing in ~ 10 seconds which is enough time to pause the debugger to demo pausing. Somehow this file got saved as UTF-16LE which is why I think GitHub thinks it was binary before??


daviwil:
Yeah I had suspected that as well.  Still can't see the diff here but maybe that will change after the file is committed.  Looks good!

PR #128, opened 2016-03-27
Updated the build tasks.

rkeithhill:
Added install and compile task. Renamed existing compile task to compile-watch.


daviwil:
:+1:

PR #130, opened 2016-03-30
Add support for loading host profile scripts

### LINKED ISSUE #124: Provide configuration option to specify a PowerShell profile to load in the language service
...or ability to configure legacy Snapins to load to gain intellisense for their cmdlets.

daviwil:
This change adds support for loading both host-specific and host-agnostic
profile scripts for both the current user and all users on the same
system.  For VS Code, the profile names are Microsoft.VSCode_profile.ps1
and profile.ps1.  The new 'enableProfileLoading' setting controls whether
profiles are loaded automatically when the language server starts up.  For
now this setting is turned off by default.

Resolves #124.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/vscode-powershell/130)

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,10 @@
                 },
                 "args": [
                     "/debugAdapter",
-                    "/logLevel:Verbose"
+                    "/logLevel:Verbose",
+                    "/hostName:\"Visual Studio Code Host\"",
+                    "/hostProfileId:Microsoft.VSCode",
+                    "/hostVersion:0.5.0"
                 ],
                 "configurationAttributes": {
                     "launch": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,7 +121,10 @@
                 },
                 "args": [
                     "/debugAdapter",
-                    "/logLevel:Verbose"
+                    "/logLevel:Verbose",
+                    "/hostName:\"Visual Studio Code Host\"",
+                    "/hostProfileId:Microsoft.VSCode",
+                    "/hostVersion:0.5.0"
```
daviwil: The debugger configuration is a little unfortunate since we have to update the version in 3 places in this file now.  We'll need a version bump script in the near future.

daviwil:
Related pull request for PowerShell Editor Services:

https://github.com/PowerShell/PowerShellEditorServices/pull/199

package.json
rkeithhill: Which "host" version is this?  VSCode or PSES?  If PSES, can't we read the version number from an assembly attribute via refection?

daviwil: It's the VS Code extension's version. There will be a different property in the runspace for getting the PSES version

rkeithhill: Gotcha.

PR #141, opened 2016-04-14
Update examples/Build.ps1

randomchance:
Fixed issue in example build where Clean task would execute after Init, removing the $PublishDir before the Build task.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/vscode-powershell/141)

### DIFF
--- a/examples/Build.ps1
+++ b/examples/Build.ps1
@@ -157,7 +157,7 @@ Task Test -depends Build {
     Invoke-Pester $PSScriptRoot
 }
 
-Task Build -depends Clean -requiredVariables PublishDir, Exclude, ModuleName {
+Task Build -depends Clean, Init -requiredVariables PublishDir, Exclude, ModuleName {
     Copy-Item -Path $PSScriptRoot\* -Destination $PublishDir -Recurse -Exclude $Exclude
 
     # Get contents of the ReleaseNotes file and update the copied module manifest file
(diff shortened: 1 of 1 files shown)


rkeithhill:
Thanks for the fix!!  I borked this with my PublishRootDir change. What I was fixing there is that if a user decided to change the name of the module, the build script wasn't picking that up because we were only deleting the contents of the `.\Release\MyModuleName` dir instead of the directory (MyModuleName) itself.  The other thing that Init does that Clean depends on, is that it creates the PublishRootDir which may not exist.  Without that dir being created, the attempt to clean the contents of PublishRootDir would fail.

Could you update the Clean task to first test for the PublishRootDir and if it doesn't exist, then it has nothing to do.   Maybe ...
```powershell
if ((Test-Path $PublishRootDir) -and $PublishRootDir.Contains($PSScriptRoot)) {
...
```

This would likely only be for the very first time the build runs (or if the user nukes their Release dir).  Otherwise, it does what it does now, nukes the contents of PublishRootDir.  BTW could you change the `requiredVariables` on the clean task to be PublishRootDir instead of PublishDir (also the comment as well).  I apparently forgot to do that.  :-)

Then we should test these scenarios - starting with no Release dir, do a build.  Do another build with empty Release dir.  Do another build with Release\ModuleName dir.  Then change the name of the module's PSD1 file and repeat that last test to make sure it picks up the new name.  Thanks again!

randomchance:
I should have time to take care of it, but it likely won't be until end of day so if you want to do it by all means feel free!

rkeithhill:
Merged and I did those few tweaks.

PR #228, opened 2016-07-17
Changes to enable dev host to launch.

rkeithhill:
The changes are mainly to launch.json. The change to package.json is simply to reorder the scripts in the order they appear when you create a brand new extension.  Yeah, a little OCD on my part.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/vscode-powershell/228)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,19 +2,15 @@
 	"version": "0.1.0",
 	"configurations": [
 		{
-			"request": "launch",
-			"name": "Launch Extension",
-			"type": "extensionHost",
-			//"runtimeExecutable": "${execPath}",
-			"runtimeExecutable": "c:/dev/Monaco/tools/code.bat",
-			"args": [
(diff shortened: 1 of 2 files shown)


rkeithhill:
BTW there is a task in tasks.json called `install` but no corresponding script in package.json.  There is a `postinstall` script.  Should the tasks.json task be changed to `postinstall`?

daviwil:
Awesome!  I haven't used that tasks.json file in a really long time so I forgot it was there :)  Not sure about the install task, I think I grabbed this file from the VS Code team a long time ago so I don't even remember why it's there.  We can take it out if you like.

rkeithhill:
In this case, maybe I'll just copy over the default tasks.json I got with the new extension I created with `yo code`.  Let me do that tonight and push the change.  I just wanted to make sure you weren't using the install task somehow.

rkeithhill:
Ah, the `install` task essentially runs `npm install` which populates the node_modules folder.  OK.  I think this PR is good to go then.

Issue #249, opened 2016-08-22
Editor no longer uses ambient PSModulePath when launched from command line


chrisvanderpennen:
I've got a DSC project set up that includes a folder for modules intended to be packed and deployed to a pull server, including some custom ones for our own software.  I created a small script, to the effect of:

```
$env:PSModulePath = (Resolve-Path './src/modules') + ';' + $env:PSModulePath
code .
```

in the project root, which until recently launched VS Code and provided completion over the DSC resources in `./src/modules`.

As of 0.7.0, this no longer works. Code launches, but code analysis reports an error resolving any of the modules in `./src/modules` for `Import-DscResource` and executing `$env:PSModulePath` as a selection shows the system default module path.

Any ideas on how I can restore this functionality?

chrisvanderpennen:
I just tried `"powershell.enableProfileLoading": false`, didn't help.

rkeithhill:
You can create a profile called `Microsoft.VSCode_profile.ps1` and configure $env:PSModulePath in there plus set `"powershell.enableProfileLoading": true`.   That should work on 0.7.0.

chrisvanderpennen:
Great, that worked.  Thanks!

I've just tinkered with making the PS host name configurable so you can load a project-specific profile, would you like me to open a pull request?

daviwil:
I think we're actually looking at a different way to accomplish this through an additional settings parameter in the project's settings.json file.  Definitely would be interested in a PR for that!  If you want, we can discuss this a little bit more before you get started.

daviwil:
Oh, and regarding the initial issue, we've changed the way that we launch our language service using a new instance of PowerShell.exe so maybe the context doesn't get carried across anymore.  I'll look into it a bit more to understand the core issue.

rkeithhill:
@daviwil I looked briefly at how we use child_proces.spawn() and it would appear that the child processes are supposed to inherent the env block.  So the env var should have been defined in the PowerShell process.  With ProcExp I do see it defined in some of the various code.exe instances.  

Man, why does it load so many code.exe processes when I only have one editor instance open?

daviwil:
Yeah, the problem is that VS Code is launching their plugin hosts in child processes and apparently those don't always inherit the env variables without Code doing some extra work.  This has actually caused a problem in VS Code running on Linux because Electron there is more strict about env var inheritance for some weird reason.

SydneyhSmith:
[...]

PR #255, opened 2016-08-22
Add setting for PS host name

chrisvanderpennen:
This allows users to use a project-specific PowerShell profile script. PowerShell/vscode-powershell#249

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/vscode-powershell/255)

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,6 +172,11 @@
                     "default": true,
                     "description": "If true, causes user and system wide profiles (profile.ps1 and Microsoft.VSCode_profile.ps1) to be loaded into the PowerShell session.  This affects IntelliSense and interactive script execution.  The debugger is not affected by this setting."
                 },
+                "powershell.profileName": {
+                    "type": "string",
+                    "default": "Microsoft.VSCode",
+                    "description": "Specifies the PowerShell session host name.  Used to change the PowerShell profile loaded."
+                },
                 "powershell.scriptAnalysis.enable": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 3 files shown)


daviwil:
Hey Chris, thanks for sending this PR!  I think I might not have been clear on what I meant when we discussed this earlier.  The idea is that the host profile name for the editor shouldn't be modifiable via settings.

I think a user would probably want to have the standard editor-level profile in addition to a project-level one.  With this change, the profile setting will override the standard VS Code profile path for anyone who opens a workspace folder which has this setting in the workspace-level settings.json file.  If you have multiple people working on a PowerShell module in the same repo, one user could break the profiles of all the others by using this setting.

The ideal way to approach this would be to add an additional "startup script" setting which could be specified in the workspace-level settings.json file.  This would allow developers working in a shared repo to have the necessary startup behavior performed only when that workspace is loaded (while maintaining their own individual VS Code profiles).

This work would definitely require some additional changes to PowerShell Editor Services, so you'd end up having to do a PR in both places.  Let me know if you're still interested in helping with that and I can help you get started!

chrisvanderpennen:
Hi David. That makes more sense and I'm definitely interested in helping out with that.

Based on a quick glance through Editor Services it _looks_ like we should be able to pass an additional optional path parameter through to `Start-EditorServicesHost` which we would then prepend to `$profilePaths` if present, but it wouldn't surprise me if I'm missing something - I'm not super familiar with how PowerShell profile loading works.

PR #280, opened 2016-09-07
Update troubleshooting.md

darrmitchell:
Corrected the symlink commands and added a note about /usr/local/lib being non-existent on a fresh install of 10.11.6

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -27,11 +27,11 @@ You can use [Homebrew](http://brew.sh) to easily install OpenSSL.  First, instal
 brew install openssl
 ```
 
-After installation, the libraries of interest must be symlinked to `/usr/local/lib`; e.g. (adjust the version numbers as needed):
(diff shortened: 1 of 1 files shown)


daviwil:
Did the symlink commands not work for you without putting the filename at the end?  At any rate, thanks a lot for making that update!

darrmitchell:
I just tested on another Mac and when /usr/local/lib exists the symlink
command works without specifying the filename. My issue was apparently
that /usr/local/lib didn't exist so when I symlinked it just created
"lib" as a file instead of a directory.

On Wed, Sep 7, 2016, at 09:41 AM, David Wilson wrote:
> Did the symlink commands not work for you without putting the filename
> at the end?  At any rate, thanks a lot for making that update!
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[1], or mute the thread[2].
>


Links:

  1. https://github.com/PowerShell/vscode-powershell/pull/280#issuecomment-245283781

daviwil:
Ahhh, great point actually, it's good to include the filename to avoid this problem.  Thanks a lot, merging this!

PR #308, opened 2016-10-03
troubleshooting.md: use /usr/local/opt/openssl for link source

cielavenir:


daviwil:
Thanks @cielavenir!  I don't have my MacBook handy, is the new path a symbolic link which Homebrew creates so that we don't need to point to a specific install of OpenSSL?

cielavenir:
Uh, actually I'm new to Homebrew... (I just installed VSCode extension and followed the information link to come here)
[It looks like] /usr/local/Library/LinkedKegs used to be used as versioning-free space, but it is now moved to /usr/local/opt (according to https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils.rb#L547 ).

daviwil:
@cielavenir thanks again for sending this PR.  I finally got a chance to verify this path on my MacBook and it seems that this path works!  Thanks a lot for locating this fix, it should help people avoid changing their link for future OpenSSL versions.

PR #324, opened 2016-10-23
Replace program with script in launch.json.

rkeithhill:
"program" is left in place with updated description indicating the user should use the  "script" property instead.  This will allow existing PowerShell projects to continue work with newer versions of PSES.

Any new launch.json files created with this updated extension will generate just "script". This means that if someone loads the project in an older version of the extension, it won't launch the debugger correctly.  The question is, given how *easy* it is to update PSES, do we think this will be a big problem?

I don't "think" so and I think the term "script" will make more sense to PowerShell users in the long run.  That said, I'm totally OK with rejecting this PR if other folks do not agree.  BTW there is a companion PR on PSES to implement the PSES side of this change.

### DIFF
--- a/examples/.vscode/launch.json
+++ b/examples/.vscode/launch.json
@@ -5,7 +5,7 @@
             "name": "PowerShell",
             "type": "PowerShell",
             "request": "launch",
-            "program": "${file}",
+            "script": "${file}",
             "args": [],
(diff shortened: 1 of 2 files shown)


daviwil:
This will definitely help reduce confusion.

PR #346, opened 2016-11-24
Update troubleshooting.md

guillem-riera:
Current homebrew openssl version is 1.0.2j.


daviwil:
Thanks for the PR @guillem-riera!  Based on what I see in PR #308, the best way to avoid this issue in the future is to use the version-agnostic path to the OpenSSL libs for the source of the link.  I've merged that PR instead of yours since it will save us from having to update the documentation for each OpenSSL update.

Thanks again!

PR #350, opened 2016-11-28
Improve PowerShell session management, status reporting, and logging

### LINKED ISSUE #281: Provide progress indicator for language/debug server startup
Some users are attempting to start debugging just after loading a PowerShell script for the first time in a session.  In some cases the language and debug servers have not yet finished initializing so there needs to be some indication in the UI that the extension isn't fully initialized yet.

daviwil:
This change introduces a pretty large refactoring of the existing extension code to improve the overall user experience.  Here is a summary of the changes:

- The PowerShell session is now restartable at any time while VS Code is running without restarting the whole VS Code window.  This is exposed through the "Restart Current Session" command.
- A new status indicator has been added to the status bar which lets the user know when the language server is starting, running, or in an error state.
- A new session menu, accessible from the status bar indicator, allows the user to easily switch their PowerShell session between x86/x64 and a specific PowerShell exe path.
- Logs for the extension are now written to an output channel in the UI allowing the user to show them at any time with the "Show PowerShell Extension Logs" command.
- Any failure to load the PowerShell session will show an error message to the user and give them the opportunity to show the extension logs.
- If the PowerShell process ends unexpectedly at any time, the user will be shown a prompt to restart the session.
- Logs for each session are now written out to individual log folders instead of creating them all in the same folder.

Resolves #281.


daviwil:
This is a pretty big change so I'll give @rkeithhill a chance to take a look before merging :)

src/logging.ts
```diff
@@ -1,5 +1,158 @@
...
+
+    public getLogFilePath(baseName: string): string {
+        return path.resolve(this.logBasePath, `${baseName}.log`);
+    }
+
+    public writeAtLevel(logLevel: LogLevel, message: string, ...additionalMessages: string[]) {
+        if (logLevel >= this.MinimumLogLevel) {
+            // TODO: Add timestamp
+            this.logChannel.appendLine(message);
+            fs.appendFile(this.logFilePath, message + "\r\n");
```
rkeithhill: Should these line endings change to `\n` for Linux and macOS?

daviwil: Hmm, yeah, good point.  In fact, I need to verify that this all works correctly on Linux/macOS anyway, so I'll change that before I merge.

daviwil:
@rkeithhill Quick question: which nomenclature do you think is more appropriate for PowerShell folks, 64-bit/32-bit or x86/x64?  The former is clearer but the latter seems to be used for the Windows PowerShell shortcuts (x86 at least).

rkeithhill:
I'd go with the latter for the reason you mention - names of shortcuts and what shows up in the window title.

package.json
```diff
@@ -32,16 +32,17 @@
         "vscode-languageclient": "1.3.1"
```
kapilmb: Is it missing "vscode-jsonrpc" ? I cannot compile it without adding "vscode-jsonrpc" as a dependency.

daviwil: Hmm, that shouldn't be the case, I'll come by your office later to take a look.

PR #358, opened 2016-12-06
Create a choice list of PSSA rules

kapilmb:


kapilmb:
@rkeithhill For now, it doesn't interact with the PSSA settings file - If the user supplies a settings file, it will not show any rule list. Otherwise, it will show a list of rules from a drop down to select from.

vscoderulesselection

kapilmb:
But, in the near future, this feature will take into consideration the data in the settings file and hence can provide an interface to udpate the settings file.

src/checkboxQuickPick.ts
```diff
@@ -0,0 +1,92 @@
...
+export class CheckboxQuickPick {
+    private options: ICheckboxOption[];
+    private readonly confirm: string;
+    private readonly checkboxOn: string;
+    private readonly checkboxOff: string;
+    private readonly confirmPlaceHolder: string;
+
+    constructor(options: ICheckboxOption[]) {
+        this.options = options;
+        this.confirm = figures.tick;
```
daviwil: Instead of using the 'figures' package, it'd be more convenient to use the built-in icons that you can reference by a special string syntax.  Here's an example: https://github.com/PowerShell/vscode-powershell/blob/master/src/session.ts#L391.  Those string names are refer to the names of GitHub's [Octicons](https://octicons.github.com/).  In this case you could just use [`$(check)`](https://octicons.github.com/icon/check/) to achieve the same effect.

kapilmb:
Here is the equivalent PR on EditorServices: PowerShell/PowerShellEditorServices#311

src/features/SelectPSSARules.ts
```diff
@@ -0,0 +1,68 @@
...
+                        });
+                    let requestParams: SetPSSARulesRequestParams = {filepath, ruleInfos};
+                    this.languageClient.sendRequest(SetPSSARulesRequest.type, requestParams);
+                });
+            });
+        });
+    }
+
+    public setLanguageClient(languageclient: LanguageClient): void {
+        this.languageClient = languageclient;
```
daviwil: One more thing here:  I think your request for getting the PSSA rules needs to be done in setLanguageClient.  Since this method is called when the session is started, restarted, or switched to a new runtime, we'll need to re-query this list to make sure the list is up to date.

It makes sense to maintain the list of selected rules across sessions, though.  This might require getting the list of rules and comparing it against the current list of rules that have been set so that you can set those rules on the new session.  If this is too complicated to get done right now, we at least need to get the set of rules from the server when the LanguageClient changes.

kapilmb: The current implementation doesn't required the client the maintain any rule state at all. The client retrieves the rule list whenever the user opens the quickpick menu and sends the updated list to the server. It does not maintain any state about the rule list. 

So, if the LangauageClient changes and if the user opens the quickpick menu, it will get a new list from the server.

daviwil: Oh! OK, that makes sense. I should have paid closer attention :) cool, will merge this now!

PR #373, opened 2016-12-14
Add support for new project creation using Plaster

daviwil:
This change adds a new command called "Create New Project from Plaster
Template" which allows the user to select a Plaster template and go
through the template creation process.  Once the project is created, a new
window is opened for the new project's path.


daviwil:
Probably should have put the CheckboxQuickPick refactoring in a different PR but I'll need it for the Plaster integration when I finish the multi-choice selection support.

PR #375, opened 2016-12-15
Show powershell status only for powershell documents

kapilmb:


daviwil:
I'm going to wait on merging this one for a little while because I think there's some value in having the indicator visible when a PowerShell file isn't being edited.

One example: the user tries to create a new PowerShell project from template while they're editing a Markdown file and haven't activated the PowerShell extension yet.  They should see the status bar indicator appear to know that PowerShell is being loaded and then active.  Once the template creation is finished though, there's not as much reason to show it.

Let's wait and see if anyone complains about it being visible and then possibly merge this in 0.9.0.

kapilmb:
I didn't see typscript or c# extension show their status on files that are not associated with them and I assumed that is the correct behavior. Also, it is a bit inconsistent that powershell status can be seen when you are actually editing a C# file but not the other way round.

daviwil:
OK, I'll merge it.

kapilmb:
Thanks @daviwil

kapilmb:
I just went back and saw the code - I think these changes should cover the example you mentioned. Lets say, you create a new PowerShell project from a plaster template while editing a markdown - in that case the activation event triggers which should show PS status even though you are on a markdown file. Only when we change the active editor, the event triggers which then will take into account the document id.

PR #419, opened 2016-12-29
Added module manifest code snippet

pcgeek86:
- [x] Added a PowerShell code snippet to `PowerShell.json`, for a module manifest skeleton

Cheers,
**Trevor Sullivan**
Docker Captain
Microsoft MVP: Cloud & Data Center Management
https://trevorsullivan.net
https://twitter.com/pcgeek86


pcgeek86:
The intent behind this is to reduce friction for a developer to create new PowerShell modules, and publish them to the PowerShell Gallery or a private module repository.

daviwil:
Reasonable addition, but have you seen the new [Plaster](https://github.com/PowerShell/Plaster) integration in 0.8.0?  Takes it one step further than psd1 creation by creating a basic module structure.

https://github.com/PowerShell/vscode-powershell/blob/master/CHANGELOG.md#create-new-modules-with-plaster

pcgeek86:
That's helpful -- didn't know about that. I guess having a snippet available is just a secondary approach to building modules.

rkeithhill:
My only concern with this is that it is a static snapshot and it's not like the module manifest is completely static.  It has changed over various releases of PowerShell.   This is the primary reason Plaster does not use a static snapshot of a PSD1 file.  It invokes the New-ModuleManifest command instead.

pcgeek86:
We can certainly update it as time goes on. There are multiple approaches to building a module. This is just one route. If Plaster is a better option for certain developers, I'm sure they'll use it.

daviwil:
I'll go ahead and merge it, let's see what people think in the next update and then we can make improvements.  Thanks Trevor!

PR #423, opened 2017-01-09
Add new "attach" and "interactive" debugging configurations

daviwil:
This change adds new configuration possibilies for PowerShell debugging in
the extension.  The first is simply to allow the user to not specify a
script file in their "launch" configuration; when such a configuration is
launched it merely drops them into an interactive shell in the Debug
Console where they can enter whatever commands they like.

The other new configuration is a new "attach" configuration which allows
the user to specify a process ID or a remote computer and process ID to
which the debugger should be attached when it launches.  This provides a
smooth attach/detach experience so that the user does not need to go into
an interactive console to initiate the Enter-PSSession and
Enter-PSHostProcess commands manually (though this is still possible and
supported in interactive debugging mode).

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
 
                 "configurationSnippets": [
                     {
-                        "label": "PowerShell: Launch Current File Configuration",
+                        "label": "PowerShell: Launch Current Script Configuration",
                         "description": "A new configuration for launching the current opened PowerShell script",
                         "body": {
                             "type": "PowerShell",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -160,7 +160,7 @@
 
                 "configurationSnippets": [
                     {
-                        "label": "PowerShell: Launch Current File Configuration",
+                        "label": "PowerShell: Launch Current Script",
```
rkeithhill: Script is better than file but you might want to leave in "Configuration" at the end.  This is what I see for the other langs that support debug configuration snippets:
image

daviwil: Yeah, you're right, I'll change it back

rkeithhill:
LGTM

daviwil:
Thanks Keith!

PR #426, opened 2017-01-12
Add feature to format PowerShell source files

kapilmb:
codeformattingdemo

Related PRs:
* PowerShell/PowerShellEditorServices#332
* PowerShell/PSScriptAnalyzer#690


src/features/DocumentFormatter.ts
```diff
@@ -0,0 +1,277 @@
...
+        options: FormattingOptions,
+        token: CancellationToken): TextEdit[] | Thenable<TextEdit[]> {
+        return this.executeRulesInOrder(document, range, options, 0);
+    }
+
+    executeRulesInOrder(
+        document: TextDocument,
+        range: Range,
+        options: FormattingOptions,
+        index: number): Thenable<TextEdit[]> | TextEdit[] {
```
daviwil: I think that we could simplify this code a lot by using the TextEdit class as the means for transporting the formatting edits back and forth between extension and server.  Check out VS Code's [TextEdit](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-nametexteditaspan-classcodeitem-id586texteditspan) class compared to the one we have in the [Protocol assembly](https://github.com/PowerShell/PowerShellEditorServices/blob/develop/src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs#L48), they are exactly the same.

I'd recommend pushing as much of this logic to the service side as you can and only sending back the valid TextEdits from the server which will be applied by the client.  In fact, I think we should implement this based on the exact protocol messages that are coming in the Language Server Protocol 3.0 since we'll be snapping to that pretty soon anyway.  Check these out:

https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#textDocument_formatting

The request type is fairly similar to yours now and the response type is simply an array of `TextEdit`.  Implementing it this way will future-proof us a lot better.

daviwil:
Really happy to have this feature now!  I think we need to take a slightly different approach with the client/server protocol for this to align with the [Language Server Protocol](https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md) since they've already specified the message types for document and range formatting.  I'm happy to chat about it tomorrow at the office if you have questions.

package.json
```diff
@@ -270,6 +270,21 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Launches the language service with the /waitForDebugger flag to force it to wait for a .NET debugger to attach before proceeding."
+                },
+                "powershell.codeformatting.openBraceOnSameLine":{
```
kapilmb: fixed!

src/features/DocumentFormatter.ts
```diff
@@ -0,0 +1,277 @@
+import vscode = require('vscode');
```
kapilmb: fixed.

src/features/DocumentFormatter.ts
```diff
@@ -0,0 +1,277 @@
...
+    private languageClient: LanguageClient;
+
+    // The order in which the rules will be executed starting from the first element.
+    private readonly ruleOrder: string[] = [
+        "PSPlaceCloseBrace",
+        "PSPlaceOpenBrace",
+        "PSUseConsistentIndentation"];
+
+    // Allows edits to be undone and redone is a single step.
+    // Should we expose this through settings?
```
kapilmb: fixed.

daviwil:
Kapil and I discussed the protocol changes I asked for, we're going to wait until the next extension update for that work.  I'll merge this PR and the one on PSES after the PSScriptAnalyzer PR gets merged.

daviwil:
Great work!  I'll play around with this a bit later today.

PR #431, opened 2017-01-13
Add startSessionCommand handler for debugger configuration

daviwil:
This change leverages the new "setSessionCommand" configuration parameter
for debugger contributions which allows us to provide a command that will
take a debug launch configuration, tweak it, and then launch the debugger
ourselves.  This allows us to do any necessary launching or
pre-configuration of our debug adapter.  It also allows us to create a
default launch configuration on the fly if one doesn't already exist.
This means that users can now debug the current script without a
launch.json either with or without a folder open!

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "activationEvents": [
         "onLanguage:powershell",
         "onCommand:PowerShell.NewProjectFromTemplate",
-        "onCommand:PowerShell.OpenExamplesFolder"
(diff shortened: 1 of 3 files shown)


daviwil:
I'm really happy with how well this works.  And the best part is that it's compatible with VS Code 1.8.1 because it's only additive, no changes to existing debugger configurations, etc.  I'll ship it with 0.9.0 and it'll light up as soon as VSC 1.9.0 ships.

PR #435, opened 2017-01-15
Provide quick pick list of processes when using debugger attach configuration

rkeithhill:
This needs **serious review/tweaking** as Type/JavaScript is not my forte.  For instance, the node code (and this) returns null when the user doesn't select anything.   In that case, the debugger still tries to do something.  

FWIW it won't break my heart if this doesn't make 0.9.0.  That said, I'm hoping that with your understand of the VSCode side of the extension, this can be fixed to work better.  :-)

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -4,7 +4,7 @@
 
 import vscode = require('vscode');
 import { IFeature } from '../feature';
-import { LanguageClient } from 'vscode-languageclient';
+import { LanguageClient, RequestType, NotificationType } from 'vscode-languageclient';
 
 export class DebugSessionFeature implements IFeature {
     private command: vscode.Disposable;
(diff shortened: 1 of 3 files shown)


src/features/DebugSession.ts
```diff
@@ -40,4 +40,109 @@ export class DebugSessionFeature implements IFeature {
...
+            return this.pickPSHostProcess();
+        }
+    }
+
+    public dispose() {
+        this.command.dispose();
+    }
+
+    // In node, the function returned a Promise<string> not sure about "Thenable<string>"
+	private pickPSHostProcess(): Thenable<string> {
```
daviwil: `Thenable<T>` is an interface that `Promise<T>` implements, so you're good.

src/features/DebugSession.ts
```diff
@@ -40,4 +40,109 @@ export class DebugSessionFeature implements IFeature {
...
+		return this.languageClient.sendRequest(GetPSHostProcessesRequest.type, null).then(hostProcesses => {
+			var items: ProcessItem[] = [];
+
+			for(var p in hostProcesses) {
```
daviwil: Maybe if the length of hostProcesses is 0 we should use `vscode.window.showInformationMessage` to inform the user that there are no host processes to choose from?

daviwil:
Looks good overall!

src/features/DebugSession.ts
rkeithhill: I pushed a commit to address that.  But we still have a problem.  When there are no host processes (and the command picker returns null) OR when the user presses Esc to close the quick pick UI without selecting anything, the extension continues to try to attach.  I did a quick search in the logs and apparently the JSON reader on the other side doesn't like me returning null when it expects an int:
```
    System.AggregateException: One or more errors occurred. ---> Newtonsoft.Json.JsonReaderException: Could not convert string to integer: null. Path 'processId', line 1, position 128.
       at Newtonsoft.Json.JsonReader.ReadAsInt32Internal()
       at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadForType(JsonReader reader, JsonContract contract, Boolean hasConverter)
```
And when the user presses Esc instead of choosing, same thing:
```
    System.AggregateException: One or more errors occurred. ---> Newtonsoft.Json.JsonSerializationException: Error converting value {null} to type 'System.Int32'. Path 'processId', line 1, position 124. ---> System.InvalidCastException: Null object cannot be converted to a value type.
       at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
```
I'm thinking we should change the PSES Attach request args to take a string for the process id to make the JSON deserializer happy and then in the code, we can test for String.IsNullOrEmpty and abort attach in that scenario. Or does JSON handle Nullable<int>?  Is so, that would probably be simpler.  Thoughts?

rkeithhill: BTW when there is a process to attach to, works slick!!

rkeithhill: Actually I think this PR was good to go.  I'll tweak PSES on a new topic branch.

daviwil: Sounds good!

rkeithhill:
Adding this as a resource for future changes - https://github.com/Microsoft/vscode/issues/6569

PR #437, opened 2017-01-16
Fix host process picker when session init'd.

rkeithhill:
Also fix the "initialConfigurations" attach setting to use the host process picker and set a default runspaceId.

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -76,13 +76,20 @@ export class PickPSHostProcessFeature implements IFeature {
 
                     // If PowerShell isn't finished loading yet, show a loading message
                     // until the LanguageClient is passed on to us
+                    var cancelled = false;
+                    var timedOut = false;
                     this.waitingForClientToken = new vscode.CancellationTokenSource();
+
                     vscode.window
                         .showQuickPick(
                             ["Cancel"],
-                            { placeHolder: "Select PowerShell Host Process to attach to: Please wait, starting PowerShell..." },
+                            { placeHolder: "Attach to PowerShell host process: Please wait, starting PowerShell..." },
                             this.waitingForClientToken.token)
-                        .then(response => { if (response === "Cancel") { this.clearWaitingToken(); } });
+                        .then(response => {
+                            if (response === "Cancel") {
+                                this.clearWaitingToken();
+                            }
+                        });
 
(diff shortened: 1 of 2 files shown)


src/features/DebugSession.ts
```diff
@@ -105,7 +114,7 @@ export class PickPSHostProcessFeature implements IFeature {
 
         if (this.waitingForClientToken) {
             this.clearWaitingToken();
-            return this.pickPSHostProcess();
```
daviwil: Shouldn't this still be called so that the menu will be displayed when the connection is finally established?

rkeithhill: It will display the UI but the resulting pick is dropped on the floor AFAICT.  As I understand it, the whole command like so:
1. User picks attach to in debug panel
2. Dbg config for attach has `${command.pickPSHostProcess` configured to get the `processId` value
3. That command is invoked and in turn, activates PowerShell since it is registered as activation event
4. When the command completes, VSCode uses the returned `string` as the `processId` value.  If the command does not return a string, the adapter gets sent "undefined" for `processId`.

So what I think we need is a synchronization primitive (like *ResetEvent / CriticalSection) so that the command can "wait /w timeout" on the setLanguageClient to signal that it has been initialized.

BTW in the process of debugging this I found a bug (two actually in VSCode) :-)  https://github.com/Microsoft/vscode/issues/6569#issuecomment-272828999

rkeithhill: BTW if we want to punt on this for 0.9.0, I could return an error message when PowerShell hasn't been started along the lines of `"You must start a PowerShell session before using attach to PowerShell host process"`.

rkeithhill: OK so reading up some on JavaScript - no need for sync primitives due to single-threaded nature of JavaScript/node. Hmm...

daviwil: Yep you shouldn't need to worry about thread synchronization in JavaScript.  If this is good enough for now I'll merge it and then poke around with it a bit.

rkeithhill: Yes, go ahead and merge it and see what you think.  Hopefully, you can figure out a better way to handle the PowerShell not initialized case.

daviwil:
Thanks Keith!

PR #441, opened 2017-01-18
Improve "Attach to Host Process" menu flow

daviwil:
This change improves the flow of the "Attach to Host Process" menu when
the language service hasn't started yet.  It also changes the text display
slightly.

Here's a screenshot of the updated menu items:

image

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -67,48 +67,14 @@ export class PickPSHostProcessFeature implements IFeature {
     private command: vscode.Disposable;
     private languageClient: LanguageClient;
     private waitingForClientToken: vscode.CancellationTokenSource;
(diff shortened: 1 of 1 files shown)


daviwil:
@rkeithhill let me know what you think about the text changes in the menu items.  I noticed that the "detail" text was getting right-aligned for some weird reason, not sure if VS Code decided to do that somehow.  I changed it to put everything in the description field so that it's more uniform.

rkeithhill:
LGTM!  Glad you figured out a better way to handle the not-initialized case!  BTW, can you verify that search in the quick pick works on the pid, as well as title and process name.

daviwil:
I learned some new things about Promises today, I can tell you that much ;)  Yep, searching by pid works perfectly.  Thanks Keith!

PR #449, opened 2017-01-19
Fix typo in PowerShell.json

bobfrankly:
Noticed the value for Copyright in the Manifest snippet was missing a closing single quote. This fixes it.


daviwil:
Good catch, thanks @bobfrankly!

PR #459, opened 2017-01-24
Upload extension VSIX to AppVeyor on successful build

daviwil:
This should make it easier for users to try out bug fixes before we ship an official release.  I'll probably restrict this to only upload artifacts on the `develop` branch before I merge the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,23 @@ To open/view the extension's examples in Visual Studio Code, run the following f
 code (Get-ChildItem $Home\.vscode\extensions\ms-vscode.PowerShell-*\examples)[-1]
 ```
 
+## Installing the Extension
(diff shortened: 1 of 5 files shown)


daviwil:
There's a bug in VS Code that makes it a little difficult to install these "insider" packages, it seems to ignore a newer "insider" version if an older official version is already installed:

https://github.com/Microsoft/vscode/issues/19259

If anyone wants to try out these new insider packages, you'll have to uninstall 0.9.0 (or the current official release) before the insider package will work.

daviwil:
Also worth noting that PSScriptAnalyzer and Plaster are not included with these builds right now.  I'll be changing that within the next couple of days.

PR #461, opened 2017-01-25
Show formatting status in status bar

kapilmb:
Sometimes, formatting gets blocked by ongoing PSSA invocations and there is no way to know if the extension is formatting the document or not. This addition allows us to show an animated status to indicate formatting status.

### DIFF
--- a/src/animatedStatusBar.ts
+++ b/src/animatedStatusBar.ts
@@ -1,3 +1,7 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
 import {
     StatusBarItem,
     StatusBarAlignment,
--- a/src/animatedStatusBar.ts
+++ b/src/animatedStatusBar.ts
@@ -9,12 +9,12 @@ import {
     window} from "vscode";
 
 export function setAnimatedStatusBarMessage(text: string, hideWhenDone: Thenable<any>): Disposable {
-    let animatedStatusBarItem: AnimatedStatuBarItem = new AnimatedStatuBarItem(text);
+    let animatedStatusBarItem: AnimatedStatusBarItem = new AnimatedStatusBarItem(text);
     animatedStatusBarItem.show(hideWhenDone);
     return animatedStatusBarItem;
 }
 
-class AnimatedStatuBarItem implements StatusBarItem {
+class AnimatedStatusBarItem implements StatusBarItem {
     private readonly animationRate: number;
     private statusBarItem: StatusBarItem;
     private maxCount: number;
--- a/src/animatedStatusBar.ts
+++ b/src/animatedStatusBar.ts
@@ -11,13 +11,15 @@ export function setAnimatedStatusBarMessage(text: string, hideWhenDone: Thenable
 }
 
 class AnimatedStatuBarItem implements StatusBarItem {
+    private readonly animationRate: number;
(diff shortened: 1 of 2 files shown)


src/animatedStatusBar.ts
```diff
@@ -0,0 +1,123 @@
...
+    Disposable,
+    window} from "vscode";
+
+export function setAnimatedStatusBarMessage(text: string, hideWhenDone: Thenable<any>): Disposable {
+    let animatedStatusBarItem: AnimatedStatuBarItem = new AnimatedStatuBarItem(text);
+    animatedStatusBarItem.show(hideWhenDone);
+    return animatedStatusBarItem;
+}
+
+class AnimatedStatuBarItem implements StatusBarItem {
```
daviwil: Slight typo, should be AnimatedStatusBarItem

src/animatedStatusBar.ts
```diff
@@ -0,0 +1,123 @@
...
+    hide(): void {
+        this._stop();
+        this.statusBarItem.hide();
+    }
+
+    dispose(): void {
+        this.statusBarItem.dispose();
+    }
+
+    _updateCounter(): void {
```
daviwil: TypeScript has 'private' visibility, you should use that instead of the underscore naming convention

src/features/DocumentFormatter.ts
```diff
@@ -110,14 +111,17 @@ class PSDocumentFormattingEditProvider implements DocumentFormattingEditProvider
         range: Range,
         options: FormattingOptions,
         token: CancellationToken): TextEdit[] | Thenable<TextEdit[]> {
-        return this.executeRulesInOrder(document, range, options, 0);
+
+        let textEdits: Thenable<TextEdit[]> = this.executeRulesInOrder(document, range, options, 0);
+        AnimatedStatusBar.setAnimatedStatusBarMessage("formatting", textEdits);
```
daviwil: I'd recommend a more obvious string like "PowerShell code formatting in progress"

daviwil:
I just realized why you used "formatting" as the status bar string, because it's a status bar item.  Maybe call it "Formatting file"?

src/animatedStatusBar.ts
kapilmb: fixed.

src/animatedStatusBar.ts
kapilmb: fixed.

kapilmb:
I though even "formatting..." looks a bit big. Here is how it looks now:

formattingstatus

daviwil:
It looks like the normal status bar text, I think it's fine to be a bit longer.  I can make the change.

kapilmb:
The gif rendering here is a bit messed up I think. 
 
I am planning on adding a few more options like these dots, e.g a spinning wheel. I can check in the longer status when I make the change.

daviwil:
I think we can wait on other animation options for now :) Bigger priority is getting the formatting stability fix checked in so that we can get a patch update out soon.  I'm going to make a PR in a sec with the text change plus a couple other things.

kapilmb:
Yes, of course, stability is more important! and I have been testing it in my branch since last weekend. But the fix doesn't work. The biggest problem is the irregular nature of the bug. It appears a bit hard to reproduce. I will send in a patch soon for that too!

daviwil:
Awesome, thanks for working on that!

PR #468, opened 2017-01-28
Add cwd to debug config snippet for inter session.

### LINKED ISSUE #467: Launch Interactive Session config should have cwd setting
Without it, you get dumped into VSCode's install folder which you rarely want.  I'm thinking it should default to `${workspaceRoot}`.

rkeithhill:
Fix #467


daviwil:
Thanks Keith!

PR #473, opened 2017-01-30
Fixes some issues with code formatter

kapilmb:
Fixes the following issues:
* If the user switches the active editor when code formatter is running, the formatter can write to the new active editor.
* If the user executes "format document" on a document when the document is already being formatted, the extension creates a redundant request for formatting the same document. 
* If for any reason PSES crashes, then the formatting status is not cleaned up from the status bar.


daviwil:
Sweet, thanks!

kapilmb:
Thanks @daviwil!

PR #484, opened 2017-02-04
Find module display quick pick list first.

rkeithhill:
This PR puts the quick pick list up first with "Cancel" as the only item and text that indicates what is happening and that it could take some time.

@daviwil - please check my use of the CancellationTokenSource - especially in the pickPowerShellModule function where the request completes but the user (or timeout) has already cancelled the op.  Not sure about returning Promise.resolve("").   My JavaScript FU is pretty weak.  Also feel free to help with the wording of the various messages in the UI.

Finally what timeout should we use? I set it for 1 minute but that might be long.   It seems to take ~20 seconds on my machine.  But as the gallery grows, so could that time.

Finally, for the future, it would be nice if we could detect a user key press and if the find isn't done - cancel and restart a new find based on what the user typed.  Also, need timer support so that you only kick off a new find after say 1 second of paused typing.

### DIFF
--- a/src/features/PowerShellFindModule.ts
+++ b/src/features/PowerShellFindModule.ts
@@ -20,28 +20,35 @@ export class FindModuleFeature implements IFeature {
 
     private command: vscode.Disposable;
     private languageClient: LanguageClient;
+    private cancelFindToken: vscode.CancellationTokenSource;
 
     constructor() {
         this.command = vscode.commands.registerCommand('PowerShell.PowerShellFindModule', () => {
-            var items: QuickPickItem[] = [];
+            // It takes a while to get the list of PowerShell modules, display some UI to let user know
+            this.cancelFindToken = new vscode.CancellationTokenSource();
+            vscode.window
+                .showQuickPick(
+                ["Cancel"],
+                { placeHolder: "Please wait, retrieving list of PowerShell modules. This can take some time..." },
+                this.cancelFindToken.token)
+                .then(response => { if (response === "Cancel") { this.clearCancelFindToken(); } });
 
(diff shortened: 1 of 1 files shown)


src/features/PowerShellFindModule.ts
```diff
@@ -20,28 +20,35 @@ export class FindModuleFeature implements IFeature {
...
-                            var moduleName = selection.label;
-                            //vscode.window.setStatusBarMessage("Installing PowerShell Module " + moduleName, 1500);
-                            this.languageClient.sendRequest(InstallModuleRequest.type, moduleName);
-                        }
-                    });
-                });
+            this.pickPowerShellModule().then((moduleName) => {
+                if (moduleName) {
+                    // vscode.window.setStatusBarMessage("Installing PowerShell Module " + moduleName, 1500);
+                    this.languageClient.sendRequest(InstallModuleRequest.type, moduleName);
```
rkeithhill: I left this comment in.  I think the thought was it would be nice to display this status bar message but the extension, should probably receive a message back from PSES that the module has been installed.  Then it would display an "Install complete" message for say 5 seconds before being cleared.

daviwil:
I'll take a look at this later today, I'm making a lot of progress on getting the interactive console working so I'm trying to keep moving on that :)

rkeithhill:
No problem.  I was watching Joey demo this feature on a PluralSight video and not having UI come up for 20-30 seconds was a bit painful to watch.  :-)

daviwil:
This is one of those cases where I just cringe at the continual spread of a bad pattern that I established ;)

PR #485, opened 2017-02-06
Add params to DebugTest.ps1.

rkeithhill:
This will align with part 2 of my debugging blog post for Scripting Guys.  This allows testing of launch  config that passes arguments and makes it easier to demo "attach to" but setting the delay a bit longer (say 2 secs).  Also, updated  launch.json to start with the default configuration before add tle Launch Script Configuration.

### DIFF
--- a/examples/DebugTest.ps1
+++ b/examples/DebugTest.ps1
@@ -1,3 +1,5 @@
+param([int]$Count=50, [int]$DelayMilliseconds=200)
+
 function Write-Item($itemCount) {
     $i = 1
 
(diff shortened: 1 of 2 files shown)


examples/DebugTest.ps1
```diff
@@ -23,4 +25,6 @@ function Do-Work($workCount) {
     Write-Host "Done!"
 }
 
-Do-Work 50
+$process = Get-Process -Id $pid
```
daviwil: Where's the $pid defined?

rkeithhill: Uh, sorry about that.  Left a little something in there I didn't intend.  :-(

daviwil: No problem ;)

PR #489, opened 2017-02-06
Enable code formatter to enforce consistent whitespace style 

### LINKED ISSUE #446: Script formatting should allow setting for space between keyword and opening paren
For instance, I prefer to see:
```powershell
if (Test-Path foo) {
   "Foo exists!"
}
```
instead of
```powershell
if(Test-Path foo) {
   "Foo exists!"
}
```
Same would apply to other keywords: `elseif`, `for`, `foreach`, `while`, `switch`, etc.

kapilmb:
Resolves #446.

### DIFF
--- a/src/features/DocumentFormatter.ts
+++ b/src/features/DocumentFormatter.ts
@@ -335,9 +335,11 @@ class PSDocumentFormattingEditProvider implements DocumentFormattingEditProvider
                 break;
 
             case "PSUseWhitespace":
-                ruleSettings["CheckOpenBrace"] = true;
-                ruleSettings["CheckOpenParen"] = true;
-                ruleSettings["CheckOperator"] = true;
+                ruleSettings["CheckOpenBrace"] = psSettings.codeFormatting.whitespaceBeforeOpenBrace;
+                ruleSettings["CheckOpenParen"] = psSettings.codeFormatting.whitespaceBeforeOpenParen;
+                ruleSettings["CheckOperator"] = psSettings.codeFormatting.whitespaceAroundOperator;
+                ruleSettings["CheckSeparator"] = psSettings.codeFormatting.whitespaceAfterSeparator;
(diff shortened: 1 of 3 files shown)


src/features/DocumentFormatter.ts
```diff
@@ -139,6 +139,7 @@ class PSDocumentFormattingEditProvider implements DocumentFormattingEditProvider
     private readonly ruleOrder: string[] = [
         "PSPlaceCloseBrace",
         "PSPlaceOpenBrace",
+        "PSUseWhitespace",
```
daviwil: Should this be named PSUseConsistentWhitespace?

kapilmb: I was in a bit of a limbo on whether to have "consistent" in its name. But looks like "PSUseConsistentWhitespace" sounds more consistent with the "PSUseConsistentIndentation" rule. On the other hand all the rules enforce some or the other kind of consistency and so having the "consistent" word seemed verbose (And I was even thinking of changing "PSUseConsistentIndentation" to "PSUseIndentation", but that would be a breaking change for PSSA). At this point, I personally don't have any preference but if you think "PSUseConsistentWhitespace" sounds better, I can change it to that.

daviwil: I think consistency would be nice between the two names.  "PSUseIndentation/Whitespace" doesn't sound specific enough to me; the "Consistent" designation helps with that.  I've tried thinking of other names and I can't come up with anything better at the moment.

kapilmb: Fixed!

Also made corresponding changes to PowerShell/PSScriptAnalyzer#702

daviwil: Thanks man!

kapilmb:
@daviwil Thanks!

PR #502, opened 2017-02-11
Fixed missing quote, clarified comment

pcgeek86:


pcgeek86:
@daviwil I'll never understand how you're so fast.

daviwil:
It's easy while sitting on the couch with my phone ;)

PR #512, opened 2017-02-22
Add option to place new line after close brace

kapilmb:


kapilmb:
This adds an option to place new line after close brace. The following setting will now allow `cuddled` else. 

```
"powershell.codeFormatting.openBraceOnSameLine": true
"powershell.codeFormatting.newLineAfterCloseBrace": false
```

```powershell
if (...) {
...
} else {
...
}
```

One thing to note is that this setting only allows a `cuddled` else, but it doesn't enforce it, i.e., if a user already has else statements written in a `cuddled` manner, then the formatter will leave it as is. But, if a user has the statements written in the following form, then the formatter won't change it to the `cuddled` style.

```powershell
if (...) {
...
} 
else {
...
}
```

daviwil:
Great, thanks Kapil!

PR #523, opened 2017-03-05
Fix issue with VSCode var syntax change . to :

rkeithhill:
BTW I'm not 100% sure about the need to require vscode 1.10.0. I've reached out to Andre (cc'd you) to see what alternatives we have.

Also replaced the deprecated OutDir setting with OutFiles in the extension's launch.json file.

FWIW I do get errors attaching to a PowerShell host process with VSCode 1.10 that I wasn't getting with 1.9 using 0.9.0 of the extension.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,12 +9,13 @@
 			"args": [ "--extensionDevelopmentPath=${workspaceRoot}" ],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outDir": "${workspaceRoot}/out",
+			"outFiles": ["${workspaceRoot}/out"],
 			"preLaunchTask": "Build"
 		},
 		{
 			"name": "Attach",
 			"type": "node",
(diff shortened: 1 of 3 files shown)


rkeithhill:
It looks like this was a VSCode regression that should be fixed by a recovery build for 1.10 - https://github.com/Microsoft/vscode/issues/21894  So we can wait on this one until such time as we're comfortable requiring VSCode 1.10 or higher.

daviwil:
Yep, I think 1.10.2 should be out soon enough that we won't need to fix this immediately.  If it takes longer we can ship a patch update.

gerane:
This won't fix existing projects that have the old syntax will it?

rkeithhill:
No. But there is supposed to be an update to VSCode that fixes this i.e. you will see the syntax marked as deprecated in your launch.json but it will work.

daviwil:
Took me way too long to merge this :)  Thanks!

PR #537, opened 2017-03-14
Release 0.10.0

daviwil:


daviwil:
May take another hour or two before this gets released, need to sign binaries, etc.

daviwil:
Screenshots for changelog:

image

image

image

image

PR #590, opened 2017-03-20
Add new integrated console settings

daviwil:
See commits for more information.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,10 @@ branches:
   - master
   - develop
 
+environment:
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true  # Don't download unneeded packages
+  DOTNET_CLI_TELEMETRY_OPTOUT: true        # Don't send telemetry
+
 install:
   - git clone https://github.com/PowerShell/PowerShellEditorServices.git ../PowerShellEditorServices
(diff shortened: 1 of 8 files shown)


daviwil:
@rkeithhill What do you think about the name of the `powershell.startAutomatically` setting?  Any better ideas?

rkeithhill:
So this really is whether to actually start PSES?  Is there no way to start the extension but not automatically show the IC?

daviwil:
@rkeithhill There are two separate settings here, one is to automatically start PSES and the other is to automatically show the console when PSES starts.  In the original issue that was filed, the author asked if we could add a setting to defer starting PSES until later.

rkeithhill:
Not starting PSES seems a bit strange given it occurs in the background (spinning up a different process).  I wouldn't think it would impact editing performance that much.  The name seems reasonable to me.  

I wonder how long it will be before we get an issue that the extension isn't working because someone set this setting to true?  😛

daviwil:
Starting up PSES does have some impact on slower machines just due to another process loading and taking up CPU cycles, but I suppose it isn't happening in the foreground so it isn't blocking the user from using the editor.  I'll leave the setting in for now even though there's a risk of someone shooting off their foot ;)

PR #607, opened 2017-03-23
Update ISSUE_TEMPLATE.md

pcgeek86:
Added VSCode diagnostics to GitHub issue template

PR #610, opened 2017-03-23
Update Issue Template

mattmcnabb:
Now that we have $PSEditor in the terminal, we can run a command to return the PowerShell extension version for issues. Look like a good idea?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -17,9 +17,10 @@ https://github.com/PowerShell/EditorSyntax/issues
 - Output from `$PSVersionTable`:
 
 ```
-Copy / paste the below commands into PowerShell, and paste the output here
+Copy / paste the below commands into the PowerShell Integrated Terminal, and paste the output here
 
 code -v
+$pseditor.EditorServicesVersion
 code --list-extensions --show-versions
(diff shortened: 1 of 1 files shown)


rkeithhill:
Hehe, take a look at https://github.com/PowerShell/PowerShellEditorServices/pull/404  After a bit more thought, I'm beginning to think that a VSCode command is the way to go.  `PowerShell: Copy Version Info to Clipboard` or something like that.  Also, David just merged in a PR from Trevor that addresses this as well https://github.com/PowerShell/vscode-powershell/pull/607.

daviwil:
I think Keith's approach definitely makes the user experience easier, but I'll go ahead and accept Matt's PR because it's useful information :)

mattmcnabb:
@rkeithhill Yeah I think I was piggybacking on Trevor's PR but added the $PSEditor command. I like the idea of a simple command that gathers all the info as well, could probably work on that.

Also, does it make sense to try to gather the OS name and version as well, or should we just leave that up to the user? Just can't think of a simple way to grab the OS Edition other than WMI, but version should be relatively easy.

rkeithhill:
@mattmcnabb  Look at that PSES PR (404) I linked to above.  That command does grab OS info as well.

PR #636, opened 2017-03-30
Fix hang caused when format on type is enabled

### LINKED ISSUE #595: Powershell extension consumes full core of CPU when formatting
### System Details

- Operating system name and version: Windows 10
- VS Code version: 1.10.2
- PowerShell extension version: 0.10.1

```
CLRVersion
BuildVersion                   3.0.0.0
PSVersion                      6.0.0-alpha
GitCommitId                    v6.0.0-alpha.15
```

### Issue Description

If I open a powershell script and make a few edits, I get a message "Formatting PowerShell document..." spinning forever in the footer of VS Code

image

Meanwhile, the powershell process which is powering the editor features consumes a cull core of CPU:

image

### Attached Logs

`EditorServices.log` is over 300MB at this point, so hopefully something relevant in there. Do you have a suggested way to share? Or maybe I can just let it run for a few seconds to generate 5-10MB of logs.

kapilmb:
The formatter would go into an infinite loop in certain cases because formatting edits were not filtered in the right way. 

Fixes #595

### DIFF
--- a/src/features/DocumentFormatter.ts
+++ b/src/features/DocumentFormatter.ts
@@ -345,7 +345,7 @@ class PSDocumentFormattingEditProvider implements
                     // we do not return a valid array because our text edits
                     // need to be executed in a particular order and it is
                     // easier if we perform the edits ourselves
-                    return this.applyEdit(editor, uniqueEdits, range, 0, index);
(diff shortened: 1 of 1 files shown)


daviwil:
Looks good, thanks man!

kapilmb:
Thanks @daviwil !

PR #638, opened 2017-03-31
Enable debugging of untitled script files

### LINKED ISSUE #555: Enable debugging of untitled script files
VS Code will soon be gaining the ability to launch the debugger on untitled, unsaved files.  Since this is a scenario that works in the ISE, we should leverage VS Code's new feature to provide it in the PowerShell extension too.

https://github.com/Microsoft/vscode/issues/19415

/cc @dfinke

daviwil:
This change enables untitled files to be debugged when they are set to the
'powershell' language mode.

Fixes #555.

### DIFF
--- a/examples/.vscode/settings.json
+++ b/examples/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     // Use a custom PowerShell Script Analyzer settings file for this workspace.
     // Relative paths for this setting are always relative to the workspace root dir.
-    "powershell.scriptAnalysis.settingsPath": "./PSScriptAnalyzerSettings.psd1"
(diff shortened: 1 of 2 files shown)


examples/.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     // Use a custom PowerShell Script Analyzer settings file for this workspace.
     // Relative paths for this setting are always relative to the workspace root dir.
-    "powershell.scriptAnalysis.settingsPath": "./PSScriptAnalyzerSettings.psd1"
+    "powershell.scriptAnalysis.settingsPath": "./PSScriptAnalyzerSettings.psd1",
+    "files.defaultLanguage": "powershell"
```
kapilmb: Since vscode-powershell is typescript project, should defaultLanguage be `typescript`?

PR #642, opened 2017-04-01
Fix detection of valid PS ext for debugging.

### LINKED ISSUE #641: Error message "is a file type that cannot be debugged by the powershell debug" with capital file extension .PS1
### System Details

- Operating system name and version:  Windows 10
- VS Code version: 1.10.2
- PowerShell extension version: 0.11.0
PSVersion                      5.1.14393.953
BuildVersion                   10.0.14393.953

```
Copy / paste the below commands into the PowerShell Integrated Terminal, and paste the output here

code -v
$pseditor.EditorServicesVersion
code --list-extensions --show-versions
$PSVersionTable
```
code -v
1.10.2
8076a19fdcab7e1fc1707952d652f0bb6c6db331
PS D:\> $pseditor.EditorServicesVersion
>>
PS D:\> code --list-extensions --show-versions
>>
Adrien.VisualStudioDarkTheme@1.0.1
alefragnani.Bookmarks@0.12.0
emilast.LogFileHighlighter@1.1.1
emmanuelbeziat.vscode-great-icons@1.1.47
IBM.output-colorizer@0.1.0
ms-vscode.cpptools@0.10.5
ms-vscode.PowerShell@0.11.0
ms-vscode.wordcount@0.1.0
msazurermtools.azurerm-vscode-tools@0.3.4
robertohuertasm.vscode-icons@7.4.0
slevesque.vscode-hexdump@1.2.1

PS D:> $PSVersionTable
>>

PSVersion                      5.1.14393.953
BuildVersion                   10.0.14393.953
### Issue Description

When the file extension is capital, it reports error like:
"is a file type that cannot be debugged by the powershell debug"

### Attached Logs

Follow the instructions in the [README](https://github.com/PowerShell/vscode-powershell#reporting-problems)
about capturing and sending logs.

rkeithhill:
Was previously only allowing lower case ps1 and psm1.  I'm assuming that on *nix file systems, that PowerShell allows all variations of ps1,PS1,pS1,Ps1 as valid script extensions.

Fix #641 

@daviwil Currently, debugging an "untitled" script file doesn't work even if you set the language to PowerShell.  That's because this code below detects "isUntitled" and sets the `script` property to `untitled:Untitled-1`.  Seems like you were working on enabling debug of untitled/unsaved script files.  Is that coming later or should the below be working now?

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -44,7 +44,15 @@ export class DebugSessionFeature implements IFeature {
                 let currentDocument = vscode.window.activeTextEditor.document;
 
                 if (currentDocument.isUntitled) {
-                    config.script = currentDocument.uri.toString();
+                    if (currentDocument.languageId === 'powershell') {
+                        config.script = currentDocument.uri.toString();
+                    }
+                    else {
+                        let msg = "In order to debug '" + currentDocument.fileName +
+                                  "', set the document's language mode to PowerShell or save the file with a PowerShell extension.";
+                        vscode.window.showErrorMessage(msg);
+                        return;
+                    }
                 }
                 else {
                     let isValidExtension = false;
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -48,7 +48,7 @@ export class DebugSessionFeature implements IFeature {
                         config.script = currentDocument.uri.toString();
                     }
                     else {
-                        let msg = "In order to debug '" + currentDocument.fileName +
+                        let msg = "To debug '" + currentDocument.fileName +
                                   "', change the document's language mode to PowerShell or save the file with a PowerShell extension.";
                         vscode.window.showErrorMessage(msg);
                         return;
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -49,7 +49,7 @@ export class DebugSessionFeature implements IFeature {
                     }
                     else {
                         let msg = "In order to debug '" + currentDocument.fileName +
-                                  "', set the document's language mode to PowerShell or save the file with a PowerShell extension.";
+                                  "', change the document's language mode to PowerShell or save the file with a PowerShell extension.";
                         vscode.window.showErrorMessage(msg);
                         return;
                     }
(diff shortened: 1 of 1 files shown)


daviwil:
@rkeithhill have you updated your PSES?  I did the debug adapter work already to enable this.  If it's not working after updating that I may need to take a look.  Has been working fine for me though.

daviwil:
Setting the `script` property to the URI is intentional, the language server already has the contents of a file with that URI so I detect that in the debug adapter and load the file contents from the current workspace.

daviwil:
Actually, approved too soon.  I think we need to ensure the language mode is still `powershell` when the file is untitled, right?  Not sure if the debug adapter client will still send us requests for untitled files that aren't marked as `powershell`.

rkeithhill:
Yeah I was wondering about that.  You made that change to VSCode but I figured that wouldn't be released for a while.  So the assumption was that this worked without having to set the language mode.  So, let me check my understanding.  You can debug an unsaved file but for now you have to set the language mode to PowerShell.  Is that right?

rkeithhill:
What do you think?
image
BTW this makes sense but I want to double check.  If the file is not saved you can run it from the debugger and break into the debugger and single step but you can't set breakpoints.  Is that right?

rkeithhill:
Is this too verbose?
image

daviwil:
Hmm... Yeah, good point.  I suppose if they've already selected the PowerShell debug configuration for Current File and they're hitting F5 on an untitled file that isn't marked `powershell` then presumably they'd want to debug it as PowerShell.  However, if the file hasn't been sent to the language server already (which it wouldn't have if it isn't marked as `powershell` then we won't be able to debug it.  I think having the error there makes sense.

Yeah, you won't be able to set breakpoints at this time with the untitled script, just stepping through.

daviwil:
Might be good to get @juneb's input on the error message string, she's been fixing up a lot of my bad verbiage :)

rkeithhill:
Sounds good to me. I just changed the "set the document's language mode" to "change the document's language mode" to better match the command entry in the command palette "change language mode".

juneb:
Thanks, @daviwil! Change is much better than set, which is ambiguous, even to native English speakers.

To shorten that error message, change "In order to" to "To". 

To debug '<filename>', change the document language mode to PowerShell or save the file with a PowerShell extension.

rkeithhill:
OK, updated to:
image
Thanks @juneb

daviwil:
Looks good, thanks June!

PR #658, opened 2017-04-06
Fix issue with dbg launch of script 'Invoke-Pester'.

rkeithhill:
Did some basic sanity testing on this and seems to work in various test cases.


daviwil:
Just tested this out, looks good man, thanks!

PR #690, opened 2017-04-20
Change initialization order when starting language client

kapilmb:
Using the newer version of language client we cannot add some handlers unless the client server is ready.


daviwil:
Looks good!

Issue #705, opened 2017-04-29
Run Script in External PowerShell Console, accepting parameter


mpearon:
### System Details

- Operating system name and version:  **Windows 10 v1607 Build 14939.1066**
- VS Code version:  **1.11.2**
- PowerShell extension version:  **0.12.2**
- Output from `$PSVersionTable`:
                                                                                                            
**PSVersion                      5.1.14393.1066**
**PSEdition                      Desktop**
**PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}**
**BuildVersion                   10.0.14393.1066**
**CLRVersion                     4.0.30319.42000**
**WSManStackVersion              3.0**
**PSRemotingProtocolVersion      2.3**
**SerializationVersion           1.1.0.1**

### Issue Description

The ability to run a script in an external console from within vsCode would be awesome.
This feature was present in PowerGUI, but broke with the release of Win10 v1607 - and I don't have a machine pre-1607 around, so I can't demonstrate the feature in its entirety...but what I can show follows.

There were two ways to execute a script, and a way to pass a parameter to both methods:
     1.  Running in integrated terminal
          This did just want vsCode does, but it offered an extra feature that I'll mention in No. 3.
![powergui-1](https://cloud.githubusercontent.com/assets/7157790/25558805/eaa4da50-2cf3-11e7-9a2d-eebcaeb2b985.gif)

     2.  Running in external console
          I can't demonstrate this one (afore mentioned reasons), but this is what I would love to see.  This saved a temporary copy of the script, then launched PowerShell externally with the path to the temporary script as the -File parameter.  This helped eliminate any "ISE-style" weirdness, and showed me exactly how the end-user's experience would be - which was awesome.

     3.  You could pass an argument list to both methods via an input box in editor.
![powergui-3](https://cloud.githubusercontent.com/assets/7157790/25558821/47a6abca-2cf4-11e7-892c-95edc83238c0.gif)

mpearon:
I'm not sure why, but the descriptions on points 2 and 3 are showing as code.  Sorry!

rkeithhill:
When you run in the integrated terminal you *can* pass parameters to this script.  Click on the gear icon in the debug view to edit the launch.json file:
```
        {
            "type": "PowerShell",
            "request": "launch",
            "name": "PowerShell Launch (current file)",
            "program": "${file}",
            "args": [],
            "cwd": "${file}"
        },
```
[...]

PR #707, opened 2017-04-30
Add optional UI to prompt for script args.

rkeithhill:
Address one issue #705

### DIFF
--- a/examples/.vscode/launch.json
+++ b/examples/.vscode/launch.json
@@ -4,7 +4,7 @@
         {
             "type": "PowerShell",
             "request": "launch",
-            "name": "PS Launch - Current File",
+            "name": "PowerShell Launch Current File",
             "script": "${file}",
             "args": [],
             "cwd": "${file}"
@@ -12,7 +12,7 @@
         {
             "type": "PowerShell",
             "request": "launch",
-            "name": "PS Launch - Current File w/Args Prompt",
+            "name": "PowerShell Launch Current File w/Args Prompt",
             "script": "${file}",
             "args": [ "${command:SpecifyScriptArgs}" ],
             "cwd": "${file}"
@@ -20,7 +20,7 @@
         {
             "type": "PowerShell",
             "request": "launch",
-            "name": "PS Launch - DebugTest.ps1",
+            "name": "PowerShell Launch DebugTest.ps1",
             "script": "${workspaceRoot}/DebugTest.ps1",
             "args": ["-Count 55 -DelayMilliseconds 250"],
             "cwd": "${workspaceRoot}"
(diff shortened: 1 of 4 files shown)


rkeithhill:
This is not quite ready yet.  I'm waiting on some resolution to this issue - https://github.com/Microsoft/vscode/issues/25712

daviwil:
Awesome!  I suppose we'll wait for the next VS Code update before shipping this in the PS extension?

rkeithhill:
Yes.  I keep testing daily builds of VSCode Insiders but haven't observed the fixed behavior yet.

rkeithhill:
OK, the issue is fixed in today's insiders build.  It should ship in the (late) May release.

rkeithhill:
One more thing to do. Need to figure out a debug config snippet for this.  I'll propose one here in a bit.

rkeithhill:
Here is what the Debug configuration list looks like now (from the Examples folder):
image

This is what the Add Configuration choices look like:
image

rkeithhill:
> I suppose we'll wait for the next VS Code update before shipping this in the PS extension?

We can ship this before the May VSCode release drops.  I added code to check the VSCode version and I only enable the "save the prev args" features if we're on 1.13 or later.

rkeithhill:
I'm not sure I like the args being saved across separate VSCode sessions.  Let me see if I can find a way to store the args just for the current session.

daviwil:
Using "PS" as the prefix just feels wrong to me even though I can understand your reason for doing it.  As far as I know, nobody has complained about not being able to see the name in the debug UI.  Maybe we should leave it the way it is so that it's clearer?

I suppose there's also the possibility of rewording them slightly.  Instead of "PowerShell Launch Current File" it could be "Launch Current PowerShell File", "Interactive PowerShell Session" instead of "PowerShell Interactive Session", etc.  What do you think?

rkeithhill:
OK putting PS back to PowerShell.  I think it is better to leave PowerShell first because so that in the mixed debug configuration scenario, the PowerShell configs sort next to one another.  PR coming.

daviwil:
Thanks Keith!  This change is pretty awesome, I like the new feature a lot.  Is it ready to be merged now?

rkeithhill:
I think so.  Do you think it is OK to remember scripts args between sessions using workspaceState?

daviwil:
Yeah, I think that's a great way to do it.  We might need a way to clear the stored args in the future, but I wouldn't worry about it until someone asks for it.  Merging it now, thanks man!

rkeithhill:
I think this will make for a nice little debugging enhancement.

PR #716, opened 2017-05-07
Add Install-VSCode.ps1 script

daviwil:
This change adds a new script that can be used to install Visual Studio
Code and the PowerShell extension.  The intent is for this script to be
shipped to the PowerShell Gallery so that one can easily get the script
and install VS Code in a single line of PowerShell.

### DIFF
--- a/scripts/Install-VSCode.ps1
+++ b/scripts/Install-VSCode.ps1
@@ -1,0 +1,111 @@
+<#
+.SYNOPSIS
+    Installs Visual Studio Code, the PowerShell extension, and optionally
+    a list of additional extensions.
+
+.DESCRIPTION
+    This script can be used to easily install Visual Studio Code and the
+    PowerShell extension on your machine.  You may also specify additional
+    extensions to be installed using the -AdditionalExtensions parameter.
+    The -LaunchWhenDone parameter will cause VS Code to be launched as
+    soon as installation has completed.
+
+    Please contribute improvements to this script on GitHub!
+
+    https://github.com/PowerShell/vscode-powershell/blob/develop/scripts/Install-VSCode.ps1
+
+.PARAMETER AdditionalExtensions
+    An array of strings that are the fully-qualified names of extensions to be
+    installed in addition to the PowerShell extension.  The fully qualified
+    name is formatted as "<publisher name>.<extension name>" and can be found
+    next to the extension's name in the details tab that appears when you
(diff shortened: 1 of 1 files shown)


daviwil:
Would definitely appreciate some feedback on the quality of this script!

dfinke:
Uninstalled both the current and insiders vs code versions. Ran the script, worked like a charm. I'll try again and make sure the PSES ext is not installed too.

image

scripts/Install-VSCode.ps1
```diff
@@ -0,0 +1,111 @@
...
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+#>
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory=$false)]
```
rkeithhill: I usually don't bother specifying Mandatory=$false since that is the default.  Other than that LGTM.

daviwil: Ahhh yep, thanks!

daviwil:
Thanks guys!

adamrushuk:
Although this worked initially, I tried uninstalling the PowerShell Extension **before** uninstalling VSCode, and this time got this error which left VSCode installed without the PowerShell Extension:
image

daviwil:
Argh, yeah, VS Code has this weird thing where they need to update their internal extension metadata by restarting the editor.  If you uninstall the extension and close the editor then try to reinstall the extension via command line, VS Code throws up.  Not sure if there's a way around that yet, but I'll see what I can figure out.

Thanks for letting me know!

adamrushuk:
Ah that was my bad then as I didn't restart code before uninstalling during the test.
It worked as expected once I did the above, but struggled to delete the previously downloaded "vscode-stable.exe" file in $env:TEMP. That looks like a known issue though, so I manually deleted before the successful test shown below:

image

PR #757, opened 2017-05-17
Debugger is now activated when breakpoints are hit in Integrated Console

### LINKED ISSUE #619: Hitting a breakpoint in the console while not debugging should activate the VS Code debugger
Desired behavior:

1. Open a .ps1 file and set breakpoints using `Set-PSBreakpoint -Script <script> -Line <##>`
2. Run the script from the console by invoking with its script name
3. PowerShell hits the breakpoint, activates VS Code's debugger, then shows the breakpoint hit at that line

This should also work for command and variable breakpoints.

daviwil:
This change adds new behavior which causes VS Code's debugger UI to be
activated when a breakpoint is hit in the Integrated Console while not
currently in debug mode.  This allows the user to set breakpoints and
run scripts from the console while still leveraging VS Code's rich
debugging UI.

Resolves #619.

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -8,6 +8,10 @@ import { IFeature } from '../feature';
 import { SessionManager } from '../session';
 import { LanguageClient, RequestType, NotificationType } from 'vscode-languageclient';
 
+export namespace StartDebuggerNotification {
+    export const type = new NotificationType<void, void>('powerShell/startDebugger');
+}
+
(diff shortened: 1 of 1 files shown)


daviwil:
Related to https://github.com/PowerShell/PowerShellEditorServices/pull/455

rkeithhill:
Surprisingly little code change on this side.  Nice!

daviwil:
Yep!  It was pretty easy to get working thankfully.

PR #780, opened 2017-05-24
Enable Travis CI with Ubuntu and MacOS builds

PowerSchill:
- Set to Trusty Tahr (Ubuntu 14.04)
- Set to xcode 8.3 (OS X 10.12)
- Downloads and installs PowerShell v6.0.0-beta.1


daviwil:
Fantastic, thanks a lot Mark!

Issue #781, opened 2017-05-25
Markers from PSScriptAnalyzer should include the rule name


Josverl:
I find the integration of scrip analyzer very helpful, 
however sometimes there is a need to suppress a specific warning or error.

That can be done by adding decorating the code with something like 
`[Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSProvideCommentHelp", "")]`

the key here is that I need to look up the PSSA RuleName, that is not easily referenced searchable based on the information provided. If 
- the Rulename could be provided as part of the message , 
- if sample suppress block could be copied / pasted  ( either from the psscript analyzer results, or as a snippit ) 

That would go a long way. 
currently I often need retrieve the rule names from a build task that runs after check-in or look up the rule name by trial and error.
I would prefer to be able to do this quickly at edit time.

daviwil:
Hey Jos, thanks for the feedback!  Issue #778 might be interesting to you, we plan to add "quick fix" options for adding rule suppressions so that you don't need to know exactly what to type for that.

I agree though, we definitely need to display the PSScriptAnalyzer rule name in the marker message so that it's easy to know what needs to be disabled.  We'll add that!

PR #783, opened 2017-05-25
Add 2 snippets to simplify supressing benign PSScriptAnalyzer errors …

Josverl:
in reference to issue #781 I have created 2 snippets that should be broadly applicable to users that use PSSA to guard the code quality.

This combined with the display of the PSSA rulename as part of the warning in the Problem pane will allow simplify the proper use of PSSA.


daviwil:
Thanks for the PR @Josverl!  Sorry for the delay in getting it accepted.

PR #793, opened 2017-05-30
Fix a handful of issues with PowerShell sessions

daviwil:


scripts/Start-EditorServices.ps1
```diff
@@ -204,203 +204,73 @@ if ((Test-ModuleAvailable "PowerShellEditorServices" -RequiredVersion $parsedVer
...
     $errorString = ""
 
     while ($e -ne $null) {
         $errorString = $errorString + ($e.Message + "`r`n" + $e.StackTrace + "`r`n")
         $e = $e.InnerException;
     }
 
-    ExitWithError ("Caught error while waiting for EditorServicesHost to complete:`r`n" + $errorString)
+    ExitWithError ("A terminating error occurred in PowerShell Editor Services:`r`n`r`n" + $errorString)
 }
```
rkeithhill: Good call on eliminating the signature for this script.  I've had several issues with folks using PSCX after we started signing the scripts.

daviwil: This was only temporary since the signature isn't useful after the script contents have changed.  Why did signing the startup script cause issues for PSCX?  Maybe it's an indication of some other issue that I haven't seen before

rkeithhill: Folks that were offline ran into an apparent hang as the CRL check takes several minutes to come back IIRC.

daviwil: Hmmm...  That could explain a complaint I've been seeing from some folks who report that the extension doesn't start up for them if they're offline due to internet not being accessible at the moment (like on a train).  Was it a combination of PSCX and the signed script that caused it, or just the signed script itself?

PR #794, opened 2017-05-30
Display installed PowerShell Core builds in session selection menu

### LINKED ISSUE #671: Provide menu to select an installed PowerShell Core from default system path
We need a way to enable the user to easily select an installation of PowerShell Core on their machine to use for their PowerShell extension session.  This needs to use the default installation paths for PowerShell Core on all supported platforms.

/cc @mgreenegit

rkeithhill:
Be sure to check over my weak (but improving) Node / JavaScript.

Resolves #671.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -604,6 +604,21 @@ export class SessionManager {
         }
     }
 
+    private getPowerShellCorePaths(): string[] {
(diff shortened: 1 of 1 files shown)


daviwil:
Looks great man, thanks a lot!

PR #796, opened 2017-05-31
Small tweaks and fixes

daviwil:


src/session.ts
```diff
@@ -734,12 +734,16 @@ export class SessionManager {
 
             var pscorePaths = this.getPowerShellCorePaths();
             for (var pscorePath of pscorePaths) {
-                var pscoreVersion = path.parse(pscorePath).base;
-                var pscoreExePath = path.join(pscorePath, "powershell.exe");
-                var pscoreItem = new SessionMenuItem(
+                let pscoreVersion = path.parse(pscorePath).base;
```
rkeithhill: I've seen this preference for `let` over `var` - oh yeah block scope.  Can we use `const` instead of `let` for non-changing variables?

daviwil: I believe so, yeah.  Not sure if it also solves the variable closure issue, but I assume it should.

PR #798, opened 2017-05-31
Add Pester problem matcher to pkg,json.

rkeithhill:
Should this matcher be called `$pesterError` or just `$pester`?


daviwil:
`$pesterError` works for me, thanks man!

daviwil:
Actually, maybe I'll tweak it to be $pester since it seems that other problem matchers just use the name of the tool being matched.

rkeithhill:
Works for me.

PR #841, opened 2017-06-07
Add ISE theme

mattmcnabb:
In the interest of time I'm submitting a first draft of the PowerShell ISE theme. Let's hope I did this right ;)

Some issues
- can't color the '$' to match the variable name - think this follows the operator key
- object members and commands both use the "functions" language syntax key
- builtin variables like $VerbosePreference are colored green instead of using the variable color orange

It's close but not exactly like the original

### DIFF
--- a/package.json
+++ b/package.json
@@ -461,7 +461,14 @@
           "description": "Switches focus to the console when a script selection is run or a script file is debugged. This is an accessibility feature. To disable it, set to false."
         }
       }
-    }
+    },
+    "themes": [
+      {
+        "label": "PowerShell ISE",
+        "uiTheme": "vs",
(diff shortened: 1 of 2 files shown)


daviwil:
Woohoo, awesome!  Don't worry about the syntax coloring issues just yet, that's a problem with our language grammar.  Thanks so much for doing this, you will be getting super huge credit in the release tomorrow!

mattmcnabb:
@daviwil I'm aware these are problems with the editor syntax - just wanted to record then for posterity in case those problems are ever solved. Then we can revisit this to see if the tokens can be sorted out.

Anyway, thanks! This was fun!

PR #844, opened 2017-06-08
Spelling mistake in DSC snippet

anwather:
It is just so tiny - provier changed to provider


daviwil:
Thanks a lot @anwather!

Issue #869, opened 2017-06-12
Never ending "Formatting Powershell document ..."


artisticcheese:
### System Details

- Operating system name and version:
- VS Code version: 1.13.0
- PowerShell extension version: 1.3.2 
- Output from `$PSVersionTable`:
```
Name                           Value                                                                                                                                
----                           -----                                                                                                                                
PSVersion                      5.1.15063.296                                                                                                                        
PSEdition                      Desktop                                                                                                                              
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                              
BuildVersion                   10.0.15063.296                                                                                                                       
CLRVersion                     4.0.30319.42000                                                                                                                      
WSManStackVersion              3.0                                                                                                                                  
PSRemotingProtocolVersion      2.3                                                                                                                                  
SerializationVersion           1.1.0.1                                                                                                                              
```

Copy / paste the below commands into the PowerShell Integrated Terminal, and paste the output here

code -v
$pseditor.EditorServicesVersion
code --list-extensions --show-versions
$PSVersionTable
```
PS C:\gd\Documents\s\docker\tools\vstsagent> code -v
1.13.0
376c52b955428d205459bea6619fc161fc8faacf
PS C:\gd\Documents\s\docker\tools\vstsagent> $pseditor.EditorServicesVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
1      3      2      0


PS C:\gd\Documents\s\docker\tools\vstsagent> code --list-extensions --show-versions
DotJoshJohnson.xml@1.8.0
mkloubert.vs-deploy@9.25.0
ms-vscode.csharp@1.10.0
ms-vscode.PowerShell@1.3.2
PeterJausovec.vscode-docker@0.0.16
```
### Issue Description

Seeing never ending formatting powershell document

![image](https://user-images.githubusercontent.com/13176884/27057417-c065d288-4f91-11e7-8411-8ad3f56756f3.png)

[1497305178-957d91d3-58fd-4e96-a470-66c2668669871497305175368.zip](https://github.com/PowerShell/vscode-powershell/files/1069687/1497305178-957d91d3-58fd-4e96-a470-66c2668669871497305175368.zip)
Sometimes even 2 of them

![image](https://user-images.githubusercontent.com/13176884/27057466-f3144d68-4f91-11e7-88e4-a1dce2ed4742.png)

daviwil:
[...]

PR #874, opened 2017-06-13
Partially move document formatting to server side

kapilmb:
This is one step closer towards moving document formatting completely to server side. The logic could not be moved completely because vscode doesn't appear to send updated values of `tabSize` and `insertSpaces` when sending `textDocument/formatting` or any other formatting related request. Hence, we register the providers on the client side and then manually send the `textDocument/rangeFormatting` request. The code is designed such that even `textDocument/formatting` and `textDocument/onTypeFormatting` requests are handled by `textDocument/rangeFormatting`. This is done to minimize code duplication. 

As and when the `tabSize` and `insertSpaces` issues get resolved, we will only need to enable the formatting capabilities on server and add a `textDocument/onTypeFormatting` handler to completely move formatting to server-side.

Related PRs:
- PowerShell/PowerShellEditorServices#516
- PowerShell/PSScriptAnalyzer#780

This and the related PRs should also help use resolve the following issues:
- #734 
- #869

### DIFF
--- a/src/features/DocumentFormatter.ts
+++ b/src/features/DocumentFormatter.ts
@@ -17,7 +17,13 @@ import {
     TextEditor,
     TextLine
 } from 'vscode';
-import { LanguageClient, RequestType, DocumentFormattingRequest } from 'vscode-languageclient';
+import {
+    LanguageClient,
+    RequestType,
+    DocumentFormattingRequest,
+    DocumentRangeFormattingParams,
+    DocumentRangeFormattingRequest
+} from 'vscode-languageclient';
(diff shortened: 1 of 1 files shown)


src/features/DocumentFormatter.ts
```diff
@@ -286,158 +247,12 @@ class PSDocumentFormattingEditProvider implements
         PSDocumentFormattingEditProvider.documentLocker.lock(document, unlockWhenDone);
     }
 
-    private executeRulesInOrder(
```
daviwil: So happy this is gone now!

PR #895, opened 2017-06-17
Replace hardcoded c:\windows with ${env:windir}.

rkeithhill:
We used to have ${env.windir} but then VSCode broke us so we switched back to C:\Windows.

It turns out that the VSCode folks changed the syntax to ${env:windir} - replacing the dot with a colon.

Here's the VSCode issue describing this: https://github.com/Microsoft/vscode/issues/28365#issuecomment-308231940

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -2,7 +2,7 @@
     "version": "2.0.0",
 
     "windows": {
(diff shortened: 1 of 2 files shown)


daviwil:
Ahhh, fantastic, thanks for looking into that!

Issue #917, opened 2017-06-22
Implement new vscode-command : PowerShell.InvokeRegisteredEditorCommand


kamilkosek:
Changes are very simple , but for now it´s enough to play around 😄 
[https://github.com/kamilkosek/vscode-powershell/tree/HTMLContentViewPlayground](https://github.com/kamilkosek/vscode-powershell/tree/HTMLContentViewPlayground)
[Test Script](https://gist.github.com/kamilkosek/c9e0c9851b50fc8198c38c836cabf656)
Note: This only works if the content view [accepts JavaScript ](https://github.com/kamilkosek/vscode-powershell/commit/d4cf59d14726c1b959f4d520244720ecaf820b44#diff-05b2b44f6b1b1bf1a151793831ca31dbR182)

daviwil:
Would you like to send the InvokeRegisteredEditorCommand as a PR too?  If you can leave the JavaScript enabling commit out then I'll add that part next week :)

kamilkosek:
Sure, but since I´m new to 'Contributing to public GitHub Repo´s' i need to figure out how to that commit out 😆

kamilkosek:
Maybe we will have someday a convenient way to achieve something like this: 
![20170623_162522](https://user-images.githubusercontent.com/3264948/27486464-f1cc1c04-5830-11e7-958b-553fbe74b09c.gif)

I ❤️ GIF´s

daviwil:
Love the GIF!  Really cool to see everything hooked up end to end.  We can definitely look for ways to make it easier to build tool UIs, the [PowerShellEditorServices.VSCode](https://github.com/PowerShell/PowerShellEditorServices/tree/master/module/PowerShellEditorServices.VSCode) module might be a good place for that.

Fixed with PR #919

PR #919, opened 2017-06-23
Implement InvokeRegisteredEditorCommand 

kamilkosek:
This PR registers a new vscode-command. The command accepts a command name as a parameter so it can be invoked through a link inside of a HTMLContentView ([see here](https://code.visualstudio.com/docs/extensionAPI/vscode-api-commands#_links))
For further information, take a look [here](https://github.com/PowerShell/vscode-powershell/issues/917)


daviwil:
Thanks a lot Kamil!

PR #941, opened 2017-07-06
Eliminate Pester duplicate failures in Problems view.

rkeithhill:
When tasks is set to 2.0.0, the tasks run in the terminal and in this case, Pester outputs a bunch of progress text to the terminal.  This results in duplicate failures being found by the problem matcher.

Work around this problem for now by eliminating the Pester progress output.  For more details see - https://github.com/Microsoft/vscode/issues/22065#issuecomment-313418250

### DIFF
--- a/examples/.vscode/tasks.json
+++ b/examples/.vscode/tasks.json
@@ -84,7 +84,7 @@
             "isTestCommand": true,
             "showOutput": "always",
             "args": [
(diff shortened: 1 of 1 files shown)


rkeithhill:
Note: we still have a problem if the terminal window is too narrow.  PowerShell will wrap the line that contains the line number and file path.  The current $pester problem matcher expects this info to be on a single line.  I'll look into updating the problem matcher but that might not be so easy.  Workaround for now is to make your terminal window wider (or use tasks 0.1.0).

daviwil:
Cool, thanks!

PR #944, opened 2017-07-11
Allow inclusion of JavaScript and CSS file paths in the HtmlContentView

### LINKED ISSUE #910: Set-VSCodeHtmlContentView should allow passing stylesheets and JavaScript files as a parameter
It would be great if we could style our HTML with the new *-VSCodeHtml cmdlets available with v1.4.0. It could possibly be implemented by giving us access to use <head> tags, or by passing in stylesheets to a parameter.

daviwil:
This change adds the client-side behavior for new parameters to
Set-VSCodeHtmlContentView which allow the user to specify JavaScriptPaths
and StyleSheetPaths to include local JavaScript and CSS files into their
HTML views.

Resolves #910.

### DIFF
--- a/examples/Assets/script.js
+++ b/examples/Assets/script.js
@@ -1,0 +1,3 @@
+var h1 = document.createElement('h1');
+h1.appendChild(document.createTextNode('Hello from JavaScript!'));
+document.body.appendChild(h1);
--- a/examples/Assets/style.css
+++ b/examples/Assets/style.css
@@ -1,0 +1,4 @@
+body {
+    color: white;
+    background-color: cornflowerblue;
+}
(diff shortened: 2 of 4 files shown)


kamilkosek:
I was just wondering what the nonce is for 😃 
[nonce](https://developers.google.com/web/fundamentals/security/csp/#if_you_absolutely_must_use_it)

BTW, any plans on parameter passing from content-view to editor command ?

daviwil:
Yep, the nonce helps avoid malicious JS and CSS being loaded into the user's browser.  You could say that someone would just as easily be able to use these new parameters to load malicious scripts but it does give us the capability to further restrict their use from within PowerShell depending on the machine's security settings.

I think the only thing needed for that right now is allowing parameters to be send to registered editor commands via the protocol.  I'll look into it for this release.

PR #966, opened 2017-07-19
Fix debug config snippet for launch file w/args prompt.

rkeithhill:
We were leaving out the call to the SpecifyScriptArgs command.  :-(


daviwil:
Whoops!  Thanks man!

PR #1010, opened 2017-08-25
Switch tasks comand to use System32 instead of sysnative. 

rkeithhill:
This should work in both 32-bit and 64-bit VSCode.

Also removed showOutput=always as that is the default plus it is deprecated.


daviwil:
Thanks Keith!  I did basically the same changes in a branch this morning but I'll go ahead and take yours :)

PR #1021, opened 2017-09-05
Various fixes for 64-bit VS Code

daviwil:
Also introduced unit testing for the extension!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ vscode-powershell.zip
 vscps-preview.zip
 *.vsix
 npm-debug.log
+.vscode-test/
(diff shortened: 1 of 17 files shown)


daviwil:
Argh, AppVeyor doesn't kick in when PR'ing to a non-master branch.  At any rate, the plan here is to ship 1.4.2 out of the `release/1.4.2` branch and then merge it back into master once that's done.

rkeithhill:
Did you add anything to the log file that indicates the architecture of the VSCode instance?  Would be handy to know when tracking down bugs.

daviwil:
Nope, good idea!  I'll add that too.

PR #1056, opened 2017-10-16
Add docker based build

TravisEz13:
Add code to define release build using docker to make sure it is reproducible.


daviwil:
This is fantastic Travis, thanks a bunch!

PR #1062, opened 2017-10-21
Update to use DebugConfigurationProvider

### LINKED ISSUE #1046: VSCODE- Insiders: Timed out waiting for the PowerShell extension to start
### System Details

- Operating system name and version: Windows 10 
- VS Code version: 1.18.0 Insiders (commit date 2017-10-06T05:26:16.770Z)
- PowerShell extension version: 1.4.3
- Output from `$PSVersionTable`:  5.1.15063.632

### Issue Description

Error: Timed out waiting for the PowerShell extension to start.

This seems to be specific for latest 1.18 insiders build. All works normally on 1.17 public version.
I tried multiple computers and its the same behaviour.  
PowerShell extension is timing out when pressing F5, or trying to start any script.

image

### Steps to reproduce
- On clean Windows 10 machine install VS Code Insiders 1.18.0. 
- Install PowerShell extension v.1.4.3
- Wirite simple script (something as Write-Host 1)
- Press F5

Expected behaviour is to diplay 1 in console. But, nothing actually happens. After some time, time out error will appear.

### Attached Logs
[logs.zip](https://github.com/PowerShell/vscode-powershell/files/1362792/logs.zip)

rkeithhill:
Fix #1046
Fix warning in status bar about ThemeColor

This fix requires PSES PR 556 to eliminate one PSES cause of the debug adapter to crash.

Also npm generated a package-lock.json file which, from what I've read, should be checked in.  I haven't included it in this PR - yet.  Thoughts?

Also, I updated min VSCode version to 1.17.0.  I'm not sure when the new DebugConfigurationProvider API was intro'd.  I  have tested with 1.17.2 and it works there so I'm reasonably confident it will work in 1.17.0.  We might be able to reach back a bit further.  We'd have to find out when this API was intro'd.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,7 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
+
+# .gitattributes in project root
+# npm now seems to be insisting on LF - see https://github.com/npm/npm/issues/17161
+package.json      text eol=lf
+package-lock.json text eol=lf
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,13 @@
     "editor.insertSpaces": true,
     "files.trimTrailingWhitespace": true,
 
+    "search.exclude": {
+        "**/node_modules": true,
+        "**/bower_components": true,
+        "logs/": true,
+        "out/": true
+    },
+
     // Lock the TypeScript SDK path to the version we use
     "typescript.tsdk": "./node_modules/typescript/lib"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3102 @@
+{
+  "name": "PowerShell",
+  "version": "1.4.3",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/mocha": {
+      "version": "2.2.43",
+      "resolved": "https://registry.npmjs.org/@types/mocha/-/mocha-2.2.43.tgz",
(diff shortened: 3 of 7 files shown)


rkeithhill:
Hmm:
```
error TS2305: Module ''vscode'' has no exported member 'DebugConfigurationProvider'.
```
Maybe we just stick with a minimum of VSCode `1.17.0` then.

daviwil:
Yep, go ahead and add package-lock.json, it's new the new standard behavior in Node 8 I think.

I suppose my only concern here is whether we ship an update later this week (I'd like to) and if this will work in VS Code 1.16.0.  Have you tried loading the built extension in that version to see if it works?

rkeithhill:
I bumped the version down to 1.16, pushed that but one of the builds failed.  So I backed that change out.

rkeithhill:
I'll add the package-lock.json file as soon as I get home tonight.

rkeithhill:
Done.

rkeithhill:
For the record, it seems the API we need only exists in 1.16.0 in the vscode.proposed.d.ts typings file.  In 1.17.0 the API appears in the vscode.d.ts file.

daviwil:
I just tried your changes in 1.17.2, they work great! Thinking about merging this now, anything else you needed to do?

rkeithhill:
I think this one is ready to go.

daviwil:
Awesome, thanks a lot for fixing it!

PR #1063, opened 2017-10-22
Fix tasks to run under VSCode 1.17.2 on Linux

rkeithhill:
None of these were working correctly on Ubuntu 16.04.

Looking at the original file in VSCode 1.17.2 showed deprecation warnings on `suppressTaskName`.  But you can't "just" remove that field other it was sticking in the task name as part of the command.  Switching to `command` seems to fix this.


daviwil:
Looks good, thanks!

PR #1065, opened 2017-10-22
Give better error msg when attach to process w/PS Core

rkeithhill:
The current approach spits out a message in the terminal about a missing wincore pkg.

Here's the new error:
image


daviwil:
Looks good, thanks Keith!

PR #1068, opened 2017-10-25
Update attach to host logic - ps core does work windows

rkeithhill:


src/features/DebugSession.ts
```diff
@@ -57,7 +57,7 @@ export class DebugSessionFeature implements IFeature {
             if (versionDetails.edition.toLowerCase() === "core" &&
                 platformDetails.operatingSystem !== OperatingSystem.Windows) {
 
-                let msg = "PowerShell Core does not support attaching to a host process on non-Windows operating systems.";
+                let msg = "PowerShell Core does not support attaching to a host process except on Windows.";
```
daviwil: I'd go with "PowerShell Core only supports attaching to a host process on Windows.", seems to flow better that way

PR #1071, opened 2017-10-26
Update to handle PS Core binary name change

### LINKED ISSUE #1064: Update extension and PSES to handle PS Core rename of binary to pwsh
Title says it all.  This one will be a bit tricky since we support side-by-side PS Core installations.  We can't wholesale change `powershell` to `pwsh` - just for PS Core 6.0.0-beta.9 and higher.  Do we already have an npm package pulled in that can compare semver v2 versions?

rkeithhill:
Fix #1064

Also fixes case where the powerShellExePath setting is no longer valid.  This happens when a Windows users installs an updated pre-release version of PS Core.  That uninstalls previous versions.

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -73,7 +73,7 @@ export function getDefaultPowerShellPath(
     }
     else if (platformDetails.operatingSystem == OperatingSystem.MacOS) {
         powerShellExePath = "/usr/local/bin/powershell";
-        if (fs.existsSync("/usr/loca/bin/pwsh")) {
+        if (fs.existsSync("/usr/local/bin/pwsh")) {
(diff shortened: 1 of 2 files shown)


src/platform.ts
```diff
@@ -73,9 +73,15 @@ export function getDefaultPowerShellPath(
     }
     else if (platformDetails.operatingSystem == OperatingSystem.MacOS) {
         powerShellExePath = "/usr/local/bin/powershell";
+        if (fs.existsSync("/usr/loca/bin/pwsh")) {
```
daviwil: Missed an 'l' in 'local' in the path

PR #1088, opened 2017-11-03
Update to snippets\powershell.json to improve folding of comment-based help block

AspenForester:
By prefixing each help keyword with two spaces (half a tab) VSCode will fold up the comment-based help block as a whole.  Both the "ex-cmdlet" and "Comment-Help" section of powershell.json have been updated.


rkeithhill:
I'm inclined to wait on VSCode to provide a proper fix to this rather than work around it.  The vast majority of comment help I've seen has the section keywords starting in column zero.  Also the built-in comment-help auto-completion (on `<#`) starts in column 0.

AspenForester:
It's a fair cop!

JB

---------------------------------------------------------------------
http://randombodger.blogspot.com

rkeithhill:
I wish I knew why it's taking VSCode so long to develop a protocol to allow language extensions to provide input on code folding.  It just doesn't seem like it'd be that hard.

daviwil:
@rkeithhill perfection takes time ;)

PR #1095, opened 2017-11-10
Workaround VSCode bug where ${script} var not expanded

rkeithhill:
This happens when you start a debug session from a git diff window.  Helps address


daviwil:
They dropped support for that, or it just isn't working for now?

rkeithhill:
Didn't drop support but if you have a Git diff window active with a PS1 file in it and press F5 to start debugging, the variables are not expanded.  This is a bug that was fixed but not for 1.18.  See https://github.com/Microsoft/vscode/issues/37821

PR #1106, opened 2017-11-18
Replace big build status table with compact badges

felixfbecker:
The left side of a badge is supposed to act as a label for the status shown on the right, so we shouldn't need a table with redundant labels. Makes the top of the readme easy to scan for the health pulse of the project without pushing the description down 🙂


daviwil:
Thanks Felix!

PR #1135, opened 2017-12-26
Example build configuration using MSBuild

MartinSGill:
* Possible implementation for #1127
* Tested both with and without EditorServices
* Tested with MSBuild 15.3

Usage: PS> ./build.ps1

Known Issues

* Had to fix TSC compile error in src/features/DebugSession.ts
* BinaryLogger switch requires ProjectImports=None to avoid: Error: EBUSY: resource busy or locked, open 'vscode-powershell\msbuild.ProjectImports.zip' during package
* Not Tested/Updated AppVeyor


rkeithhill:
I appreciate the effort but I don't see this PR being accepted given issue #1127 and more importantly - most of the other PowerShell projects, including the big Kahuna PowerShell Core, are based on the InvokeBuild module.

Personally, I'd like to standardize on MSBuild too but not until such time as it supports language specific DSLs - see https://github.com/Microsoft/msbuild/issues/1289

The proj file in this PR is quite a bit less readable IMO when compared to either InvokeBuild or psake scripts.  I'd really love to see that MSBuild issue gain traction.  Just imagine using MSBuild's engine, task libraries and platform specific targets files coupled with task/property/itemgroup/target scripting in PowerShell.  That'd be frickin' awesome.

MartinSGill:
@rkeithhill I wasn't really expecting the pull request to be accepted. It was more intended to kick-start and contribute to the conversation.

My main concern is the concept of "aligning" on PSCore, which does not use Invoke-Build. Just a collection of functions that are imported which I see as a step backwards.

Aligning on Invoke-Build I'd not object to, but unless I entirely misunderstood #1127 the idea was to move to custom imported functions, as build1.psm1 does.

rkeithhill:
>  "aligning" on PSCore, which does not use Invoke-Build

My bad.  It is not use Invoke-Build.  PSReadLine uses it and PSScriptAnalyzer as well (I think).  Plaster is using psake.

> the idea was to move to custom imported functions

Yeah that is @tylerl0706 idea AFAICT.  While I agree that consistency is good, the "context" is quite different between PS Core and the extension and frankly just about every other project because PS Core is fairly unique.  Hmm, I think I'll continue this "thought" over on issue #1127.

MartinSGill:
Closing as this PR was mostly to stimulate discussion.

PR #1161, opened 2018-01-12
Check for the expected version of the PowerShell Editor Services module fails because of the wrong function parameters

ant-druha:
'Test-ModuleAvailable' function does not have -RequiredVersion parameter, hence the script fails to evaluate the condition (at least for PS version 3.0.1).


rkeithhill:
Seems like a legit issue / fix.  👍  Thanks!

daviwil:
I'll merge it for now since if nothing else, it impacts PSv3 users.

PR #1198, opened 2018-02-15
WIP Enhance Start-EditorServices.ps1 for better logging and fix bugs

rkeithhill:
The first commit, fixes a few bugs with the original Test-PortAvailability function. The second commit. try a different approach to looking for ports in use - see https://stackoverflow.com/questions/570098/in-c-how-to-check-if-a-tcp-port-is-available.

BTW I had to remove the signature to test this and there's no point putting it back because it would be invalid.  Seems like we should not commit scripts that are signed but instead, sign them just before publishing.

Here is a sample of `Start-EditorServices.log` contents:
```
**********************
Windows PowerShell transcript start
Start time: 20180214232028
Username: KEITH-YOGA2\Keith
RunAs User: KEITH-YOGA2\Keith
Configuration Name: 
Machine: KEITH-YOGA2 (Microsoft Windows NT 10.0.16299.0)
Host Application: C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command & 'C:\Users\Keith\GitHub\rkeithhill\vscode-powershell\scripts\Start-EditorServices.ps1' -EditorServicesVersion '1.5.1' -HostName 'Visual Studio Code Host' -HostProfileId 'Microsoft.VSCode' -HostVersion '1.5.1' -AdditionalModules @('PowerShellEditorServices.VSCode') -BundledModulesPath 'C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\module' -EnableConsoleRepl -LogLevel 'Verbose' -LogPath 'C:\Users\Keith\GitHub\rkeithhill\vscode-powershell\logs\1518675624-someValue.sessionId\EditorServices.log' -SessionDetailsPath 'C:\Users\Keith\GitHub\rkeithhill\vscode-powershell\sessions\PSES-VSCode-4788-759658' -FeatureFlags @()
Process ID: 6228
PSVersion: 5.1.16299.248
PSEdition: Desktop
PSCompatibleVersions: 1.0, 2.0, 3.0, 4.0, 5.0, 5.1.16299.248
BuildVersion: 10.0.16299.248
CLRVersion: 4.0.30319.42000
WSManStackVersion: 3.0
PSRemotingProtocolVersion: 2.3
SerializationVersion: 1.1.0.1
**********************
Transcript started, output file is C:\Users\Keith\GitHub\rkeithhill\vscode-powershell\logs\1518675624-someValue.sessionId\Start-EditorServices.log
VERBOSE: 
#-- Updated PSModulePath to: -------------------------------------------------
VERBOSE: C:\Users\Keith\Documents\WindowsPowerShell\Modules
VERBOSE: C:\Program Files\WindowsPowerShell\Modules
VERBOSE: C:\WINDOWS\system32\WindowsPowerShell\v1.0\Modules\
VERBOSE: C:\Program Files\Microsoft Message Analyzer\PowerShell\
VERBOSE: C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\module
VERBOSE: 
#-- Check required modules available -----------------------------------------
VERBOSE: Testing module availability PowerShellGet
VERBOSE: Populating RepositorySourceLocation property for module PowerShellGet.
VERBOSE: Loading module from path 'C:\Users\Keith\Documents\WindowsPowerShell\Modules\PowerShellGet\1.6.0\PSModule.psm1'.
VERBOSE: Loading module from path 'C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.0.0.1\PSModule.psm1'.
VERBOSE: PowerShellGet  found
VERBOSE: Testing module availability PowerShellEditorServices 1.5.1
VERBOSE: Loading module from path 'C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\module\PowerShellEditorServices\PowerShellEditorServices.psm1'.
VERBOSE: PowerShellEditorServices 1.5.1 found
VERBOSE: 
#-- Start up PowerShellEditorServices ----------------------------------------
VERBOSE: Importing PowerShellEditorServices
VERBOSE: Loading module from path 'C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\module\PowerShellEditorServices\PowerShellEditorServices.psd1'.
VERBOSE: Loading module from path 'C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\module\PowerShellEditorServices\PowerShellEditorServices.psm1'.
VERBOSE: Exporting function 'Start-EditorServicesHost'.
VERBOSE: Exporting function 'Compress-LogDir'.
VERBOSE: Exporting function 'Get-PowerShellEditorServicesVersion'.
VERBOSE: Importing function 'Compress-LogDir'.
VERBOSE: Importing function 'Get-PowerShellEditorServicesVersion'.
VERBOSE: Importing function 'Start-EditorServicesHost'.
VERBOSE: Ports in use in range 10000-30000:
VERBOSE: 12325,15059
VERBOSE: Searching for available socket port for the language service
VERBOSE: Checking port: 12777, attempts remaining 10 --------------------
VERBOSE: Port: 12777 is available
VERBOSE: Searching for available socket port for the debug service
VERBOSE: Checking port: 24342, attempts remaining 10 --------------------
VERBOSE: Port: 24342 is available
PowerShell Integrated Console
VERBOSE: Invoking Start-EditorServicesHost
VERBOSE: Start-EditorServicesHost returned Microsoft.PowerShell.EditorServices.Host.EditorServicesHost
VERBOSE: Writing session file with contents:
VERBOSE: {"debugServicePort":24342,"status":"started","channel":"tcp","languageServicePort":12777}
VERBOSE: Wrote out session file
VERBOSE: 
#-- Waiting for EditorServicesHost to complete execution ---------------------

```

### DIFF
--- a/scripts/Start-EditorServices.ps1
+++ b/scripts/Start-EditorServices.ps1
@@ -86,6 +86,7 @@ function LogSection([string]$msg) {
 function Log([string[]]$msg) {
     $msg | Write-Verbose
 }
+
 function ExitWithError($errorString) {
     Write-Host -ForegroundColor Red "`n`n$errorString"
 
--- a/scripts/Start-EditorServices.ps1
+++ b/scripts/Start-EditorServices.ps1
@@ -71,13 +71,22 @@ param(
     $ConfirmInstall
 )
 
+$minPortNumber = 10000
+$maxPortNumber = 30000
+
 if ($LogLevel -ne "Normal") {
     $VerbosePreference = 'Continue'
     Start-Transcript (Join-Path (Split-Path $LogPath -Parent) Start-EditorServices.log) -Force
 }
 
-function ExitWithError($errorString) {
+function LogSection([string]$msg) {
+    Write-Verbose "`n#-- $msg $('-' * ([Math]::Max(0, 73 - $msg.Length)))"
+}
 
+function Log([string[]]$msg) {
+    $msg | Write-Verbose
+}
+function ExitWithError($errorString) {
     Write-Host -ForegroundColor Red "`n`n$errorString"
 
     # Sleep for a while to make sure the user has time to see and copy the
@@ -97,7 +106,10 @@ if ($PSVersionTable.PSVersion.Major -le 2) {
 }
 
 function WriteSessionFile($sessionInfo) {
-    ConvertTo-Json -InputObject $sessionInfo -Compress | Set-Content -Force -Path "$SessionDetailsPath" -ErrorAction Stop
+    $sessionInfoJson = ConvertTo-Json -InputObject $sessionInfo -Compress
+    Log "Writing session file with contents:"
+    Log $sessionInfoJson
+    $sessionInfoJson | Set-Content -Force -Path "$SessionDetailsPath" -ErrorAction Stop
 }
 
 if ($host.Runspace.LanguageMode -eq 'ConstrainedLanguage') {
@@ -116,6 +128,7 @@ $isPS5orLater = $PSVersionTable.PSVersion.Major -ge 5
 # If PSReadline is present in the session, remove it so that runspace
 # management is easier
 if ((Get-Module PSReadline).Count -gt 0) {
+    LogSection "Removing PSReadLine module"
     Remove-Module PSReadline -ErrorAction SilentlyContinue
 }
 
@@ -125,28 +138,64 @@ if ((Get-Module PSReadline).Count -gt 0) {
 $resultDetails = $null;
 
 function Test-ModuleAvailable($ModuleName, $ModuleVersion) {
+    Log "Testing module availability $ModuleName $ModuleVersion"
+
     $modules = Get-Module -ListAvailable $moduleName
     if ($modules -ne $null) {
         if ($ModuleVersion -ne $null) {
             foreach ($module in $modules) {
                 if ($module.Version.Equals($moduleVersion)) {
+                    Log "$ModuleName $ModuleVersion found"
                     return $true;
                 }
             }
         }
         else {
+            Log "$ModuleName $ModuleVersion found"
             return $true;
         }
     }
 
+    Log "$ModuleName $ModuleVersion NOT found"
     return $false;
 }
 
-function Test-PortAvailability($PortNumber) {
-    $portAvailable = $true;
+function Get-PortsInUse {
+    $portsInUse = @{}
+    $ipGlobalProps = [System.Net.NetworkInformation.IPGlobalProperties]::GetIPGlobalProperties()
+
+    $tcpConns = $ipGlobalProps.GetActiveTcpConnections()
+    foreach ($tcpConn in $tcpConns) {
+        $port = $tcpConn.LocalEndpoint.Port
+        if (($port -ge $minPortNumber) -and ($port -le $maxPortNumber)) {
+            $portsInUse[$port] = 1
+        }
+    }
+
+    $tcpListeners = $ipGlobalProps.GetActiveTcpListeners()
+    foreach ($tcpList in $tcpListeners) {
+        $port = $tcpList.Port
+        if (($port -ge $minPortNumber) -and ($port -le $maxPortNumber)) {
+            $portsInUse[$port] = 1
+        }
+    }
+
+    $udpListeners = $ipGlobalProps.GetActiveUdpListeners()
+    foreach ($udpList in $udpListeners) {
+        $port = $udpList.Port
+        if (($port -ge $minPortNumber) -and ($port -le $maxPortNumber)) {
+            $portsInUse[$port] = 1
+        }
+    }
+
+    $portsInUse
+}
+
+<#
+function Test-PortAvailability([Parameter(Mandatory=$true)][int]$PortNumber) {
+    $portAvailable = $false;
 
     try {
-        Write-Host "Testing for availability of port $PortNumber"
         if ($isPS5orLater) {
             $ipAddress = [System.Net.Dns]::GetHostEntryAsync("localhost").Result.AddressList[0];
         }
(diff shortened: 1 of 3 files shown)


rkeithhill:
Well, shoot.  The `NetworkInformation` classes I tested with on PS Core (on Windows) don't seem to work on Linux.  That's a bummer because when I tested ports I knew were in use (via tcpview, netstat), the old `Test-PortAvailability` function showed them as available. 

Try this.  Open a VSCode and open a PS1 file, not the lang/debug ports in the log file.  Now try this (substitute in your port #):
```posh
$port = 29991
$ipAddress = [System.Net.Dns]::GetHostEntryAsync("localhost").Result.AddressList[0];
try { $tcp = New-Object System.Net.Sockets.TcpListener @($ipAddress, $port); $tcp.Start(); $tcp.Stop() } catch [System.Net.Sockets.SocketException] { $_ }
```
Even though that port is in use by an instance of PSES, it shows as available.  :-(

rkeithhill:
Hmm, if I retrieve the IP address like so, the above code works better:

tylerl0706:
I noticed that you can do:

```
$ipAddress = [System.Net.IPAddress]::Parse("127.0.0.1")
ipAddress.MapToIPv6() # or .MapToIPv4
```

but I haven't figured out how to get the ports just yet

rkeithhill:
OK, I think this is ready for a serious review.  I'd like to get this into 1.6.0 to see if we can get some info to track down why PSES isn't starting up for some folks.  I would appreciate someone testing this on macOS.

tylerl0706:
Any particular test steps or would you like me to just start/stop vscode a bunch of times?

rkeithhill:
The tcp port range is so broad - 10000 to 30000 that it is unlikely that you'll run into ports in use.  Could you start one VSCode session - note one of the ports in use (in the log file).  Then temporarily hack the $minPortNumber / $maxPortNumber values in this file (just under params decl) to a range of say 4 around the port number you know to be in use.  Then run the extension in the debug host (with log level set to Verbose) and watch the PSIC window as it starts up.  I'd like to see that it finds the port known to be in use to be, well, in use.  And then attempt another port.  Thanks.

rkeithhill:
I wonder if these tests are going to fail until a new signature is applied to this script file?

tylerl0706:
Hmmm I'm not sure that's the case here. I'm seeing this issue across AppVeyor and Travis:

https://travis-ci.org/PowerShell/vscode-powershell/jobs/342931436#L666

tylerl0706:
If that is the case, could we fix the Travis and AppVeyor definitions to be ok with a miss matched signature?

If we merge this in, every build will fail (if the mismatched sig is the issue), right?

rkeithhill:
I'm not sure. This is the first time I've touched this file but that build failure seems more related to a node/npm issue. Hmm...

tylerl0706:
I've restarted them. Let's see what happens. It's also possible that one of our dependencies broke us.

If these fail, I want to trigger a build of master which _should_ pass.

tylerl0706:
Looks like Travis and AppVeyor had one too many that night and have recovered :)

scripts/Start-EditorServices.ps1
```diff
@@ -120,75 +139,108 @@ if ((Get-Module PSReadline).Count -gt 0) {
...
+        do {
+            $port = $rand.Next($minPortNumber, $maxPortNumber)
+        }
+        while ($portsInUse.ContainsKey($port))
+
+        # Whether we succeed or fail, don't try this port again
+        $portsInUse[$port] = 1
+
+        Log "Checking port: $port, attempts remaining $triesRemaining --------------------"
+        if ($true -eq (Test-PortAvailability -PortNumber $port)) {
```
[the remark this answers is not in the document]
rkeithhill: I think I'll put that back the way it was.  The reasoning here is if someone tweaks the function and has it output more values the condition should handle that.  Turns out that `$true -eq '',''` is still true but `'','' -eq $true` is false.  We want the latter in this case.

scripts/Start-EditorServices.ps1
```diff
@@ -120,75 +139,108 @@ if ((Get-Module PSReadline).Count -gt 0) {
...
+        [int]
+        $PortNumber
+    )
+
+    $portAvailable = $true
 
     try {
         if ($isPS5orLater) {
-            $ipAddress = [System.Net.Dns]::GetHostEntryAsync("localhost").Result.AddressList[0];
+            $ipAddresses = [System.Net.Dns]::GetHostEntryAsync("localhost").Result.AddressList
```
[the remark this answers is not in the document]
rkeithhill: The above returns both the ipv6 and ipv4 addresses.  Note that I'm not limiting the value to the first index and below we iterate the values.  IOW the port has to be open on both ipv6 and ipv4 before we will use it.

scripts/Start-EditorServices.ps1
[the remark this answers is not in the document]
tylerl0706: That's an interesting gotcha 😅

scripts/Start-EditorServices.ps1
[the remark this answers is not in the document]
tylerl0706: Ah. I see! Thanks!

scripts/Start-EditorServices.ps1
```diff
@@ -71,8 +71,23 @@ param(
     $ConfirmInstall
 )
 
-function ExitWithError($errorString) {
+$minPortNumber = 10000
+$maxPortNumber = 30000
+
+if ($LogLevel -ne "Normal") {
```
daviwil: There are "lesser" log levels than Normal (though honestly they're kinda pointless), so maybe you should check for Verbose and Diagnostic specifically?  Or maybe just Diagnostic?  Can definitely change this in a separate PR so that the release isn't held up.

daviwil: Yeah, I think "Diagnostic" is the right level to enable the logging since we'd only use that level in cases where things have _really_ gone sideways :)

rkeithhill: I'll change this tonight.  BTW how do we handle the fact that this file had an Authenticode signature which I had to remove because my changes invalidated it?

daviwil: @tylerl0706 needs to sign the script again when he ships the update.  In the past I wasn't checking in the authenticode part, but as time went on it seemed easier to just check it in since the script doesn't change that often between releases.

rkeithhill: On the VSCode side, `Diagnostics` isn't one of the logging options although I suppose we could add it:

image

Thoughts?  `Diagnostics`  is available on the PSES side.

daviwil:
`Diagnostic` should be there in the VS Code setting, someone added that for 1.5.1. I'm OK with Verbose until next release if it isn't doable though.

rkeithhill:
It's not there but I'm working on adding it right now.

daviwil:
Weird, could have sworn it was there but it's not!

rkeithhill:
OK, I've about got this licked.  Just need to do a bit of testing.  Expect update in 15 mins or so.

rkeithhill:
This will be nicer:

image

rkeithhill:
And now we get DIAGNOSTIC log output in the PSES log file.
```
2/20/2018 9:25:00 PM [DIAGNOSTIC] - Method "ReadMessage" at line 114 of C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\src\PowerShellEditorServices.Protocol\MessageProtocol\MessageReader.cs
```

rkeithhill:
@daviwil Can you review the last commit that adds the Diagnostic log level - https://github.com/PowerShell/vscode-powershell/pull/1198/commits/913d9f77c54899b6e19c87290ccdc254e66efebc

Issue #1208, opened 2018-02-22
Temporary windows still close after script run


Jaykul:
If I configure `"createTemporaryIntegratedConsole": true` in a particular configuration in the launch.json ...
then consoles launched via that config stay open. 

But if I set my workspace settings to
```json
"powershell.debugging.createTemporaryIntegratedConsole": true
```

The new console is created, but still closes after run.

### System Details
Windows 10
VS Code Insiders 1.21 (2018-02-21)
PowerShell extension version: 1.6

I can attach more logs or whatever if you need them, but I have a feeling they're redundant in this case.

rkeithhill:
No need for logs.  Got the problem spotted and fixed.  Will be in next update.

PR #1209, opened 2018-02-23
Set createTempIntCon when createing launch config for Pester test

rkeithhill:
Fixex #1208

### DIFF
--- a/src/features/PesterTests.ts
+++ b/src/features/PesterTests.ts
@@ -7,6 +7,7 @@ import vscode = require("vscode");
 import Window = vscode.window;
 import { IFeature, LanguageClient } from "../feature";
 import { SessionManager } from "../session";
+import Settings = require("../settings");
 import utils = require("../utils");
 
 export class PesterTestsFeature implements IFeature {
@@ -33,6 +34,7 @@ export class PesterTestsFeature implements IFeature {
     private launchTests(uriString, runInDebugger, describeBlockName?) {
         const uri = vscode.Uri.parse(uriString);
         const currentDocument = vscode.window.activeTextEditor.document;
+        const settings = Settings.load();
 
         const launchConfig = {
             request: "launch",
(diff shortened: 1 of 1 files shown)


src/features/PesterTests.ts
```diff
@@ -47,6 +49,7 @@ export class PesterTestsFeature implements IFeature {
             ],
             internalConsoleOptions: "neverOpen",
             noDebug: !runInDebugger,
+            createTemporaryIntegratedConsole: settings.debugging.createTemporaryIntegratedConsole,
```
daviwil: Should the setting only be used when runInDebugger is true?  I could go either way

rkeithhill: I think there is benefit in being able to run tests in a fresh session.

daviwil: Maybe if "runInDebugger" is false we can use a different title?  That was short-sighted on my part

daviwil: Or `[TEMP]` just to be extra clear, but anything works

rkeithhill: I wish the VSCode folks would make this dropdown a tad wider:
image
What do you think?

rkeithhill:
Which specific comment?  

Right now, the behavior is that whether you are running Pester tests or debugging, if  `powershell.debugging.createTemporaryIntegratedConsole` is set to `true`, either of these operations happen in the terminal window `[TEMP] PowerShell Integrated Console`.  That terminal window sticks around until the next Pester test run or debug session at which time the existing PowerShell process is closed and recreated, attached to a new VSCode terminal window with the exact same name.  `[TEMP] PowerShell Integrated Console` is effectively the one and only terminal window for `"powershell.debugging.createTemporaryIntegratedConsole": true`  even though the associated PowerShell process changes with each run.  I think that is what we want but if not, let me know.

tylerl0706:
Sorry, the comment on changing `dbg` to `temp`

Issue #650, opened 2017-04-04
F5 on Untitled-1 file in 0.12.1 with existing .vscode\launch.json from prev version hangs debugger


daviwil:
_From @dfinke on April 4, 2017 22:9_

This is the log.
[EditorServices.txt](https://github.com/PowerShell/PowerShellEditorServices/files/894657/EditorServices.txt)

and the launch.json
[launch.txt](https://github.com/PowerShell/PowerShellEditorServices/files/894658/launch.txt)

_Copied from original issue: PowerShell/PowerShellEditorServices#417_

daviwil:
Thanks Doug!  I see the problem here, I wonder if other people will hit it too.  Might ship another patch update tomorrow.

dfinke:
Not a show stopper.

daviwil:
I *believe* this will be fixed in 0.12.2 but I'm keeping this open so we can verify just in case.

daviwil:
Just checked with Doug, this is fixed in 0.12.2!

PR #1261, opened 2018-04-06
Add CurrentFile SaveAs() support

rjmholt:
In combination with [PSES #650](https://github.com/PowerShell/PowerShellEditorServices/pull/650), this adds the ability to run `$psEditor.GetEditorContext().CurrentFile.SaveAs("newFile")`.

Some things to note (that I'm looking for input on):
* The `SaveAs` method supports both absolute and relative paths, relative being relative to the current file path.
* I've been trying to make VSCode open the newly saved file in a a tab/window, but after trying quite a few different things, it seems to really want to close the window. Any ideas?
* There are a few ways the saving can fail. My current solution is for PSES to handle that possibility with pre-validation, but maybe we should send a success/fail message back instead?


src/features/ExtensionCommands.ts
```diff
@@ -382,23 +387,54 @@ export class ExtensionCommandsFeature implements IFeature {
...
+                                return editor.edit((editBuilder) => {
+                                    editBuilder.insert(new vscode.Position(0, 0), oldDoc.getText());
+                                })
+                                .then(() => {
+                                    return newDoc.save()
+                                        .then(() => EditorOperationResponse.Completed);
+                                });
+                            });
+                    });
+            });
```
tylerl0706: I think instead of this scary nested code, we could use something like `Promise.all` and fire all the ones we can in that and then use the data.

Promise.all notes:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

tylerl0706: I've been trying to make this with less nesting... it's tough :(

Here's the best I've got so far:

```typescript
vscode.workspace.openTextDocument(newFileUri)
    .then((newDoc) => vscode.window.showTextDocument(newDoc, 1, false))
    .then((editor) => editor.edit((editBuilder) => {
            vscode.workspace.openTextDocument(saveFileDetails.filePath)
            .then((oldDoc) => {
                editBuilder.insert(new vscode.Position(0, 0), oldDoc.getText());
            })
        })
        .then(() => editor.document.save()))
    .then(() => EditorOperationResponse.Completed);
```

tylerl0706: this looks a bit cleaner IMO, thoughts?

cc @rkeithhill .

rjmholt:
I think I tried using `saveFileDetails.newPath` but it opens a strange editor window which isn't really native. I had to pass it the new document handle to get it to work.

src/features/ExtensionCommands.ts
tylerl0706: triggered

tylerl0706:
I'm considering posting on Stack Overflow on how one might saveAs using the vscode API. I feel like it's a reasonable question and people would have use of it.

rjmholt:
Yeah I agree. Either there or on their [GitHub page](https://github.com/microsoft/vscode)

src/features/ExtensionCommands.ts
rjmholt: Needs a nice syntactic sugar I think..[.](https://en.wikibooks.org/wiki/Haskell/do_notation)

tylerl0706:
Here it is: https://stackoverflow.com/questions/49787354/implementing-a-save-as-using-the-vscode-api

SeeminglyScience:
Any reason an async function won't work?

```typescript
private async saveFile(saveFileDetails: ISaveFileDetails): Promise<EditorOperationResponse> {

    // If the file to save can't be found, just complete the request
    if (!this.findTextDocument(this.normalizeFilePath(saveFileDetails.filePath))) {
        return EditorOperationResponse.Completed;
    }

    // If no newFile is given, just save the current file
    if (!saveFileDetails.newPath) {
        const doc = await vscode.workspace.openTextDocument(saveFileDetails.filePath);
        if (doc.isDirty) {
            await doc.save();
        }
        return EditorOperationResponse.Completed;
    }

    // Otherwise we want to save as a new file

    // First turn the path we were given into an absolute path
    // Relative paths are interpreted as relative to the original file
    const newFileAbsolutePath = path.isAbsolute(saveFileDetails.newPath) ?
        saveFileDetails.newPath :
        path.resolve(path.dirname(saveFileDetails.filePath), saveFileDetails.newPath);

    // Create an "untitled-scheme" path so that VSCode will let us create a new file with a given path
    const newFileUri = vscode.Uri.parse("untitled:" + newFileAbsolutePath);

    // Now we need to copy the content of the current file,
    // then create a new file at the given path, insert the content,
    // save it and open the document
    const oldDoc = await vscode.workspace.openTextDocument(saveFileDetails.filePath);
    const newDoc = await vscode.workspace.openTextDocument(newFileUri);
    const editor = await vscode.window.showTextDocument(newDoc, 1, false);
    await editor.edit((editBuilder) => {
        editBuilder.insert(new vscode.Position(0, 0), oldDoc.getText());
    });
    await newDoc.save();
    return EditorOperationResponse.Completed;
}
```

rjmholt:
Wow, that's much better. I had no idea `async` was a think in TypeScript...

tylerl0706:
didn't realize we could use async/await here 😄 thanks for that @SeeminglyScience!

tylerl0706:
I wonder if it's because you're saving an untitled file:

```typescript
const newFileUri = vscode.Uri.parse("untitled:" + newFileAbsolutePath);
```

rjmholt:
I've been trying various ways to "save-as" and then open that window.

Opening with a new "file://" URI after the save, I get the new file opened. But then the old one closes, which isn't what I imagine people wanting.

I'm mildly reluctant to bypass VSCode and just use `fs` like the SO answer suggests, but it might be much easier...

rjmholt:
Ok, I've fixed up the code and it now seems to do what it wants. We should probably test it out a bit and validate the behaviour (and see if using different paths breaks it) before merging, but I think it's good to go.

rkeithhill:
I wonder if we should just ignore the `package-lock.json` file altogether?  We've started doing this at work as this file kept getting changed by everyone's dev environment.  We could nuke this by adding a `.npmrc` with this in it:
```
package-lock=false
```

rjmholt:
Yeah, I had added it to the ignore on the basis that it was generated by the user's environment, but @tylerl0706 suggested we should keep it in as it represents some pre-resolved work for npm.

npm's pages state that it's [intended to be tracked](https://docs.npmjs.com/files/package-lock.json), but if it keeps changing in every commit because different machines use it differently, then I'd argue that it's noise in the commit and we should ignore it (and npm should work out what they're trying to achieve with this file...). But I don't have a strong opinion either way (as long as deleting `package-lock.json` doesn't get us into dependency hell or something).

src/features/ExtensionCommands.ts
```diff
@@ -382,24 +388,45 @@ export class ExtensionCommandsFeature implements IFeature {
...
+        const newFileAbsolutePath = path.isAbsolute(saveFileDetails.newPath) ?
+            saveFileDetails.newPath :
+            path.resolve(path.dirname(saveFileDetails.filePath), saveFileDetails.newPath);
+
+        // Retrieve the text out of the current document
+        const oldDocument = await vscode.workspace.openTextDocument(saveFileDetails.filePath);
+        const oldDocText = oldDocument.getText();
+
+        // Write it to the new document path
+        fs.writeFileSync(newFileAbsolutePath, oldDocText);
```
tylerl0706: Nitpick:
```
fs.writeFileSync(newFileAbsolutePath, oldDocument.getText());
```
deletes a line.

might be worth it but don't care that much.

src/features/ExtensionCommands.ts
```diff
@@ -382,24 +388,45 @@ export class ExtensionCommandsFeature implements IFeature {
...
+
+        // Finally open the new document
+        const newFileUri = vscode.Uri.file(newFileAbsolutePath);
+        const newFile = await vscode.workspace.openTextDocument(newFileUri);
+        vscode.window.showTextDocument(newFile, { preview: false });
```
tylerl0706: Why did you choose to open it not as a preview? Just curious. Not really leaning either way personally.

src/features/ExtensionCommands.ts
```diff
@@ -382,24 +388,45 @@ export class ExtensionCommandsFeature implements IFeature {
         return promise;
     }
 
-    private saveFile(filePath: string): Thenable<EditorOperationResponse> {
+    private async saveFile(saveFileDetails: ISaveFileDetails): Promise<EditorOperationResponse> {
```
tylerl0706: Did you have to switch this from a Thenable to a Promise to use async/await?

rjmholt: Yeah -- as in @SeeminglyScience's snippet. `Thenable` doesn't type check.

src/features/ExtensionCommands.ts
rjmholt: Ah, didn't know what it was (just followed the response on SO figuring it prevented some behaviour). I feel like maybe it actually should be a preview...

tylerl0706: I'm leaning on preview actually too 😄

src/features/ExtensionCommands.ts
tylerl0706: good to know!

src/features/ExtensionCommands.ts
```diff
@@ -382,24 +388,44 @@ export class ExtensionCommandsFeature implements IFeature {
...
+        // Retrieve the text out of the current document
+        const oldDocument = await vscode.workspace.openTextDocument(saveFileDetails.filePath);
+
+        // Write it to the new document path
+        fs.writeFileSync(newFileAbsolutePath, oldDocument.getText());
+
+        // Finally open the new document
+        const newFileUri = vscode.Uri.file(newFileAbsolutePath);
+        const newFile = await vscode.workspace.openTextDocument(newFileUri);
+        vscode.window.showTextDocument(newFile, { preview: true });
```
tylerl0706: preview should be true by default. Were you seeing otherwise?

rjmholt: No, and I originally just took it out. But I couldn't find a reference to what the default is in the documentation and decided that we may as well make the expected behaviour explicit.

brwilkinson:
I have tested out the SaveAs('d:\test.ps1')

This appears to work if the file is already saved as another file, however not if the file is Untitled?

E.g.
Path          : untitled:Untitled-1
WorkspacePath : untitled:Untitled-1

Is someone able to test this and see if they have the same experience ?

rjmholt:
You're quite right. I've opened an issue: https://github.com/PowerShell/vscode-powershell/issues/1303.

PR #1312, opened 2018-05-06
Updates to Examples PSSA settings file to include more rule config

rkeithhill:


rkeithhill:
@bergmeister Could you take a quick look at this?  This is for the Examples folder we ship with VSCode. I've updated the PSSA settings file in it to include info on various rule configurations.

examples/PSScriptAnalyzerSettings.psd1
```diff
@@ -1,6 +1,5 @@
-# The PowerShell Script Analyzer will generate a warning
-# diagnostic record for this file due to a bug -
-# https://github.com/PowerShell/PSScriptAnalyzer/issues/472
+# For more information on PSScriptAnalyzer settings see:
+# https://github.com/PowerShell/PSScriptAnalyzer/blob/master/README.md#settings-support-in-scriptanalyzer
```
bergmeister: Also linking to the shipped settings files could be useful: https://github.com/PowerShell/PSScriptAnalyzer/tree/master/Engine/Settings

examples/PSScriptAnalyzerSettings.psd1
```diff
@@ -25,4 +24,48 @@
...
+    #     }
+    #     PSProvideCommentHelp = @{
+    #         Enable                  = $true
+    #         ExportedOnly            = $false
+    #         BlockComment            = $true
+    #         VSCodeSnippetCorrection = $false
+    #         Placement               = "before"
+    #     }
+    #     PSUseCompatibleCmdlets = @{
+    #         compatibility = @("core-6.0.0-alpha-windows", "core-6.0.0-alpha-linux")
```
bergmeister: I want to merge [this](https://github.com/PowerShell/PSScriptAnalyzer/pull/954) PR before the release next week, which will delete those files and replace it with newer ones. Therefore I would remove thise example, you do not need to document PSSA.

rkeithhill: I'd much prefer a link to a list of configurable PSSA rules but I didn't see one.  I spent ~10 minutes going through every single rule doc, looking for the ones that are configurable.  I think it would be nice if PSSA had a list of configurable rules that could be linked to.  :-)

bergmeister: `Get-ScriptAnalyzer` gives you all the rules and one has to look at each rule doc for configuration. The plan is to include one settings file in the PSSA repo itself with all rules and options being present. I had the same problem when doing it the first time for me a few months ago and had only the rule docs and example files. Since I didn't write the tool, I can also only use as much as currently is there. I learn more about it as I read the code and improve docs as I go along.

rkeithhill: My point is that somewhere in this [README](https://github.com/PowerShell/PSScriptAnalyzer/blob/master/RuleDocumentation/README.md) it would be helpful to have a list of the configurable rules.

rkeithhill:
OK, so this PR title should now be - "Improved documentation & links in the Examples PSSA settings file".

examples/PSScriptAnalyzerSettings.psd1
bergmeister: I have deleted this ReadMe already in the `development` branch because it got too much out of date and there is a lot of duplicated and outdated information on the repo. Why do you want to know whether a rule is configurable? The idea would be that one takes the monolithic example file with all rules and configuration examples and then one deletes the rules that one is not interested in and adapt the configuration of the remaining rules.
There is an outstanding issue to get all configurable rules using `Get-ScriptAnalyzer` and I remember giving it a quick try in an attempted PR and there were various problems, hence why I left the issue for the moment: https://github.com/PowerShell/PSScriptAnalyzer/issues/823

rkeithhill: Ah, that's too bad. I just submitted a PR to fix the broken links and add a column for `Configurable`.  Because I'm too lazy and don't have enough time to read through all the docs.  If a rule isn't working for me, I'll just disable it. OTOH if I saw that the rule was configurable it wouldn't take me long to make it work for me, then I'd leave it enabled.

PR #1317, opened 2018-05-12
Change SpecifyScriptArgs command to only return string - not string[]

### LINKED ISSUE #1315: Intercative variable replacment also supported for string commands
Hello friends, I am from the vscode team and some users are reporting an issue with debugging powershell. More datils can be found here https://github.com/Microsoft/vscode/issues/49390

After investigating I noticed that your command `SpecifyScriptArgs ` which is used to substitue values before debugging is returning an array of strings instead of a string. This was never officialy supported, and with the refactoring on the vscode side we are more strict and only allow strings.
Bottom line you should change a cast to array to simply return a string and not an array here https://github.com/PowerShell/vscode-powershell/blob/6c1e30e2a2f24731f394e55c364f5c09e11e7e6c/src/features/DebugSession.ts#L232

Make sure that all your commands which are used for launch.json configs are returning strings and not array.

rkeithhill:
Fix #1315 but folks won't see this fix until they are on VSCode 1.24
or insiders build >= 5/11.
This is because a corresponding fix is required in VSCode.

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -178,20 +178,9 @@ export class SpecifyScriptArgsFeature implements IFeature {
     private command: vscode.Disposable;
     private languageClient: LanguageClient;
     private context: vscode.ExtensionContext;
-    private emptyInputBoxBugFixed: boolean;
 
     constructor(context: vscode.ExtensionContext) {
         this.context = context;
-
-        const vscodeVersionArray = vscode.version.split(".");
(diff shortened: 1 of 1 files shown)


rkeithhill:
Yup, VSCode 1.11 (and earlier) had a bug so we had code to avoid that bug.  Our min version for VSCode has surpassed 1.11 for some time now so I took the opportunity to remove that unnecessary code.

PR #1366, opened 2018-06-14
WIP: Add support for side-by-side PS Core preview on Linux/macOS

### LINKED ISSUE #1361: PowerShell preview executables no longer at /usr/[local/]bin/pwsh on *nix
PowerShell 6.1.0-preview.3 will install its executable as `pwsh-preview` as of the upcoming release.

We will need to work out how to handle this.

@s2p16 reported this and a workaround here: https://github.com/PowerShell/PowerShell/issues/6132#issuecomment-396079102.

Essentially, for now, the best solution is to specify a new `powerShellExePath`:
```json
{
  "powershell.powerShellExePath": "/usr/bin/pwsh-preview"
}
```

Another alternative would be to use @rkeithhill's EXE path feature:
https://github.com/PowerShell/vscode-powershell/pull/1270

But we should also fix this so that it works out of the box. Since preview and ordinary releases are now allowed side-by-side, we should figure out which we prefer if both are present.

rkeithhill:
## PR Summary

This PR supports listing the symlinks for pwsh & pwsh-preview.

Fixes #1361

The preferred default version is pwsh if both are present however, the
default will be pwsh-preview if it's installed and pwsh isn't.

So this only support side-by-side on Linux/macOS up to two versions:
stable and preview.

In the future, we may want to do what we do on Windows and search thru
the install dir for pwsh.
The would be /opt/microsoft/poweshell on LInux.
Not sure what it is on macOS.

Or maybe we should just do that now. Thoughts? 

I marked this as WIP because I need someone to test this on Linux and macOS.  :-)

## PR Checklist

- [x] Summarized changes

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -7,11 +7,9 @@ import path = require("path");
 import process = require("process");
 import Settings = require("./settings");
 
-const linuxLegacyExePath  = "/usr/bin/powershell";
 const linuxExePath        = "/usr/bin/pwsh";
 const linuxPreviewExePath = "/usr/bin/pwsh-preview";
 
-const macOSLegacyExePath  = "/usr/local/bin/powershell";
 const macOSExePath        = "/usr/local/bin/pwsh";
 const macOSPreviewExePath = "/usr/local/bin/pwsh-preview";
 
@@ -53,6 +51,13 @@ export function getPlatformDetails(): IPlatformDetails {
     };
 }
 
+/**
+ * Gets the default instance of PowerShell for the specified platform.
+ * On Windows, the default version of PowerShell is "Windows PowerShell".
+ * @param platformDetails Specifies information about the platform - primarily the operating system.
+ * @param use32Bit On Windows, this boolean determines will find the 32-bit version of Windows PowerShell.
+ * @returns A string containing the path of the default version of PowerShell.
+ */
 export function getDefaultPowerShellPath(
     platformDetails: IPlatformDetails,
     use32Bit: boolean = false): string | null {
@@ -74,17 +79,15 @@ export function getDefaultPowerShellPath(
                     : SysnativePowerShellPath;
         }
     } else if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
-        powerShellExePath = macOSLegacyExePath;
-        if (fs.existsSync(macOSExePath)) {
-            powerShellExePath = macOSExePath;
-        } else if (fs.existsSync(macOSPreviewExePath)) {
+        // Always default to the stable version of PowerShell (if installed) but handle case of only Preview installed
+        powerShellExePath = macOSExePath;
+        if (!fs.existsSync(macOSExePath) && fs.existsSync(macOSPreviewExePath)) {
             powerShellExePath = macOSPreviewExePath;
         }
     } else if (platformDetails.operatingSystem === OperatingSystem.Linux) {
-        powerShellExePath = linuxLegacyExePath;
-        if (fs.existsSync(linuxExePath)) {
-            powerShellExePath = linuxExePath;
-        } else if (fs.existsSync(linuxPreviewExePath)) {
+        // Always default to the stable version of PowerShell (if installed) but handle case of only Preview installed
+        powerShellExePath = linuxExePath;
+        if (!fs.existsSync(linuxExePath) && fs.existsSync(linuxPreviewExePath)) {
             powerShellExePath = linuxPreviewExePath;
         }
     }
@@ -118,6 +121,12 @@ export function fixWindowsPowerShellPath(powerShellExePath: string, platformDeta
     return powerShellExePath;
 }
 
+/**
+ * Gets a list of all available PowerShell instance on the specified platform.
+ * @param platformDetails Specifies information about the platform - primarily the operating system.
+ * @param sessionSettings Specifies the user/workspace settings. Additional PowerShell exe paths loaded from settings.
+ * @returns An array of IPowerShellExeDetails objects with the PowerShell name & exe path for each instance found.
+ */
 export function getAvailablePowerShellExes(
     platformDetails: IPlatformDetails,
     sessionSettings: Settings.ISettings | undefined): IPowerShellExeDetails[] {
@@ -178,7 +187,7 @@ export function getAvailablePowerShellExes(
             exePath: defaultExePath,
         });
 
-        // If defaultExePath is pwsh, check to see if pwsh-preview is installed and if so, make it available.
+        // If defaultExePath is pwsh, check to see if pwsh-preview is also installed and if so, make it available.
         // If the defaultExePath is already pwsh-preview, then pwsh is not installed - nothing to do.
         let osExePath;
         let osPreviewExePath;
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -174,7 +174,7 @@ export function getAvailablePowerShellExes(
         // Handle Linux and macOS case
         const defaultExePath =  this.getDefaultPowerShellPath(platformDetails);
         paths.push({
-            versionName: "PowerShell Core",
+            versionName: "PowerShell Core" + (/-preview/.test(defaultExePath) ? " Preview" : ""),
             exePath: defaultExePath,
         });
 
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -10,6 +10,9 @@ import Settings = require("./settings");
 const linuxExePath        = "/usr/bin/pwsh";
 const linuxPreviewExePath = "/usr/bin/pwsh-preview";
 
+const snapExePath         = "/snap/bin/pwsh";
+const snapPreviewExePath  = "/snap/bin/pwsh-preview";
+
 const macOSExePath        = "/usr/local/bin/pwsh";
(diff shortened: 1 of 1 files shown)


src/platform.ts
```diff
@@ -162,10 +172,29 @@ export function getAvailablePowerShellExes(
...
+                    exePath: macOSPreviewExePath,
+                });
+            }
+        } else if (platformDetails.operatingSystem === OperatingSystem.Linux) {
+            if ((defaultExePath === linuxExePath) && fs.existsSync(linuxPreviewExePath)) {
+                paths.push({
+                    versionName: "PowerShell Core Preview",
+                    exePath: linuxPreviewExePath,
+                });
+            }
```
tylerl0706: I refactored this a bit to define the paths conditionally and then do the logic. I think it's a little DRY-er. Totally, not a blocker but what do you think?

```typescript
let osExePath, osPreviewExePath;
if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
    osExePath = macOSExePath;
    osPreviewExePath = macOSPreviewExePath;
} else if (platformDetails.operatingSystem === OperatingSystem.Linux) {
    osExePath = linuxExePath;
    osPreviewExePath = linuxPreviewExePath;
}

if ((defaultExePath === osExePath) && fs.existsSync(osPreviewExePath)) {
    paths.push({
        versionName: "PowerShell Core Preview",
        exePath: linuxPreviewExePath,
    });
}

```

rkeithhill: Updated.  Please test on macOS and Linux.

src/platform.ts
```diff
@@ -162,10 +172,30 @@ export function getAvailablePowerShellExes(
...
+        let osPreviewExePath;
+        if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
+            osExePath = macOSExePath;
+            osPreviewExePath = macOSPreviewExePath;
+        } else if (platformDetails.operatingSystem === OperatingSystem.Linux) {
+            osExePath = linuxExePath;
+            osPreviewExePath = linuxPreviewExePath;
+        }
+
+        if ((osExePath === defaultExePath) && fs.existsSync(osPreviewExePath)) {
```
tylerl0706: It's super rare, but isn't it possible for them to have legacy (`powershell`) and preview (`pwsh-preview`)?

If so, this if check might have to be:

`defaultExePath !== osPreviewExePath && ...`

tylerl0706:
Is there any work that needs to be done here:
https://github.com/PowerShell/vscode-powershell/blob/master/test/platform.test.ts

I just saw that it had to something with the exe path.

src/platform.ts
rkeithhill: I can fix this by testing for the legacy path here or maybe we should just remove support for the legacy "powershell"on Linux and macOS.  This was abandoned somewhere around the transition from 6.0.0 alpha to beta.  At this point, I'm not sure why anyone would still be running an alpha of 6.0.0 on Linux/macOS.

rkeithhill:
Those tests still pass for me on Windows.  Looks like the tests mainly check to see that we get back the correct 32/64-bit version of Windows PowerShell on Windows.

tylerl0706:
Sounds good :)

tylerl0706:
So I was messing with the [VSCode](https://snapcraft.io/vscode) & [PowerShell](https://snapcraft.io/powershell) snaps... Snap is essentially an attempt at a "universal *nix package manager... more info [here](https://snapcraft.io/).

Anyway... on all of their supported OS's you should be able to run:
```
sudo snap install --classic vscode
sudo snap install --classic powershell
code --install-extension="ms-vscode.powershell"
```
and get VSCode, PowerShell and the extension installed...

However the extension fails because the "powershell exe" it is trying to use is `/usr/bin/powershell` but that doesn't exist... Snap does something different:

```
PS /home/parallels/Desktop> which powershell                                    
/snap/bin/powershell
PS /home/parallels/Desktop> which pwsh                                          
/snap/powershell/2/opt/powershell/pwsh
```

we should probably handle this in the linux case so we can support the Snap build of PowerShell out of the box.

@rkeithhill would you be ok adding a check for this in this PR?

rkeithhill:
Are those paths on Linux or macOS?  Also, is the binary called `powershell`?? Or is that just a dir name?  I thought the binary would be either `pwsh` or `pwsh-preview`.  And speaking of PowerShell preview, do we need to handle that case with snap?

tylerl0706:
@rkeithhill AFAIK, snap is only linux:
image

tylerl0706:
the binary is called `pwsh` found here:
`/snap/powershell/2/opt/powershell/pwsh` but we can rely on:
`/snap/bin/powershell` or `/snap/bin/pwsh`
which are symlinks

and we should also handle the `pwsh-preview` scenario for snap as well.

tylerl0706:
I've been talking to @TravisEz13 about this who has been driving the Snap work

tylerl0706:
I can't wait to install VSCode & PowerShell on my WRT based Router lol

rkeithhill:
@tylerl0706 Can you test this in a Snap environment?

tylerl0706:
gladly :) I'll test now.

tylerl0706:
@rkeithhill it works!

src/platform.ts
```diff
@@ -162,10 +189,30 @@ export function getAvailablePowerShellExes(
...
+
+        // If defaultExePath is pwsh, check to see if pwsh-preview is also installed and if so, make it available.
+        // If the defaultExePath is already pwsh-preview, then pwsh is not installed - nothing to do.
+        let osExePath;
+        let osPreviewExePath;
+        if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
+            osExePath = macOSExePath;
+            osPreviewExePath = macOSPreviewExePath;
+        } else if (platformDetails.operatingSystem === OperatingSystem.Linux) {
+            osExePath = linuxExePath;
```
tylerl0706: do we not need to do the same ternary op here to check for snap?

tylerl0706:
You might hate me for saying this... but I've thought about this part of the code for a bit and wonder if there's some opportunity to clean up the code a bit...

maybe something like this:
```

function getAvailablePowerShellExes() {
    let exes = [];
    
    if(Windows) {
        ... similar as below
    }
    if(macOS) {
        ... similar to below
    }
    if(Linux) {
        linuxExes = ["/usr/bin/pwsh", "/snap/bin/pwsh", "/usr/bin/pwsh-preview", "/snap/bin/pwsh-preview"]
        linuxExes.forEach((exe) => {
            if(exe exists) {
                exes.push(exe);
            }
        });
    }
    return exes;
}

function getDefualtPowerShellExe() {
    return getAvailablePowerShellExes()[0];
}
```

if we're super worried about perf (in this case, I don't think we are) we could optionally give `getAvailablePowerShellExes` a parameter bool called `stopAtFirstExe` that will break out of the for loop when it finds a valid exe.

I wont block on this, but I feel like it might be worth it...

src/platform.ts
rkeithhill: Not needed here because `getDefaultPowerShellPath` already returns just one of: linuxExePath, linuxPreviewExePath, snapExePath, snapPreviewExePath.  It searches in that order.

rkeithhill:
Not at all. I have no particularly affinity for this code.  :-)  We could make that change but we can't get rid of the test against the default PowerShell exe simply because we don't want the same entry to show up twice.  Also, I don't think perf is that much of a concern here and we do want to list all the available PowerShell instances that can be used.

src/platform.ts
rkeithhill: BTW I get the impression (perhaps incorrect impression) that on a Snap system, PowerShell will only be at /snap/bin/pwsh* and that you wouldn't also have PowerShell at /usr/bin/pwsh*.  And on a straight (non-Snap) Linux system, there would be no /snap/bin/pwsh*.  That is, these two sets are mutually exclusive. Is that correct?

tylerl0706: So snap is really just a package manager that you first install with apt. So you _could_ have installed `pwsh` by Snap and Apt on the same system.

Not quite sure why one would do that... but it's physically possible. I guess I could see something along the lines of installing pwsh with apt and pwsh-preview with snap maybe...

So it's physically possible to have a `/snap/bin/pwsh` and a `/usr/bin/pwsh`.

tylerl0706:
Thanks for putting up with my comments :smile: 

> We could make that change but we can't get rid of the test against the default PowerShell exe simply because we don't want the same entry to show up twice

I don't think I quite follow how we would end up with something showing up twice? Think you could elaborate?

rkeithhill:
The issue is that for some reason (I don't remember) in the current impl of `getAvailablePowerShellExes` we use `getDefaultPowerShellPath` to get at least the one path we know will be on the system.  But we could probably remove that and simplify the impl.  That said, I don't think we can use the first result of `getAvailable` for `getDefault` because `getDefault` takes an extra parameter - `use32` - which on Windows 64-bit, requests the 32-bit version of PowerShell.

rkeithhill:
Updated to simplify per your suggestion.  Can you test again on Linux/macOS?

tylerl0706:
@rkeithhill no surprise your code works beautifully :)

Issue #1405, opened 2018-07-08
Show-Command like panel


corbob:
**Summary of the new feature**

ISE has a pane for Show-Command. It would be good to have something similar to this in VS Code.

Numerous users have expressed the lack of this feature for why they don't use VS Code for PowerShell.

**Proposed technical implementation details (optional)**

A side bar view for this seems to make the most sense.

corbob:
Just putting some thoughts out there so I don't lose them.

I've created a project on my fork for this, adding cards for small chunks as I think of them or encounter someone mentioning it. The project is here: https://github.com/corbob/vscode-powershell/projects/1

Unfortunately due to this issue: https://github.com/Microsoft/vscode/issues/50062 We may not be able to get searching working just yet. My focus will be around what's available to us today and if we get search/filter in the future then we can work on that.

Ayanmullick:
<a><img align="right" src="https://user-images.githubusercontent.com/17693249/139517342-d540ad0f-22d8-490d-8dc1-c2d1bf360427.png" width="330" height="600" border="10"/></a>

@rjmholt, @andschwa, PFB the features missing from the VSC PowerShell extension's Command Explorer, [as discussed during the Community call for October 21](https://www.youtube.com/embed/_j5bO9yTexY?start=2971&end=3133).

<br>

<br>

<br>

### 1. Filter by modules
### 2. Search for specific cmdlets
### 3. Fill out specific parameter sets
### 4. Add common parameters
### 5. Insert filled out cmdlet

SeeminglyScience:
Personally I'd really like to see some work put into this extension's API to give the community the tools to provide a separate extension with this functionality.

nthoman:
If you want Visual Studio Code and PS7 to be successful you need to include this functionality into the plugin. ISE mode and the command bar has been vital to grow my skills (and many others) in PowerShell since its introduction. It allows one to quickly filter commands and find examples how to use them correctly. By excluding this functionality you are increasing the barrier of entry for PowerShell 7. If you are experienced with PowerShell then VSC is fine and you can get by. I am constantly teaching inexperienced staff on my team on how useful PowerShell can be to simplify repetitive operational tasks and encourage them to get familiar with it. However I find the VSC way too "opaque" and usually just fall back to the original ISE. 

[...]

PR #1406, opened 2018-07-08
WIP: Show-Command explorer

corbob:
## PR Summary

Implement a Show-Command type panel for https://github.com/PowerShell/vscode-powershell/issues/1405

## PR Checklist

- [ ] Summarized changes


package.json
```diff
@@ -58,6 +58,23 @@
     "test": "node ./node_modules/vscode/bin/test"
   },
   "contributes": {
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "command-explorer",
+          "title": "Command Explorer",
+          "icon": "media/dep.svg"
```
tylerl0706: I feel like this should be the PowerShell logo

package.json
```diff
@@ -58,6 +58,23 @@
...
+        }
+      ]
+    },
+    "views": {
+      "command-explorer": [
+        {
+          "id": "commands",
```
tylerl0706: Could we prefix this with  `PowerShell.` so that we have a common prefix and a better chance of it being unique.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,139 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+// tslint:disable:no-console
```
tylerl0706: I don't think we need this anymore

src/features/GetCommands.ts
```diff
@@ -0,0 +1,139 @@
...
+import * as vscode from "vscode";
+import { LanguageClient, RequestType } from "vscode-languageclient";
+import { IFeature } from "../feature";
+
+export const GetCommandsRequestType = new RequestType<any, any, void, void>("powerShell/getCommands");
```
tylerl0706: You should be able to slim down this file by specifying Command instead of any. It should automatically cast it.

Like this:
https://github.com/PowerShell/vscode-powershell/blob/master/src/features/Console.ts#L10

But I'm not 100%. It would save a lot of code I think.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export class GetCommandsFeature implements IFeature {
+    private command: vscode.Disposable;
+    private languageClient: LanguageClient;
+    private commandsExplorerProvider: CommandsExplorerProvider;
+
+    constructor() {
+        this.command = vscode.commands.registerCommand("PowerShell.GetCommands", () => {
+            if (this.languageClient === undefined) {
+                // We be screwed
```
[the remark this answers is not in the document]
corbob: You mean like line 19 of ExpandAlias.ts? I completely agree we should log this... I'll change the comment to the TODO for now, but we should log it prior to merge.

package.json
corbob: This has been updated with the PowerShell Logo. Properly sized and everything 😀

src/features/GetCommands.ts
corbob: Assume you mean the tslint. I agree! This was never meant to make it into a commit, but apparently it did. Removed now.

src/features/GetCommands.ts
corbob: I'm not sure how to do this if possible. I attempted to replace the any with Command and it crashes PSES among other things.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,132 @@
...
+                return;
+            }
+            this.languageClient.sendRequest(GetCommandsRequestType, "").then((result) => {
+                const toCommand = (command: any): Command => {
```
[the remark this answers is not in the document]
corbob: If we need to keep the map, I will pull it out into a standalone function. Currently not able to get the casting that @tylerl0706 spoke of to work.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    public getTreeItem(): vscode.TreeItem {
+        return {
+            label: this.Name,
+            collapsibleState: this.collapsibleState,
+        };
+    }
+
+    public async getChildren(element): Promise<Command[]> {
+        if (undefined !== this.populatedCommands) {
```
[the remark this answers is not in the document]
corbob: This was resolved by removing the class that was used for troubleshooting and left in the code 😀

src/features/GetCommands.ts
```diff
@@ -0,0 +1,132 @@
...
+                    );
+                };
+                this.commandsExplorerProvider.powerShellCommands = result.map(toCommand);
+                // this.commandsExplorerProvider.refresh();
+            });
+        });
+        this.commandsExplorerProvider = new CommandsExplorerProvider();
+        vscode.window.registerTreeDataProvider("powerShellCommands", this.commandsExplorerProvider);
+        vscode.commands.registerCommand(
+            "PowerShell.refreshCommandsExplorer",
```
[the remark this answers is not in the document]
corbob: Good call, it looks like I did for the other command, just not this one.

package.json
corbob: Done!

corbob:
v1 of this is ready for review. 

One thing that is bothering me is: when you select a command, right click and choose to Insert Command, it inserts at the current location in editor, but does not focus editor, and it selects the entire command that is insert instead of putting cursor at the end of the inserted command.

media/PowerShell.svg
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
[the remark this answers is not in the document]
Jaykul: Just curious, why would you mark svg files as binary?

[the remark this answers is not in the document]
rjmholt: Might be misguided, but I imagine the SVG is not going to be created or maintained with a text editor. Marking it as binary would hopefully mean we can treat it as a generated, checked-in blob. So ideally we can avoid:
- Big local git diffs if the SVGs are changed
- Big diffs to scroll past in PRs that touch them
- Merge conflicts if they get edited in both branches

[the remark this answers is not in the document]
Jaykul: Makes sense, I guess! Thanks 👍

package.json
```diff
@@ -86,6 +104,20 @@
         "title": "Expand Alias",
         "category": "PowerShell"
       },
+      {
+				"command": "PowerShell.RefreshCommandsExplorer",
+				"title": "Refresh",
```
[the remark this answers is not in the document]
corbob: Should be fixed with vscode's auto-formatting 😃

src/features/GetCommands.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    public getTreeItem(): vscode.TreeItem {
+        return {
+            label: this.label,
+            collapsibleState: this.collapsibleState,
+        };
+    }
+
+    public async getChildren(element): Promise<Command[]> {
+        return [];
```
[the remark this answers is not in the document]
corbob: I think it's because we have to return something and we're not actually using the tree view part just yet... I think ideally we'll start using getChildren to treeview the modules, but right now it has to be something.

src/features/ShowOnlineHelp.ts
```diff
@@ -15,20 +15,24 @@ export class ShowHelpFeature implements IFeature {
...
-            const text = doc.getText(cwr);
+                const selection = editor.selection;
+                const doc = editor.document;
+                const cwr = doc.getWordRangeAtPosition(selection.active);
+                const text = doc.getText(cwr);
 
-            this.languageClient.sendRequest(ShowOnlineHelpRequestType, text);
+                this.languageClient.sendRequest(ShowOnlineHelpRequestType, text);
+            } else {
+                this.languageClient.sendRequest(ShowOnlineHelpRequestType, item.Name);
```
tylerl0706: Is this a part of the PSES PR that adds the extra logic for OnlineHelp?

src/features/GetCommands.ts
```diff
@@ -0,0 +1,105 @@
...
+    private languageClient: LanguageClient;
+    private commandsExplorerProvider: CommandsExplorerProvider;
+
+    constructor() {
+        this.command = vscode.commands.registerCommand("PowerShell.RefreshCommandsExplorer", () => {
+            if (this.languageClient === undefined) {
+                // TODO: Log error message
+                return;
+            }
+            this.languageClient.sendRequest(GetAllCommandsRequestType, "").then((result) => {
```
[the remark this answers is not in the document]
tylerl0706: the network call is in the anonymous function - not the constructor

src/features/GetCommands.ts
```diff
@@ -0,0 +1,104 @@
...
+                editBuilder.replace(range, item.Name);
+            });
+        });
+    }
+
+    public dispose() {
+        this.command.dispose();
+    }
+
+    public setLanguageClient(languageclient: LanguageClient) {
```
tylerl0706: Where does this function get called?

tylerl0706:
I didn't see the "powerShell/getCommand" message used at all in this PR even though it's in the PSES PR. Is that on purpose?

src/features/ShowOnlineHelp.ts
corbob: No, I don't believe these conflict as this is just adding logic to what we send on if they're selecting the text or not. Based on your suggestion here: https://github.com/PowerShell/PowerShellEditorServices/pull/697#discussion_r213069053 I would be more inclined to change these too PRs back to WIP and refactor that to a single message call.

corbob:
The "powerShell/getCommand" message was added in PSES as a precursor to that we will need it on this side, but the next step I believe will be mostly on this side. However, based on the suggestion of a single message I think it's the better way.

src/features/GetCommands.ts
corbob: If you're referring to setLanguageClient, I believe this is called by vscode itself, I'd have to look at the API docs, but I do know it gets called (or at least a breakpoint in it triggers)

corbob:
Changing PR back to WIP to address the refactoring @tylerl0706 suggested.

tylerl0706:
@Corbob, since "powerShell/getCommand" is not used in this feature (yet). You could just remove the message from the PSES PR so we can get this in 😊

Then V2 of the feature could do the refactoring.

If you want to do the refactoring now, that's great but I won't block on that 😊

src/features/GetCommands.ts
rkeithhill: This method is part of the PowerShell extension infrastructure defined in feature.ts.   When a PowerShell session is started (in session.ts), it starts the language client.  When the language client is ready, the SessionManager walks through its array of features and updates each with the new language client.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+        return element;
+    }
+
+    public getChildren(element?: Command): Thenable<Command[]> {
+        return Promise.resolve(this.powerShellCommands || []);
+    }
+
+}
+
+// TODO: Define and export an ICommand interface to describe the properties we require.
```
rkeithhill: Yeah, that would be nice.  :-)

src/main.ts
```diff
@@ -139,7 +141,6 @@ export function activate(context: vscode.ExtensionContext): void {
         new CustomViewsFeature(),
         new FoldingFeature(logger, documentSelector),
     ];
-
```
rkeithhill: What's the point of getting rid of this blank line?

src/features/ShowOnlineHelp.ts
corbob: These only conflicted this time around because we renamed the request 😄

src/features/GetCommands.ts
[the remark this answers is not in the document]
corbob: getChildren is defined by the API and must return something: https://code.visualstudio.com/docs/extensionAPI/vscode-api#TreeDataProvider.getChildren

src/main.ts
corbob: Because reasons?

Honestly, not sure. Must have been by accident.

corbob:
I think I've got most of the stuff. Only thing I'm not 100% on is some of the comment based documentation parts.

src/features/GetCommands.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    public getTreeItem(element: Command): vscode.TreeItem {
+        return element;
+    }
+
+    public getChildren(element?: Command): Thenable<Command[]> {
+        return Promise.resolve(this.PowerShellCommands ? this.PowerShellCommands : []);
+    }
+
+}
```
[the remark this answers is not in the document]
corbob: I'm not sure what is meant by needing a space here. The function after this is indented the same amount (0), and there is a blank line before and after the closing brace here. Also not getting any changes when trying to format the document.

src/features/GetCommands.ts
corbob: As @rkeithhill said, it is called by the SessionManager. It doesn't work well (at all) without the function.

src/features/GetCommands.ts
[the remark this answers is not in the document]
corbob: As @tylerl0706 mentioned, the network call is in the anonymous function defined on line 19.

Do we still want to consider a later refactor of it? I can add a TODO comment if we'd like to do that. Otherwise we can resolve the conversation and carry on.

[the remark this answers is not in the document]
rjmholt: Ah, I see now. Well it might be worth spinning it out into a non-anonymous function — keeps the code out of the constructor and prevents silly people like me from misreading it.

src/features/GetCommands.ts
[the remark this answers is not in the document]
rjmholt: Hmmm not sure either. Only thing I'd change is to take the newline between the inner and outer closing brace out though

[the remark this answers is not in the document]
corbob: Took out the empty line :)

src/features/GetCommands.ts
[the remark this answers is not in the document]
corbob: Spun this and the InsertCommand anonymous function into named functions 😄

src/features/GetCommands.ts
```diff
@@ -0,0 +1,105 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+import * as vscode from "vscode";
+import { LanguageClient, RequestType } from "vscode-languageclient";
+import { IFeature } from "../feature";
+
+export const GetAllCommandsRequestType = new RequestType<any, any, void, void>("powerShell/getAllCommands");
```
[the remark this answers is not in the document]
corbob: I thought I tried it, but apparently not hard enough. I've been able to eliminate both of the any instances here 👍

package.json
```diff
@@ -58,6 +59,23 @@
     "test": "node ./node_modules/vscode/bin/test"
   },
   "contributes": {
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "PowerShellCommandExplorer",
+          "title": "PowerShell Command Explorer",
+          "icon": "media/PwSh.svg"
```
TylerLeonhardt: if it's not too much trouble can we rename this to `pwsh.svg`?

corbob: Done 👍

PR #1430, opened 2018-07-12
(GH-1428) Make region folding case insensitive and strict whitespace 

### LINKED ISSUE #1428: PowerShell Extension 1.8.1 breaks editor folding on region tags
### System Details

- Operating system name and version:  Windows 10.1803.17134.137
- VSCode Version: 1.25.0
- PowerShell extension version: 1.8.1
PSVersion                      5.1.17134.137
BuildVersion                   10.0.17134.137

$pseditor.EditorServicesVersion
code --list-extensions --show-versions
$PSVersionTable
1.25.0
0f080e5267e829de46638128001aeb7ca2d6d50e
x64

Major  Minor  Build  Revision
-----  -----  -----  --------
1      8      1      0
cdsama.tortoise-svn-for-vscode@0.1.1
dakara.transformer@1.6.0
emmanuelbeziat.vscode-great-icons@2.1.36
GrapeCity.gc-excelviewer@2.1.24
ilich8086.coldfusion@1.0.1
mechatroner.rainbow-csv@0.4.2
ms-vscode.csharp@1.15.2
ms-vscode.PowerShell@1.8.1
PKief.material-icon-theme@3.5.1
schneiderpat.aspnet-helper@0.6.4
tuxtina.json2yaml@0.2.0

Key   : PSVersion
Value : 5.1.17134.137
Name  : PSVersion

Key   : PSEdition
Value : Desktop
Name  : PSEdition

Key   : PSCompatibleVersions
Value : {1.0, 2.0, 3.0, 4.0...}
Name  : PSCompatibleVersions

Key   : BuildVersion
Value : 10.0.17134.137
Name  : BuildVersion

Key   : CLRVersion
Value : 4.0.30319.42000
Name  : CLRVersion

Key   : WSManStackVersion
Value : 3.0
Name  : WSManStackVersion

Key   : PSRemotingProtocolVersion
Value : 2.3
Name  : PSRemotingProtocolVersion

Key   : SerializationVersion
Value : 1.1.0.1
Name  : SerializationVersion

### Issue Description

Searched for existing issues, didn't find anything.
Others in the chat room also confirmed
Reproduced with latest Insiders Build of vscode.

Steps to Reproduce:

1.  Install vscode
2. Open a powershell script file containing working region and endregion tags
3. Observe that the editor folding works
4.  Install Powershell Extension (ms-vscode.powershell) version 1.8.1
5. Relaunch vscode
6. Observe that editor folding is not working

glennsarti:
## PR Summary

Previously the code folding feature would fail to find region blocks starting
with `Region` or ending with `EndRegion`.  This was due to the regex only
looking for lower case.  This commit updates the start and end region
detection to be case insensitive, similar to that defined in
https://github.com/Microsoft/vscode/commit/64186b0a262f0ff89a060cf8dbbf8de7ff831a00

This commit adds has strict whitespace rules for the region/endregion keyword
which means only `#region` will match whereas `# region` will not.

Previously the code folding had trouble detecting the beginning of comment
blocks and regions which were not indented.  This was due to the empty line
regex expecting at least one whitespace character to be considered "an empty
line".  This commit modifies the empty line regex to also allow an empty string
which is indeed, an empty line.

Fixes #1428

## PR Checklist

- [x] Summarized changes

### DIFF
--- a/src/features/Folding.ts
+++ b/src/features/Folding.ts
@@ -354,7 +354,7 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ILineNumberRangeList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
+        const emptyLine = /^\s*$/;
 
         let startLine: number = -1;
         let nextLine: number = -1;
@@ -421,9 +421,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ITokenList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
-        const startRegionText = /^#\s*region\b/;
-        const endRegionText = /^#\s*endregion\b/;
+        const emptyLine = /^\s*$/;
+        const startRegionText = /^#region\b/i;
+        const endRegionText = /^#endregion\b/i;
 
         tokens.forEach((token) => {
             if (token.scopes.indexOf("punctuation.definition.comment.powershell") !== -1) {
--- a/test/features/folding.test.ts
+++ b/test/features/folding.test.ts
@@ -38,16 +38,18 @@ suite("Features", () => {
 
         suite("For a single document", async () => {
             const expectedFoldingRegions = [
-                { start: 1,  end: 6,  kind: 1 },
-                { start: 7,  end: 51, kind: 3 },
-                { start: 8,  end: 13, kind: 1 },
-                { start: 14, end: 17, kind: 3 },
-                { start: 19, end: 22, kind: 3 },
-                { start: 26, end: 28, kind: 1 },
-                { start: 30, end: 40, kind: 3 },
-                { start: 32, end: 36, kind: 3 },
-                { start: 42, end: 44, kind: 3 },
-                { start: 47, end: 50, kind: 3 },
+                { start: 0,  end: 4,  kind: 3 },
+                { start: 1,  end: 3,  kind: 1 },
+                { start: 10, end: 15, kind: 1 },
+                { start: 16, end: 60, kind: 3 },
+                { start: 17, end: 22, kind: 1 },
+                { start: 23, end: 26, kind: 3 },
+                { start: 28, end: 31, kind: 3 },
+                { start: 35, end: 37, kind: 1 },
+                { start: 39, end: 49, kind: 3 },
+                { start: 41, end: 45, kind: 3 },
+                { start: 51, end: 53, kind: 3 },
+                { start: 56, end: 59, kind: 3 },
             ];
 
             test("Can detect all of the foldable regions in a document with CRLF line endings", async () => {
--- a/test/fixtures/folding-crlf.ps1
+++ b/test/fixtures/folding-crlf.ps1
@@ -1,3 +1,12 @@
+#RegIon This should fold
+<#
+Nested different comment types.  This should fold
+#>
+#EnDReGion
+
+# region This should not fold due to whitespace
+$shouldFold = $false
+#    endRegion
 function short-func-not-fold {};
 <#
 .SYNOPSIS
@@ -30,7 +39,7 @@ double quoted herestrings should also fold
 
   #region This fools the indentation folding.
   Write-Host "Hello"
-    # region Nested regions should be foldable
+    #region Nested regions should be foldable
     Write-Host "Hello"
     # comment1
     Write-Host "Hello"
@@ -38,7 +47,7 @@ double quoted herestrings should also fold
     Write-Host "Hello"
     # comment2
     Write-Host "Hello"
-    # endregion
+    #endregion
 
   $c = {
     Write-Host "Script blocks should be foldable"
--- a/test/fixtures/folding-lf.ps1
+++ b/test/fixtures/folding-lf.ps1
@@ -1,3 +1,12 @@
+#RegIon This should fold
+<#
+Nested different comment types.  This should fold
+#>
+#EnDReGion
+
+# region This should not fold due to whitespace
+$shouldFold = $false
+#    endRegion
 function short-func-not-fold {};
 <#
 .SYNOPSIS
@@ -30,7 +39,7 @@ double quoted herestrings should also fold
 
   #region This fools the indentation folding.
   Write-Host "Hello"
-    # region Nested regions should be foldable
+    #region Nested regions should be foldable
     Write-Host "Hello"
     # comment1
     Write-Host "Hello"
@@ -38,7 +47,7 @@ double quoted herestrings should also fold
     Write-Host "Hello"
     # comment2
     Write-Host "Hello"
-    # endregion
+    #endregion
 
   $c = {
     Write-Host "Script blocks should be foldable"


src/features/Folding.ts
```diff
@@ -354,7 +354,7 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ILineNumberRangeList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
+        const emptyLine = /^[\s]*$/;
```
rkeithhill: Does JS really require the `[]` here?  Wouldn't `/^\s*$/` be sufficient?

src/features/Folding.ts
```diff
@@ -421,9 +421,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ITokenList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
-        const startRegionText = /^#\s*region\b/;
-        const endRegionText = /^#\s*endregion\b/;
+        const emptyLine = /^[\s]*$/;
```
rkeithhill: Same comment as above.

src/features/Folding.ts
```diff
@@ -421,9 +421,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
...
+        const startRegionText = /^#\s*[rR]egion\b/;
```
rkeithhill: Does JS support `(?i)` to set the regex to ignore case?

src/features/Folding.ts
glennsarti: Huh....Good point.  Will test

src/features/Folding.ts
glennsarti: Depends.  This commit brings it into line with the VSCode "default" behaviour, but, we could add this.  I'm not fussed either way on this issue.

rkeithhill: Good enough for me. FWIW, ISE only supports lower-case `#region / #endregion`.

rkeithhill: OTOH leading folks down a path that doesn't work in ISE … is that a bad thing or if so, do we care?  Also, ISE doesn't like a space between the `#` and `region|endregion`.  It's pretty picky.

rkeithhill: Agreed!

src/features/Folding.ts
glennsarti: Fixed

src/features/Folding.ts
glennsarti: Fixed

src/features/Folding.ts
glennsarti: Mostly agreed.  I've updated the regexes however I've kept the "space between `#` and `region`" e.g. `# region` is allowed.

While ISE and other langauges don't allow a space, the existing VS Code default for powershell is to allow zero or more space (https://github.com/Microsoft/vscode/blob/master/extensions/powershell/language-configuration.json#L26-L31).  Removing that is a kinda-breaking change.

glennsarti:
Ready for review/merge again.

src/features/Folding.ts
rkeithhill: Not sure I agree with the VSCode defaults.  Adding the space allows me to write a normal comment that just happens to begin with the word region but isn't the start of a "region" e.g. 
```
# blah blah blah blah in the same 
# region as blah blah blah
```

glennsarti: yeah .... could also argue the same for;

```
#blah blah blah blah in the same 
#region as blah blah blah
```

rkeithhill: Except that if I type a comment like `#Requires that the user run as administrator` at the top of my file, I can fix it by adding the space e.g. `# Requires that the user run as administrator`

glennsarti: True....

glennsarti: Interesting the tests for that, don't include spaces between `#`
https://github.com/Microsoft/vscode/commit/9e05d4b635b403cdb94bb5fe6600e1d9710ab320#diff-e10026493ca8c706f842f8b49b1ccdfa

rkeithhill: I also find it odd that they allow any number of spaces after the `#` e.g. `#                Region` would be a valid region start.

glennsarti: So given all this, plus the powershell region snippet has no space (https://github.com/PowerShell/vscode-powershell/blob/705f1747844a8f11f187cd008622f175094c2b72/snippets/PowerShell.json#L947-L955) I'll change the regex to not allow whitespace.

Can always add it back in :-)

src/features/Folding.ts
```diff
@@ -421,9 +421,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ITokenList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
-        const startRegionText = /^#\s*region\b/;
-        const endRegionText = /^#\s*endregion\b/;
+        const emptyLine = /^\s*$/;
+        const startRegionText = /^#\s*region\b/i;
```
rjmholt: I still think we shouldn't match `#    region` for this; I can imagine ordinary comments matching this, and I don't think I've ever seen any usage of a pragma that permitted whitespace like this.

glennsarti: May want to hit up the VSCode team on that, as they're ALL like that

src/features/Folding.ts
glennsarti: Fixed.

glennsarti:
Pending Green CI, this is ready for review again.

rkeithhill:
LGTM!

src/features/Folding.ts
rjmholt: I can only find examples where the whitespace comes before the `#`, like:
* [The extension folding snippets](https://github.com/Microsoft/vscode/search?q=%22%2F%23region%22&unscoped_q=%22%2F%23region%22)
* [The asks in the original issue](https://github.com/Microsoft/vscode/issues/12146)
* [The feature announcement](https://code.visualstudio.com/updates/v1_17#_folding-regions)
* [VSCode's folding tests](https://github.com/Microsoft/vscode/blob/0faa5d117cced88ab9d507cadf697c2acd3dcbff/src/vs/editor/contrib/folding/test/foldingModel.test.ts)
* [VSCode's test definition of `FoldingMarkers`](https://github.com/Microsoft/vscode/blob/059ead3c7bf1173477a73011d722b27384989448/src/vs/editor/contrib/folding/test/indentRangeProvider.test.ts#L150-L153)

Having an indented `#region` like:
```
if ($x)
{
    #region

       DoStuff

    #endregion
}
```

seems pretty reasonable to me.

But I feel like setting folds for
```
#   Region
if ($x)
{
    ... Blah
}

# EndRegion
```
could cause problems. Any line comment beginning with the word "region" would become a fold, meaning we would then look for `#   endregion`, and may not find one.

glennsarti: Oh I am so sorry...I need glasses 👎 

-100 points to me.

rjmholt: Haha, no I thought that's what was going on, just wanted to be sure 😄 

Thanks so much for this @glennsarti!

src/features/Folding.ts
```diff
@@ -421,9 +421,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
     ): ITokenList {
         const result = [];
 
-        const emptyLine = /^[\s]+$/;
-        const startRegionText = /^#\s*region\b/;
-        const endRegionText = /^#\s*endregion\b/;
+        const emptyLine = /^\s*$/;
+        const startRegionText = /^#region\b/i;
```
rjmholt: Wait, I think you were right about the whitespace here -- no? That the regex should look like `/^\s*#region/i`?

rjmholt: As in, like the other VSCode folding regexes

glennsarti: Nope....the regex I use is slightly different, because it's checking the grammar token (it's within the line) not the whole line.  That's why I do an Empty Line check.

rjmholt: In the [code below it](https://github.com/glennsarti/vscode-powershell/blob/c61ca6d7561cb13e23398b8cd17a457f9ab58d8c/src/features/Folding.ts#L431-L432), it looks like `subsequentText()` gets the whole comment line and then tests it against the regex, which starts with `^`. Which I read as not matching an indented `#region`. Does that sound right?

glennsarti: So the logic I'm using is;

```
    Find the `#` character (This is the `punctuation.definition.comment.powershell` scope)
        If there is only whitespace before the `#` then
            if the text after, and including the `#` matches the starting regex then
                #Do Something
            end
            if the text after, and including the `#` matches the ending regex then
                #Do Something
          end
       end
     end
```

That said, it could be optimized by grabbing the whole line and regex-ing on that...but that's for another PR.

rkeithhill: Oh, I see.  I think we want `/^\s*#region\b/i`, right?  We don't want variable # spaces between the `#` and `region|endregion` but we do want to allow that *before* the `#`.

glennsarti: Also, looks like the `subsequentText()` named is wrong.  Should be `textFrom` or something like that.

rjmholt: Oh, so it does this:
```
    #   Some text is here
    ^                   ^
    [offset]            [end of line]
```

`subsequentTest()` seems as good a name as any - but maybe the comment should be something like:
> Get the text from a given offset to the end of the line it's on

glennsarti: yup

rjmholt: Anyway, I'm satisfied!

glennsarti: @rjmholt PR https://github.com/PowerShell/vscode-powershell/pull/1438 refactors the detection to be easier to understand.

PR #1438, opened 2018-07-16
(GH-1437) Fix detecting contiguous comment blocks and regions

### LINKED ISSUE #1437: Syntax folding should differentiate between regions and comment blocks
### System Details

- Operating system name and version: N/A
- VS Code version: N/A
- PowerShell extension version: N/A
- Output from `$PSVersionTable`: N/A

### Issue Description

While working on #1428 I thought of another possible issue with the syntax folding.  Currently regions and comment blocks are detected as the same thing which will cause issues when comments are not separated by empty lines.

Given a script of

``` powershell
# Comment Block 1
# Comment Block 1
# Comment Block 1
#region
# Comment Block 2
# Comment Block 2
# Comment Block 2
$something = $true
#endregion
```

Expected Folding Regions

```powershell
# Comment Block 1  --+-- Folding Line 1-3
# Comment Block 1    |
# Comment Block 1  --+
#region                                   --+-- Folding Line 4-9
# Comment Block 2  --+-- Folding Line 5-7   |
# Comment Block 2    |                      |
# Comment Block 2  --+                      |
$something = $true                          |
#endregion                                --+
```

Actual Folding regions
```powershell
# Comment Block 1  --+-- Folding Line 1-7
# Comment Block 1    |
# Comment Block 1    |
#region              |  --+-- Folding Line 4-9
# Comment Block 2    |    |
# Comment Block 2    |    |
# Comment Block 2  --+    |
$something = $true        |
#endregion              --+
```

glennsarti:
## PR Summary

Previously the region comment detection used a little convoluted method to
detect regions in a document.  This commit simplifies the detection by
extracting the line from the document and using regex's similar to that used by
the PowerShell language configuration.  This also removes the need for the
emptyline and subsequentText method calls.  While the performance of the folder
is pretty quick, this should in theory make it faster on larger documents by
doing less calls to the VSCode Document API.

---

Previously the syntax folding feature would not correctly identify comment
blocks and comment regions if they appeared all together.  This commit changes
the comment block detection to ignore line comments that start with region and
endregion, i.e. region block start/end directives.  This commit also adds test
for this scenario.

Fixes #1437 

## PR Checklist

- [x] Summarized changes

### DIFF
--- a/src/features/Folding.ts
+++ b/src/features/Folding.ts
@@ -173,6 +173,14 @@ interface ILineNumberRangeList extends Array<LineNumberRange> { }
 export class FoldingProvider implements vscode.FoldingRangeProvider {
     private powershellGrammar: IGrammar;
 
+    /**
+     * These regular expressions are used to match lines which mark the start and end of region comment in a PowerShell
+     * script. They are based on the defaults in the VS Code Language Configuration at;
+     * https://github.com/Microsoft/vscode/blob/64186b0a26/extensions/powershell/language-configuration.json#L26-L31
+     */
+    private readonly startRegionText = /^\s*#region\b/i;
+    private readonly endRegionText = /^\s*#endregion\b/i;
+
     constructor(
         powershellGrammar: IGrammar,
     ) {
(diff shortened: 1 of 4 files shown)


cadayton:
The following was folding previously but not since the extension assume the folding logic.
# Initializing Rountine
   <block of code>
#

glennsarti:
@cadayton It would be better if you raised an issue (https://github.com/PowerShell/vscode-powershell/issues/new/choose) instead of commenting on this PR.

tylerl0706:
@cadayton this is because we've switched away from indentation-based folding to syntax-based folding.

A work around for you would be to use `#region Initialize Routine` and `#endregion`. This is recommended. 

Alternatively, you can simply disable syntax-based folding which will fall back to indentation-based folding. (not at my computer at the moment so I don't have the exact name but just search for "folding" in the settings. 

I'm not sure we really want to support syntax-based folding and indentation-based folding at the same time so I would really encourage you to do one of the above workarounds.

tylerl0706:
Like Glenn said, feel free to open an issue if you feel otherwise! 👍

cadayton:
Thanks Tyler.

I'll try the workaround, but have quite few references that need to be modified to switch over. Not to mention several GitHub projects that need to change too.   Prior to using VS code, I used PowerGUI which required #region & #endregion too and after switching over to VS Code I converted everything to syntax folding since it worked.  It is nice to avoid adding #region and $endregion, but understand if it is making it harder for VS Code parser too difficult to manage.

Looks like I need to switch back now. :(

Craig Dayton
cadayton@protonmail.com
360-466-8778

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #1477, opened 2018-08-14
adding Bits-Transfer & user switch to install latest user profile insiders edition

tabs-not-spaces:
## PR Summary

Quick update,

- Moved the download line from `Invoke-WebRequest` to `Start-BitsTransfer` (better end user experience).
- Added a new `$BuildEdition` value - Insider-User that allows the User Profile Insiders Edition to be installed.
- Renamed current Insiders Edition value in `$BuildEdition` to Insider-System to reflect the two options.

## PR Checklist

- [x] Summarized changes


tabs-not-spaces:
@rjmholt currently the script only works for windows, but is set out in such a way to prepare for Linux / OSX in a future state.

I lean towards using start-bitstransfer as it provides a download progress bar which is beneficial for anyone on a slower internet connection.

Good idea on the async - I'll add that in and handle waiting for the file to complete.

rjmholt:
Yes reading the script it's pretty clear it's only for Windows at the moment, so we can just switch over to `Start-BitsTransfer` and then ideally can work later to bring proper macOS/Linux support

scripts/Install-VSCode.ps1
```diff
@@ -167,13 +175,21 @@ if (!($IsLinux -or $IsOSX)) {
             break;
         }
     }
+    if (($User -eq $true) -and ($BuildEdition -eq "Insider")) {
```
rjmholt: You should be able to do just `if ($User -and ...)` here

scripts/Install-VSCode.ps1
```diff
@@ -122,6 +126,10 @@ param(
     [ValidateSet("stable","insider")]
     [string]$BuildEdition = "stable",
 
+    [Parameter()]
+        [switch]
```
rjmholt: The convention in the rest of the file seems to be putting the type constraint on the same line as the parameter

scripts/Install-VSCode.ps1
```diff
@@ -56,6 +56,10 @@
     A validated string defining which build edition or "stream" to download - stable or
     insiders edition. If the parameter is not used, then stable is downloaded as default.
 
+.PARAMETER User
```
rjmholt: The main function of this flag is to install VSCode-Insiders (user edition) right? Instead of non-Insiders edition. If that's the case, I would vote for the flag being called something like `Insiders` for now, and then if we want to make it more complex (e.g. choose between User Profile and non-User Profile) later, we could add another flag.

rjmholt:
@tabs-not-spaces, this is a really great addition -- thanks for contributing!

scripts/Install-VSCode.ps1
tabs-not-spaces: @rjmholt I agree with you. In that case would it make sense for now then to move from the switch into the `$BuildEdition` parameter as a validation value?

Something like: `[ValidateSet("stable","insider-system","insider-userprofile")]` ??

scripts/Install-VSCode.ps1
tabs-not-spaces: fixed and about to check in

scripts/Install-VSCode.ps1
tabs-not-spaces: fixed and about to check in

scripts/Install-VSCode.ps1
rjmholt: Yeah that's a really good idea!

tabs-not-spaces:
@rjmholt found a major flaw with my code relating to the async behavior of bits transfer. have resolved and validated it now works!

rjmholt:
Thanks for your contribution @tabs-not-spaces!

PR #1548, opened 2018-09-27
Explicitly return undefined from resolveDbgConfig when sessn not started

rkeithhill:
## PR Summary

In the September drop of VSCode this fixes the issue with VSCode
opening launch.json in this case.  Technically using return with no expr 
works but better to be explicit in this case I think.

https://github.com/Microsoft/vscode/issues/54213#issuecomment-420965778

## PR Checklist

- [x] Summarized changes


rkeithhill:
It will be nice in the September drop to not have `launch.json` pop up every time you attempt to start debugging before the PS session has finished initializing!

PR #1591, opened 2018-10-26
PowerShell color themes

mattmcnabb:
## PR Summary
This is to add color themes mimicking those built in to VSCode, but with familiar PowerShell console colors. There are 14 in total including the PowerShell ISE theme that was already included.

powershellthemes_50

## PR Checklist

- [x] Summarized changes
- [ `NA`] PR has tests


rkeithhill:
Also, we don't use tabs anywhere in this codebase (AFAIK).  Perhaps you could update the theme JSON files to replace `tab` with four spaces and then auto-format them?

mattmcnabb:
Crap! Didn't think about the tabs! I'll get that fixed up tomorrow.

It is possible to inherit theme settings, but I'd have to figure out how to get the built in theme file paths right.

Maybe this should be a separate extension? I could possibly add another console theme for the core folks, but I guess I didnt realize there was an official color scheme.

I'm fine to hold on this for the moment. Thanks!

rkeithhill:
It might be cool if the console colors switched depending on whether you have Windows PowerShell (blue) or PowerShell Core (black-ish) loaded.  I don't know if that is possible but it would be another hint as to which "PowerShell" is running.

mattmcnabb:
> It might be cool if the console colors switched depending on whether you have Windows PowerShell (blue) or PowerShell Core (black-ish) loaded. I don't know if that is possible but it would be another hint as to which "PowerShell" is running.

I have a couple of ideas about better ways to do this:
1. An extension could potentially configure settings in the user settings, and trigger changes to those values based on an event. However, I don't believe the color theme settings are available in the user settings.

It might be possible to dynamically generate a color theme whenever a user changes the theme in VSCode, but this seems a little heavy-handed.

The only other route I can think of at the moment is to simply break this out into a third party extension. That also might give me the chance to create additional settings around the Core terminal colors.

rjmholt:
@rkeithhill Re: inheritance: the first idea that comes to mind is tracking the themes as JS/TS classes and writing an instance of each out at build time... A bit hacky and complicated though perhaps

mattmcnabb:
> @rkeithhill Re: inheritance: the first idea that comes to mind is tracking the themes as JS/TS classes and writing an instance of each out at build time... A bit hacky and complicated though perhaps

@rjmholt the color themes already have support for the "include" keyword to inherit another color theme, but the problem is getting the right path for the built-in themes. These live in the VSCode install folder, and there is an environment variable "appRoot" that I could use to reference this. However, I can't get the theme files to evaluate the variable in the way that a tasks.json file would.

Example "include:"

```
{
    "$schema": "vscode://schemas/color-theme",
    "name": "PowerShell Abyss",
    "include": "./ps-dark_defaults.json",
	"tokenColors": [
		{
            "scope": [
                "meta.embedded",
```

tylerl0706:
@Tyriar - do you think you could add someone who might be able to help here?

Basically we want to extend the vscode default themes with some updates to the colors in the console.

Not sure if there's a good workflow for this

Tyriar:
The only way to do this currently is to add some user settings like this: https://github.com/Tyriar/vscode-snazzy#install

Maybe you could add a command to add the PS theme colors to your user settings? This seems preferable to getting into the business of tracking VS Code's themes and shipping a bunch of duplicate themes (also then you support build-in themes as well).

So yeah, I'd recommend a command like "Add PowerShell Terminal Theme Colors to User Settings" which adds a bunch of keys to `workbench.colorCustomizations`.

tylerl0706:
oooo I like that. Then any user can keep any theme they want in their editor - but have a powershell.exe-like theme in the terminal.

mattmcnabb:
@tyriar i agree editing the user settings is probably the way to go. I thought about this but couldn't find where the user settings could include the colorizations. I'll check into this and see if I can figure out a command.

tylerl0706:
It should be just a little bit of typescript 😄 if you need help, we're here to help!

mattmcnabb:
@tylerl0706 @Tyriar I made some headway on this, and I have a command that can set user settings, but I can't quite figure out how to set the workbench colors. They're nested and I can't find the right way to configure these with the vscode.workspace API. What I have now is something like this:

```
function changeColor(settingName: string, newValue: any, global: boolean = false): Thenable<void> {
    const configuration: vscode.WorkspaceConfiguration =
        vscode.workspace.getConfiguration("workbench.colorCustomizations");

    return configuration.update(settingName, newValue, global);
}

return changeColor("terminal.background", "#012456", true);
```

But this returns the error:

Unable to write to User Settings because workbench.colorCustomizations.terminal.background is not a registered configuration.

I also tried:

```
return vscode.workspace.getConfiguration().update("workbench.colorCustomizations.terminal.background", "#012456", true);
```

But that gave me the same error. Any thoughts on how this might be possible?

Tyriar:
@sandy081 what's the best way to set nested settings like this, while at the same time not overriding existing keys that already exist (we want to write to `workbench.colorCustomizations.terminal.background`, not replace `workbench.colorCustomizations`).

mattmcnabb:
Kinda thinking this might be a dead end. The settings gui doesn't seem to support editing any nested settings. They all just launch in the json file for direct editing. Maybe the API doesn't support the nested keys?

sandy081:
You can only edit registered keys like `workbench.colorCustomizations` but not nested keys through API.

Tyriar:
@tylerl0706 that seems like the way to go then.

mattmcnabb:
> Could we get the entirety of `workbench.colorCustomizations`,
> change the `workbench.colorCustomizations.terminal.background`,
> then apply the edit to `workbench.colorCustomizations`?

@tylerl0706 Are you asking me or @sandy081? If that is possible I'll take a stab at it, but given my lack of Typescript skills it might be a challenge.

I have my current changes local, but I'll see if I can push them up later tonight to make it easier to review.

mattmcnabb:
@tylerl0706 @rkeithhill here's an inelegant attempt that kind of works:

https://github.com/mattmcnabb/vscode-powershell/blob/terminalColors/src/features/TerminalColors.ts

This seems to work if the `"terminal.background"` or `"terminal.foreground"` settings are not configured. If they are configured then nothing happens. Any thoughts on what I'm doing wrong (probably everything)? :)

Also it throws an error if the PowerShell terminal session has not started. I was hoping it would handle that gracefully but I suppose it doesn't. I can probably figure out how to get around that, and maybe just apply the blue background if it can't detect which edition is running.

Please let me know what you think.

tylerl0706:
This looks good! 

Instead of grabbing the version of PowerShell, what about just having a vscode setting instead? Then people using Core can still have the classic color.

Because then the setting could be:

`powershell.terminalTheme` and have `None` (aka use the theme), `Classic` (Windows PowerShell), `Core` (PowerShell Core)

Or maybe do both. The setting, and the command that will query the version of PowerShell and apply Classic or Core

Thoughts?

tylerl0706:
> This seems to work if the "terminal.background" or "terminal.foreground" settings are not configured. If they are configured then nothing happens. Any thoughts on what I'm doing wrong (probably everything)? :)

I'm not sure :/ maybe it's a different method to use? Might have to lean on @Tyriar to answer that question.

mattmcnabb:
> This looks good!
> 
> Instead of grabbing the version of PowerShell, what about just having a vscode setting instead? Then people using Core can still have the classic color.
> 
> Because then the setting could be:
> 
> `powershell.terminalTheme` and have `None` (aka use the theme), `Classic` (Windows PowerShell), `Core` (PowerShell Core)
> 
> Or maybe do both. The setting, and the command that will query the version of PowerShell and apply Classic or Core
> 
> Thoughts?

yeah, but that means I have to figure out how to wire up a setting!

Seriously, I did plan on asking about the possibility of implementing this via a setting, but wanted to make sure the behavior was solid first. The failure of the "Update" function to actually update the settings has me puzzled.

So I think I'd need to somehow set this up to execute on a trigger, like a settings change, right? More research needed...

TylerLeonhardt:
@mattmcnabb don't want this to die! Can you open an issue in VSCode about this:

```
This seems to work if the "terminal.background" or "terminal.foreground" settings are not configured. If they are configured then nothing happens. Any thoughts on what I'm doing wrong (probably everything)? :)
```

I don't want to randomize @Tyriar too much :)

mattmcnabb:
@TylerLeonhardt thanks for sticking with me on this one! I haven't had time to revisit this since the last update.

I'm wondering now if it's even worth pursuing, since the payoff will be so minimal? Like we could literally just advise users to configure one color setting in their settings file to achieve this, rather than implement a new setting that will set it for them. Thoughts?

I think the only possible benefit would be if it opens up new opportunities down the road for additional PowerShell-related customizations, but I'm not sure what those would be.

Tyriar:
> This seems to work if the "terminal.background" or "terminal.foreground" settings are not configured. If they are configured then nothing happens. Any thoughts on what I'm doing wrong (probably everything)? :)

Are you sure it's not a programming error on your side? If you can repro this when changing the setting.json manually can you create an issue? Or maybe it's a problem with the `workspace.update` call?

mattmcnabb:
> Are you sure it's not a programming error on your side?

I'm certain of it ;)

I did test this manually - I set the terminal background color manually in the user settings file, then attempted to run this command to change it, with no effect. Is that what you meant?

Tyriar:
> I did test this manually - I set the terminal background color manually in the user settings file, then attempted to run this command to change it, with no effect. Is that what you meant?

Yeah, could you open an issue with a repro (the settings you tried to use)?

TylerLeonhardt:
@mattmcnabb  and link it to this issue so we can bother @Tyriar and company to look at it 😄

mattmcnabb:
So I think I now have a working setting that will configure the terminal background color based on user's preference.

image


However, this only applies on window load. Any thoughts on how to achieve getting this to work on settings change, or at least on PS session restart?

TylerLeonhardt:
@mattmcnabb do you think you can open a PR (or reuse this one) with these new changes?

I'd be able to help further. Unless @Tyriar knows off the top of his head.

Tyriar:
Aren't you just rewriting settings? The settings files are watched so we can react to changes.

mattmcnabb:
> Aren't you just rewriting settings? The settings files are watched so we can react to changes.

Well, I'm writing a setting that rewrites settings. So the PS extension reads these settings in on load and sets the color customizations. I think I'll need to leverage some sort of event or trigger to actually change the colors immediately, right?

Tyriar:
Ah, you could use this API to watch the settings file, not sure if there's a convenient way to get the settings paths though:

https://github.com/Microsoft/vscode/blob/a684fe7ee136f89d92fa25ee0b8f9bdeacd104b6/src/vs/vscode.d.ts#L5528

TylerLeonhardt:
@Tyriar yikes - that gets tricky with workspace/user level settings (though this probably is mostly a user setting) ... OS's...

Maybe we could have a command in the command pallet that will trigger the color change and set the `powershell.terminalColorTheme` for future start ups?

mattmcnabb:
@TylerLeonhardt Lol... that's kind of back to where I started!

Tyriar:
Think I found what you actually want (I searched for the wrong words last time):

https://github.com/Microsoft/vscode/blob/3a0be849744cab03679d1f25b6f3823bf6611425/src/vs/vscode.d.ts#L7602-L7605

TylerLeonhardt:
That's perfect!

TylerLeonhardt:
I'm gonna close this for now - but @mattmcnabb I think you're close to what we want in your other PR.

PR #1597, opened 2018-11-01
Make the Install-VSCode script work on *nix

rjmholt:
## PR Summary

Made a few changes so the `Install-VSCode.ps1` script supports macOS and various Linux distros as well.

It's probably not exhaustive and could do with a bit of testing.

I added a `-WhatIf` parameter as well, so you can use that to test and it will do the downloads but won't install anything.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/scripts/Install-VSCode.ps1
+++ b/scripts/Install-VSCode.ps1
@@ -579,7 +579,7 @@ try {
     elseif ($IsLinux -or $IsMacOS) {
         # On *nix we need to install extensions as the user -- VSCode refuses root
         $extsSlashes = $extensions -join '/'
-        sudo -H -u $env:SUDO_USER pwsh -c "`$exts = $extsSlashes -split '/'; foreach (`$e in `$exts) { $codeExePath --install-extension `$e }"
+        sudo -H -u $env:SUDO_USER pwsh -c "`$exts = '$extsSlashes' -split '/'; foreach (`$e in `$exts) { $codeExePath --install-extension `$e }"
     }
     else {
         foreach ($extension in $extensions) {
--- a/scripts/Install-VSCode.ps1
+++ b/scripts/Install-VSCode.ps1
@@ -144,7 +144,17 @@ param(
     [switch]$WhatIf
 )
 
-function Test-IsOsX64 {
+# Taken from https://code.visualstudio.com/docs/setup/linux#_installation
+$script:VSCodeYumRepoEntry = @"
+[code]
+name=Visual Studio Code
+baseurl=https://packages.microsoft.com/yumrepos/vscode
+enabled=1
+gpgcheck=1
+gpgkey=https://packages.microsoft.com/keys/microsoft.asc
+"@
+
+function Test-IsOsArchX64 {
     if ($PSVersionTable.PSVersion.Major -lt 6) {
         return (Get-CimInstance -ClassName Win32_OperatingSystem).OSArchitecture -eq "64-bit"
     }
@@ -152,14 +162,19 @@ function Test-IsOsX64 {
     return [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture -eq [System.Runtime.InteropServices.Architecture]::X64
 }
 
-function Get-LinuxReleaseInfo {
-    if (-not (Test-Path '/etc/*-release')) {
-        return $null
+function Get-AvailablePackageManager
+{
+    if (Get-Command 'apt' -ErrorAction SilentlyContinue) {
+        return 'apt'
     }
 
-    return Get-Content -Raw '/etc/*-release' -ErrorAction SilentlyContinue `
-        | ConvertFrom-Csv -Delimiter '=' -Header 'Key','Value' `
-        | ForEach-Object { $obj = @{} } { $obj[$_.Key] = $_.Value } { [pscustomobject]$obj }
+    if (Get-Command 'dnf' -ErrorAction SilentlyContinue) {
+        return 'dnf'
+    }
+
+    if (Get-Command 'yum' -ErrorAction SilentlyContinue) {
+        return 'yum'
+    }
 }
 
 function Get-CodePlatformInformation {
@@ -215,18 +230,18 @@ function Get-CodePlatformInformation {
 
     switch ($os) {
         'Linux' {
-            $releaseInfo = Get-LinuxReleaseInfo
+            $pacMan = Get-AvailablePackageManager
 
-            switch ($releaseInfo.NAME) {
-                { 'Ubuntu','Debian' -contains $_ } {
+            switch ($pacMan) {
+                'apt' {
                     $platform = 'linux-deb-x64'
                     $ext = 'deb'
                     break
                 }
 
-                { 'Fedora','CentOS','RedHat' -contains $_ } {
+                { 'dnf','yum' -contains $_ } {
                     $platform = 'linux-rpm-x64'
-                    $ext = 'deb'
+                    $ext = 'rpm'
                     break
                 }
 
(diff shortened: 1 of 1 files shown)


scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,402 @@ param(
     [ValidateSet("Stable", "Insider-System", "Insider-User")]
     [string]$BuildEdition = "Stable",
 
-    [Parameter()]
-    [switch]$User,
```
rjmholt: This parameter didn't seem to be used...

tylerl0706: Hmm sure it wasn't used for User vs System install?

rjmholt: Yeah I had a search through. `$BuildEdition` already encodes that information, and it looks like `$User` meant to do that but got superseded and left in

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,400 @@ param(
...
-        "64-bit" {
-            if ((Get-CimInstance -ClassName Win32_OperatingSystem).OSArchitecture -eq "64-bit") {
-                $codePath = $env:ProgramFiles
-                $bitVersion = "win32-x64"
+function Test-IsOsX64 {
+    if ($PSVersionTable.PSVersion.Major -lt 6) {
+        return (Get-CimInstance -ClassName Win32_OperatingSystem).OSArchitecture -eq "64-bit"
+    }
+
+    return [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture -eq [System.Runtime.InteropServices.Architecture]::X64
```
tylerl0706: You could probably use `using namespace System.Runtime.InteropServices`

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,400 @@ param(
...
+
+function Get-CodePlatformInformation {
+    param(
+        [Parameter(Mandatory=$true)]
+        [ValidateSet('32-bit', '64-bit')]
+        [string]
+        $Bitness,
+
+        [Parameter(Mandatory=$true)]
+        [ValidateSet('Stable', 'Insider-System', 'Insider-User')]
```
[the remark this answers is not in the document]
tylerl0706: Consider having two parameters.

1. A switch to determine Insiders
2. A valid set (or another switch) between User and System installs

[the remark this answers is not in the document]
tylerl0706: Since this is an internal function, I don't care as much

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,400 @@ param(
...
+            break
+        }
+    }
+
+    switch ($os) {
+        'Linux' {
+            $releaseInfo = Get-LinuxReleaseInfo
+
+            switch ($releaseInfo.NAME) {
+                { 'Ubuntu','Debian' -contains $_ } {
```
tylerl0706: Rip elementary lol If we can confirm that it works on elementary, I'll send a PR... I guess you cover general Linux below

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,400 @@ param(
...
+    [switch]$WhatIf
 )
 
-if (($PSVersionTable.PSVersion.Major -le 5) -or $IsWindows) {
-    switch ($Architecture) {
-        "64-bit" {
-            if ((Get-CimInstance -ClassName Win32_OperatingSystem).OSArchitecture -eq "64-bit") {
-                $codePath = $env:ProgramFiles
-                $bitVersion = "win32-x64"
+function Test-IsOsX64 {
```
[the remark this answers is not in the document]
rjmholt: Have to distinguish from the process being 32/64-bit though

scripts/Install-VSCode.ps1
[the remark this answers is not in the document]
rjmholt: There's no stable user yet. This scheme matches the script parameters and prevents using `-User` with `Stable` for now. I also prefer having two switches, which will make sense later on. But I figure that will make sense to change when a stable user build comes out.

scripts/Install-VSCode.ps1
rjmholt: Is `using namespace` available in PowerShell v3?

scripts/Install-VSCode.ps1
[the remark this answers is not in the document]
rjmholt: And this only checks x86_64/amd64. If something returned arm64 (which `[system.runtime.interopservices.runtimeinformation]::OSArchitecture` can) this function would still return false

scripts/Install-VSCode.ps1
rjmholt: I actually thought about just testing if `apt`/`dpkg` vs `rpm` is available... Maybe we should do that instead?

scripts/Install-VSCode.ps1
tylerl0706: ugh... that's right.

scripts/Install-VSCode.ps1
```diff
@@ -322,14 +341,20 @@ function Get-CodePlatformInformation {
...
+    $info = @{
         AppName = $appName
         ExePath = $exePath
         Platform = $platform
         Channel = $channel
         FileUri = "https://vscode-update.azurewebsites.net/latest/$platform/$channel"
         Extension = $ext
     }
+
+    if ($pacMan) {
```
tylerl0706: You could probably just add this to the hashtable no?

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,476 @@ param(
...
     [Parameter()]
     [ValidateNotNull()]
     [string[]]$AdditionalExtensions = @(),
 
     [switch]$LaunchWhenDone,
 
-    [switch]$EnableContextMenus
+    [switch]$EnableContextMenus,
+
+    [switch]$WhatIf
```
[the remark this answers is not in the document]
rjmholt: Ah I wondered about the right way for this -- I didn't find anything that discussed `WhatIf` very well. I'll change it :slightly_smiling_face:

scripts/Install-VSCode.ps1
```diff
@@ -462,7 +460,7 @@ try {
 
     $installerPath = [System.IO.Path]::Combine($tmpdir, $installerName)
 
-    if ($onWindows) {
+    if ($PSVersionTable.PSVersion.Major -lt 6) {
```
tylerl0706: Not Core compatible huh?

scripts/Install-VSCode.ps1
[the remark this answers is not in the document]
rjmholt: Ok @rkeithhill, I've implemented this with ShouldProcess now. Let me know if I'm using it the right way :)

scripts/Install-VSCode.ps1
rjmholt: BitsTransfer wasn't loading for me

scripts/Install-VSCode.ps1
```diff
@@ -133,118 +133,461 @@ param(
...
-    [switch]$User,
-
     [Parameter()]
     [ValidateNotNull()]
     [string[]]$AdditionalExtensions = @(),
 
     [switch]$LaunchWhenDone,
 
-    [switch]$EnableContextMenus
+    [switch]$EnableContextMenus,
```
SeeminglyScience: Wouldn't this <kbd>,</kbd> cause a parse exception?

```suggestion
    [switch]$EnableContextMenus
```

rjmholt:
Ok, I've tested it on Windows and Ubuntu with WhatIf (and without -- it's part of an install script I wrote). Everything seems to be working :)

PR #1600, opened 2018-11-04
Add CodeAction support to show PSSA rule documentation

rkeithhill:
This corresponds to PSES PR:
https://github.com/PowerShell/PowerShellEditorServices/pull/789

Currently this points to the PSSA development branch.  Maybe that should
point to the master branch.  Could be a setting I suppose but that seems
like overkill.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/src/features/CodeActions.ts
+++ b/src/features/CodeActions.ts
@@ -6,13 +6,15 @@ import vscode = require("vscode");
 import { LanguageClient } from "vscode-languageclient";
 import Window = vscode.window;
 import { IFeature } from "../feature";
+import { ILogger } from "../logging";
 
 export class CodeActionsFeature implements IFeature {
-    private command: vscode.Disposable;
+    private applyEditsCommand: vscode.Disposable;
+    private showDocumentationCommand: vscode.Disposable;
     private languageClient: LanguageClient;
 
-    constructor() {
-        this.command = vscode.commands.registerCommand("PowerShell.ApplyCodeActionEdits", (edit: any) => {
+    constructor(private log: ILogger) {
+        this.applyEditsCommand = vscode.commands.registerCommand("PowerShell.ApplyCodeActionEdits", (edit: any) => {
             Window.activeTextEditor.edit((editBuilder) => {
                 editBuilder.replace(
                     new vscode.Range(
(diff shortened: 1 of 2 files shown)


rkeithhill:
There is more work to do here but I thought this would be a good start.  For instance, the tslint extension is adding menu entries to the context menu in the Problems view.  So you can right-click there and show documentation.  With the PR you have to use the Code Fix icon in the source code.  Still, it's a big improvement over not having it at all.  :-)

rkeithhill:
Here's an example of what this looks like:
image

bergmeister:
I think the development branch is OK to be used, especially since that is the branch against which PRs should be made (since I expect more people to do that now with this feature)

src/features/CodeActions.ts
```diff
@@ -23,13 +25,34 @@ export class CodeActionsFeature implements IFeature {
...
+    public showRuleDocumentation(ruleId: string) {
+        const pssaDocBaseURL = "https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation";
+
+        if (!ruleId) {
+            this.log.writeWarning("Cannot show documentation for code action, no ruleName was supplied.");
+            return;
+        }
+
+        if (ruleId.startsWith("PS")) {
+            ruleId = ruleId.substr(2);
```
rjmholt: Damn that's annoying

src/features/CodeActions.ts
```diff
@@ -23,13 +25,34 @@ export class CodeActionsFeature implements IFeature {
...
+        }
+
+        vscode.commands.executeCommand("vscode.open", vscode.Uri.parse(pssaDocBaseURL + `/${ruleId}.md`));
```
rjmholt: Nice!

src/features/CodeActions.ts
rkeithhill: Yeah, it's one of those things that on the [rule docs page](https://github.com/PowerShell/PSScriptAnalyzer/tree/development/RuleDocumentation) having every topic header start with `PS` makes it a bit harder to find a rule.  Well, I "believe" that was the motivation for the naming there.  :-)

rkeithhill:
Thanks.  I'll merge this right after I get the PSES version buttoned up.

Issue #1677, opened 2019-01-01
[Command Explorer] Configuration Options


wsmelton:
**Summary of the new feature**

Have the ability to configure a few options for the command explorer:

1. Filter modules, just exclusions

Anyone that uses AzureRM or the Az module has commonly found that majority of the "noise" in the command explorer for the ISE was taken up by that module. Having the ability to filter out the modules would allow users to more easily find modules they may want to work with in a particular workspace. (e.g. If I'm working on say something with dbatools then I don't need to see modules like AzureRm and Az in that list, because they may not pertain to the project.)

1. Group By

As it stands the commands are listed by letter. I would like to be able to have them listed by module right off, it offers a more focused discovery on whatever project I may need to use a module. I can find that module listed faster and discover the commands around it.

**Proposed technical implementation details (optional)**

I would think having the filter ability would be having a configuration like `powershell.cmdexplorer.filter` and the value provided be comma separated list (much like the font setting in Code supports. So something like:

```json
"powershell.cmdexplorer.filter": "'Az','AzureRm','Microsoft.PowerShell.Security'"
```

For the group by I would expect it only to support certain values, so it could start out by allowing group by feature for modules. If any additional ideas come up they just be setup as keywords that have to be provided based on what you want to group by. The default would be letters so it does not effect those that have gotten used to the current setup.

```json
"powershell.cmdexplorer.listby": "letter"
```

```json
"powershell.cmdexplorer.listby": "module"
```

```json
"powershell.cmdexplorer.listby": "verb"
```

SydneyhSmith:
@wsmelton thanks for this request, seems like a reasonable enhancement. The command explorer is a community driven feature, so I am marking it as "up for grabs", feel free to submit a pr 😄

corbob:
I've done up a proof of concept for an exclusion filter. The sorting one may take more thought, and might not be needed if we can get a search box.

I'll put together a PR and start to clean it up a bit.

andyleejordan:
Closing this during a backlog-triage pass. This enhancement request has had little or no activity since ~2019 and isn't on the current roadmap. That's not a judgment on the idea — if it's still something you'd like, please leave a comment or open a fresh issue with an up-to-date use case and we'll gladly reconsider. Thanks for the original suggestion!

<sub>🤖 Drafted by GitHub Copilot at @andyleejordan's request — pending maintainer review.</sub>

PR #1686, opened 2019-01-08
Add an Exclusion filter to the Command Explorer.

corbob:
## PR Summary
Add an Exclusion filter to the Command Explorer.

Suggested in  #1677

## PR Checklist

- [x] Summarized changes
- NA PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -482,6 +482,11 @@
           "default": true,
           "description": "Specifies the visibility of the Command Explorer in the PowerShell Side Bar."
         },
+        "powershell.sideBar.CommandExplorerExcludeFilter": {
+          "type":"string",
+          "default":"",
+          "description": ""
+        },
         "powershell.powerShellExePath": {
           "type": "string",
           "default": "",
--- a/package.json
+++ b/package.json
@@ -483,9 +483,9 @@
           "description": "Specifies the visibility of the Command Explorer in the PowerShell Side Bar."
         },
         "powershell.sideBar.CommandExplorerExcludeFilter": {
-          "type":"string",
+          "type":"array",
           "default":"",
-          "description": ""
+          "description": "Specify array of Modules to exclude from Command Explorer listing."
         },
         "powershell.powerShellExePath": {
           "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -482,6 +482,11 @@
           "default":true,
           "description": "Specifies the visibility of the Command Explorer in the PowerShell Side Bar."
         },
+        "powershell.sideBar.CommandExplorerExcludeFilter": {
+          "type":"string",
```
TylerLeonhardt: I know they said a comma separated list, but we can just use an array, right?

src/features/GetCommands.ts
```diff
@@ -52,6 +52,15 @@ export class GetCommandsFeature implements IFeature {
...
         }
         this.languageClient.sendRequest(GetCommandRequestType, "").then((result) => {
+            const SidebarConfig = vscode.workspace.getConfiguration("powershell.sideBar");
+            const excludeFilter: string = SidebarConfig.CommandExplorerExcludeFilter;
+            if (excludeFilter !== "") {
+                const filters = excludeFilter.split(",");
+                filters.forEach((filter) => {
+                    result = result.filter((command) =>
+                        !command.moduleName.includes(filter));
+                });
```
TylerLeonhardt: I think this forEach block can be shortened:

```ts
result.filter((command) => !filters.indexOf(command.moduleName));
```

package.json
corbob: IMHO the user experience of a comma separated string is better than for an array. If we put it in as an array, it creates a link to edit in the settings.js. I have been trying to replicate the `file.exclude` that is contributed by vscode itself. I appears to come from this code: https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/parts/files/electron-browser/files.contribution.ts#L192-L216 unfortunately when I replicate it, it still gives an edit in settings.js.

src/features/GetCommands.ts
corbob: Almost correct. Just had to test for indexOf returning -1 since that's what it returns when there's nothing there.

TylerLeonhardt:
@corbob do you think that `result` could ever be null here? That's my only concern.

corbob:
> @corbob do you think that `result` could ever be null here? That's my only concern.

I don't believe it should ever happen, but I'll do some modifications of PSES to send back a null and make sure we handle that properly on the off chance null comes back.

TylerLeonhardt:
cool cool. Looks good then!

corbob:
So I did a test, and result comes back as an empty array when get-command just does a write-host:

image

Tested on both PowerShell 5.1 and 6.2.

Issue #1667, opened 2018-12-20
Run `Invoke-Pester` on tests.ps1 file using right-click context menu on file tab and/or shortcut


bergmeister:
**Summary of the new feature**

As a user, I quite often want to run ALL Pester tests of a `tests.ps1` file but I have either multiple `Describe` blocks or don't want to scroll up and down to run tests. I then use the `Copy Path` context menu on the file tab just to run `Invoke-Pester` on that in the console.

**Proposed technical implementation details (optional)**

If the file ends in `tests.ps1` (case insensitive lookup), then a context menu on the file tab should be there that says `Run Pester Tests` and `Debug Pester Tests` and under the hood it just calls `Invoke-Pester -Path ` with the full path of the file in the integrated PowerShell console. A keyboard shortcut for this would be cool as well.

rkeithhill:
Fixed in the next update.

PR #1698, opened 2019-01-13
Add 'Run/Debug Pester tests' command and file tab menu

bergmeister:
## PR Summary

Implements #1667 by adding a command and file tab menu to run all Pester tests on a file. The file tab menu will only appear for `.tests.ps1` files (using a case insensitive comparison).
I tested this on Windows 1809 using vscode and the latest vscode-insiders when using Windows PowerShell 5.1. I also made sure that the code lens being used by PSES does not get broken, hence why new commands were added rather than trying to adapt for the different kind of uri in the `launchTests` function. Here is a zip of my locally built vsix for testing:  [PowerShell-insiders.zip](https://github.com/PowerShell/vscode-powershell/files/2765727/PowerShell-insiders.zip)

New file tab menu entries:
image
Commands:
image

*
## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -240,7 +240,7 @@
       ],
       "editor/title/context": [
         {
-          "when": "resourceFilename =~ /\\.[Tt]ests\\.ps1$/",
+          "when": "resourceFilename =~ /\\.tests\\.ps1$/i",
           "command": "PowerShell.RunPesterTests"
         }
       ],
--- a/package.json
+++ b/package.json
@@ -240,7 +240,7 @@
       ],
       "editor/title/context": [
         {
-          "when": "resourceExtname  == .ps1",
+          "when": "resourceFilename =~ /\\.[Tt]ests\\.ps1$/",
           "command": "PowerShell.RunPesterTests"
         }
       ],
--- a/package.json
+++ b/package.json
@@ -205,6 +205,11 @@
         "category": "PowerShell"
       },
       {
+        "command": "PowerShell.RunPesterTests",
+        "title": "Run Pester tests",
+        "category": "PowerShell"
+      },
+      {
         "command": "PowerShell.OpenExamplesFolder",
         "title": "Open Examples Folder",
         "category": "PowerShell"
@@ -231,6 +236,12 @@
           "when": "editorLangId == powershell",
           "command": "PowerShell.ShowHelp",
           "group": "2_powershell"
+        }
+      ],
+      "editor/title/context": [
+        {
+          "when": "resourceExtname  == .ps1",
+          "command": "PowerShell.RunPesterTests"
         }
       ],
       "view/title": [
--- a/package.json
+++ b/package.json
@@ -205,8 +205,13 @@
         "category": "PowerShell"
       },
       {
-        "command": "PowerShell.RunPesterTests",
+        "command": "PowerShell.RunPesterTestsFromFile",
         "title": "Run Pester tests",
+        "category": "PowerShell"
+      },
+      {
+        "command": "PowerShell.DebugPesterTestsFromFile",
+        "title": "Debug Pester tests",
         "category": "PowerShell"
       },
       {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -233,6 +238,12 @@
           "group": "2_powershell"
         }
       ],
+      "editor/title/context": [
+        {
+          "when": "resourceExtname  == .ps1",
```
rkeithhill: Does this work:
```
"when":  "resourceFilename =~ /\.[Tt]ests\.ps1$/"
```

bergmeister: Thanks for the tip with the regex, I did not know this was supported until now, I tried using `/\\.[Tt]ests\\.ps1$/` to escape the backslash in JSON but it did not work. Neither did `=~ /.*tests\\.ps1$/` work as proposed [here](https://github.com/Microsoft/vscode/issues/66461#issuecomment-453879835) by the vscode team. I suppose I have to play with this a bit more but thanks so far.

bergmeister: Thanks, I had to escape the backslash in JSON and then your suggestion worked: `"resourceFilename =~ /\\.[Tt]ests\\.ps1$/"`

rkeithhill:
You know, we could also add a `PowerShell.DebugPesterTests` or ``PowerShell.DebugPesterTestsFile` command.

bergmeister:
@rkeithhill Thanks for the tip with the completely case insensitive comparison for the `tests.ps1` file extension, I tested it and it works :)
Yes, I was thinking about adding a `Debug Pester tests` command/menu as well and when I was testing  it again, I realised that the current implementation is actually `Debug Pester tests` due to the defaults in [PesterTests.ts](https://github.com/PowerShell/vscode-powershell/blob/master/src/features/PesterTests.ts). I guess I'll have to pass in args but maybe the command should rather be split I guess? I'll therefore make it `WIP` again

rkeithhill:
Just as with your original post, I was thinking that we could have two functions `RunPesterTestsFile()` and `DebugPesterTestsFile()` that internally called `RunPesterTests` with the appropriate arguments.  Then we'd need to register those commands in the PesterTests.ps1 file.

bergmeister:
Yes, I was even thinking of just changing to make `RunPesterTests` just run the tests without the debugger and create a new function `DebugPesterTests` (internally both would lookup the file name if it is null and use the common `launchTests` function), this way we can avoid creating an additional layer of wrapping that is not necessary. The downside of that is that it requires a change in PSES which calls into `RunPesterTests` at the moment

rkeithhill:
Keep in mind that we are sharing this command with the code lens request message we get from PSES.

bergmeister:
@rkeithhill Yes, after playing a bit with it, it turned out to be easier to add new commands as you suggested instead of trying to correct the uri logic inside `launchTests`. I also added support for `Debug Pester test` and made sure that `Run Pester tests` does not attach the debugger any more. The PR is now ready for review again.

src/features/PesterTests.ts
```diff
@@ -15,6 +15,17 @@ export class PesterTestsFeature implements IFeature {
     private languageClient: LanguageClient;
 
     constructor(private sessionManager: SessionManager) {
+        this.command = vscode.commands.registerCommand(
+            "PowerShell.RunPesterTestsFromFile",
+            (uriString, runInDebugger, describeBlockName?) => {
+                this.launchTests(vscode.window.activeTextEditor.document.uri, false, describeBlockName);
```
rjmholt: Maybe worth documenting the `true` and `false` somehow? I assume it's "start debugger" or similar?

bergmeister: It's the `runInDebugger` parameter, which is defined on `launchTests`, isn't that sufficient? I simplified and minimised the lambda expression, maybe it is clearer now?

src/features/PesterTests.ts
```diff
@@ -17,13 +17,13 @@ export class PesterTestsFeature implements IFeature {
     constructor(private sessionManager: SessionManager) {
         this.command = vscode.commands.registerCommand(
             "PowerShell.RunPesterTestsFromFile",
-            (uriString, runInDebugger, describeBlockName?) => {
-                this.launchTests(vscode.window.activeTextEditor.document.uri, false, describeBlockName);
+            () => {
```
rkeithhill: Thanks!  I was just going to suggest this but see you've already done it.  :-)

PR #1714, opened 2019-01-22
Do not run pester tests when user cancels questions using the x button (bugfix)

bergmeister:
## PR Summary

@rkeithhill added a dialog in #1701 to ask the user whether he/she wants to run the whole Pester test file when there is an interpolated describe block string. However, when the user does not answer yes or no but cancels via the x button, the answer string will be `undefined` and the extension will run the test as if the user said yes. Therefore invert the logic to avoid running the test despite the user cancelling
image

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests -> Manually tested the 3 test scenarios


rkeithhill:
For my own edification, I would appreciate being added as a reviewer - lessening the chances that I make the same mistake again.  :-)

PR #1741, opened 2019-02-01
Update build to clear node modules directory

corbob:
## PR Summary

* Add removal of the node_modules directory on Clean task.
* Move condition for restore to the build prerequisite.
  * This allows the restoring of node modules even if the folder exists.

The main benefit of this is it keeps the build process in invoke-build when new node modules are added as they were recently.

## PR Checklist

- [x] Summarized changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -508,8 +508,8 @@
           "description": "Specifies the visibility of the Command Explorer in the PowerShell Side Bar."
         },
         "powershell.sideBar.CommandExplorerExcludeFilter": {
-          "type":"array",
-          "default":[],
+          "type": "array",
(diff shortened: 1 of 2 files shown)


corbob:
While I was doing more work with this, I noticed that node_modules wasn't being deleted when I cleaned. It turns out that vscode was holding the @types directory locked and not allowing it to be removed. A side effect of this was that you also couldn't restore modules. A happy medium I think is to do an `npm prune` to remove unused  modules.

I've also included a slight change to `package.json` because `npm install` keeps adjusting the formatting 🤷‍♂️

PR #1772, opened 2019-02-22
Escape "-Command" arguments to prevent Terminal API failure

foresthoffman:
Fixes: https://github.com/codercom/bugs/issues/270.

## PR Summary

On Ubuntu 16.04 containers, the Terminal API fails to start the pwsh executable with any "-Command" arguments. This appears to be due to unescaped single quotes within the argument list.

## PR Checklist

- [x] Summarized changes
- [x] ~PR has tests~ (NA)

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -69,9 +69,10 @@ export class PowerShellProcess {
                         powerShellArgs.push("-ExecutionPolicy", "Bypass");
                     }
 
+                    this.startArgs = this.startArgs.replace(/'/g, '"');
                     powerShellArgs.push(
                         "-Command",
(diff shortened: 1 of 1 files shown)


TylerLeonhardt:
Tested this out on Windows which uses PowerShell instead of bash as the driver and unfortunately, this change won't work there.

It simply prints out the full command which could be due to the use of `'..'` which just prints out whatever is inside.

I tried to leverage PowerShell's `-EncodedCommand` which is like `-Command` only it let's you pass a base64 encoded string instead of some mess of a command but I had trouble with it playing nice with the base64 encoded strings that the node side gave.

Ill have to investigate further.

TylerLeonhardt:
I've got a possible fix opened in #1774 so I'll close this in favor of that.

foresthoffman:
Sounds good!

PR #1818, opened 2019-03-29
Add additional changelog details

TylerLeonhardt:


rkeithhill:
Should we also update the release date in this changelog as well?

Issue #1849, opened 2019-04-06
How to prevent code formatting "findstr--->findstr.exe"


chuanjiao10:
vscode version = 1.33
powershell version = win10-17763.379

In my path there is grep.exe findstr.exe
My ps1 code will be formatted incorrectly.
ps | grep xxx ---->ps | grep.exe xxx
Findstr ---->findstr.exe
I need block this formatting rule.
Also, keep this code formatted  "? --->where-object"

I put PSScriptAnalyzerSettings.psd1 to：
both
 ```
c:\Users\userabcd\.vscode\extensions\ms-vscode.powershell-1.12.0\modules\PSScriptAnalyzer\1.18.0\Settings\PSScriptAnalyzerSettings.psd1
```
and
```
c:\Users\userabcd\.vscode\extensions\ms-vscode.powershell-1.12.0\modules\PSScriptAnalyzer\1.18.0\PSScriptAnalyzerSettings.psd1
```
Text content：
```
@{
    Severity=@('Error','Warning')
    ExcludeRules=@(' PSAvoidUsingCmdletAliases '  =  @ {
             ' Whitelist '  =  @（' cd '）
    })
}
```
rerurn format crlf,file format utf8+bom.
but ,dont work.
**Can't stop this formatting  "? --->where-object"**

Q1：Where is the default path for PSScriptAnalyzerSettings.psd1?
i no set powershell.scriptAnalysis.settingsPath
Q2：How to write exclusion rules?

Q3：Which exclusion rule meets my needs?

**Comment by mkht** [comment]
@chuanjiao10 
I am able to reproduce it.
This issue is related #1842 (and there is some duplication)

My testing code before formatting
```PowerShell
gci -Path $TempFile | ? { test-path -Path $_ -PathType Leaf } | % { get-content $_ } | findstr 'aaa'
```

This code is formatted by vscode-powershell@1.12.0.
```PowerShell
Get-ChildItem -Path $TempFile | Where-Object { Test-Path -Path $_ -PathType Leaf } | ForEach-Object { Get-Content $_ } | findstr.exe 'aaa'
```

<details><summary>Testing environment</summary>
<p>

### Operating system
```
OsName               : Microsoft Windows 10 Pro
OsOperatingSystemSKU : 48
OsArchitecture       : 64-bit
WindowsVersion       : 1809
WindowsBuildLabEx    : 17763.1.amd64fre.rs5_release.180914-1434
OsLanguage           : en-US
[...]

Issue #1842, opened 2019-04-03
Cmdlet aliases are expanded automatically when formatting document


mkht:
#### VSCode version
```
1.32.3 a3db5be9b5c6ba46bb7555ec5d60178ecc2eaae4 x64
```

#### VSCode extensions
```
ms-vscode.powershell@1.12.0
```
No other extensions is installed.  
The vscode settings have not changed anything.

#### PSES version
```
1.12.0.0
```

#### PowerShell version:
```PowerShell
Name                           Value
----                           -----
PSVersion                      5.1.17763.1
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17763.1
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

### Issue Description
All aliases in my PS codes are expanded unexpectedly when executing "Format Document".

I tried formatting the code below.
```PowerShell
1..10|%{$_*2}
```

#### Expected result
I expect this code will format like this.
```PowerShell
1..10 | % { $_ * 2 }
```

#### Actual result
But vscode-powershell@1.12.0.0 formatted.
```PowerShell
1..10 | ForEach-Object { $_ * 2 }
```

I know that formatting is provided by the PSScriptAnalyzer module.  
However, I think this issue is specific to vscode-powershell.  
I execute these commands in the Windows PowerShell terminal. (Not in the integrated terminal)

```PowerShell

PS C:\> Install-Module PSScriptAnalyzer -Scope CurrentUser
PS C:\> Get-Module PSScriptAnalyzer -ListAvailable

    Directory: C:\Users\Administrator\Documents\WindowsPowerShell\Modules

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.18.0     PSScriptAnalyzer                    {Get-ScriptAnalyzerRule, Invoke-ScriptAnalyzer, Invoke-Formatter}

PS C:\> Invoke-Formatter '1..10|%{$_*2}'
1..10 | % { $_ * 2 }
```

In this result, The original formatter of the PSScriptAnalyzer@1.18.0 is not expand aliases automatically.  

Is this bug?  Does anyone have an idea of how to work around it?

SydneyhSmith:
[...]

PR #1852, opened 2019-04-07
Turn powershell.codeformatting.useCorrectCasing setting off by default until PSSA issues are fixed

bergmeister:
## PR Summary

Due to the following formattig PSSA issues it's probably better to not enable the setting by default until a patch of PSSA is available (we could also do this only for version 1 of the extension but I am currently targeting master, it is up to you and the preferred timeline of next extension or PSSA release)
- `Alias "?" is changed to "%" by Invoke-Formatter`: https://github.com/PowerShell/PSScriptAnalyzer/issues/1209
- `Cmdlet aliases are expanded automatically when formatting document`: https://github.com/PowerShell/vscode-powershell/issues/1842
- Applications such as e.g. `git` get corrected to `git.exe` (which would make a script not cross-platform any more): https://github.com/PowerShell/vscode-powershell/issues/1849

Note that this requires the following PSES PR because the vscode extension setting was not actually used by PSES: https://github.com/PowerShell/PowerShellEditorServices/pull/910

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -697,7 +697,7 @@
         },
         "powershell.codeFormatting.useCorrectCasing": {
           "type": "boolean",
-          "default": true,
+          "default": false,
           "description": "Use correct casing for cmdlets."
         },
         "powershell.integratedConsole.showOnStartup": {
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -150,7 +150,7 @@ export function load(): ISettings {
         WhitespaceAroundPipe: true,
         ignoreOneLineBlock: true,
         alignPropertyValuePairs: true,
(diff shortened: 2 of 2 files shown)


rkeithhill:
I wonder if in a separate PR we should correct the casing on these two settings?
```
        WhitespaceInsideBrace: true,
        WhitespaceAroundPipe: true, 
```
If we are going to correct this, probably better to do it sooner before more folks customize these settings.

bergmeister:
Yes, since the settings are new, the likelihood that people already use them is low, the release blog post mentions them though.

PR #2035, opened 2019-06-18
#1019: Get format settings from document editor instead of global.

tillig:
## PR Summary

Change the document formatter so indent settings come from the document-specific editor rather than global configuration. This allows language-specific overrides and other things like EditorConfig to modify the settings and the document formatter will obey.

## PR Checklist

- [X] Summarized changes
- [NA] PR has tests


corbob:
> This LGTM, thanks for your contribution.
> 
> I wonder if there are other locations in the code where we are doing this naive `get` that should also be changed to an `editor.options.*`

Probably the Command Explorer filter and such...

PR #2062, opened 2019-07-02
Remove redundant snippets

travisclagrone:
Remove the following snippets (listed by name, not prefix):

- `Cmdlet`

The *removed-redundant-snippet* to *kept-redundantee-snippet* mappings are as follows (listed by name, not prefix):

- `Cmdlet` : `Function-Advanced`  
  The decision to remove `Cmdlet` instead of `Function-Advanced` is because a function defined in PowerShell with the `CmdletBinding` attribute is technically an advanced function rather than a cmdlet, which would be defined in C# with the `Cmdlet` attribute.  

  However, the `cmdlet` prefix has also been added to the `Function-Advanced` snippet (whose other prefix is `function-advanced`), because `cmdlet` is an ISE snippet prefix for an advanced function and so should be preserved for compatibility.

### DIFF
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -513,7 +513,10 @@
         "description": "Enum definition snippet"
     },
     "Function-Advanced": {
-        "prefix": "function-advanced",
+        "prefix": [
+            "function-advanced",
+            "cmdlet"
+        ],
         "body": [
             "function ${1:Verb-Noun} {",
             "\t[CmdletBinding()]",
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -539,7 +539,7 @@
         ],
         "description": "Script advanced function definition snippet"
     },
-    "Cmdlet-Comment-Help": {
+    "Comment-Help": {
         "prefix": "comment-help",
         "body": [
             "<#",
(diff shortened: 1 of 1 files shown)


travis-c-lagrone:
Aha. In that case, either a more descriptive name or a source-code comment would be better.

SydneyhSmith:
@travis-c-lagrone this is in two snippets because of an [earlier limitation imposed by vscode](https://github.com/microsoft/vscode/issues/42838) , now you can update the prefix to an array to get this functionality ` prefix:["cmdlet", "Function-Advanced"]`

travis-c-lagrone:
> @travis-c-lagrone this is in two snippets because of an [earlier limitation imposed by vscode](https://github.com/microsoft/vscode/issues/42838) , now you can update the prefix to an array to get this functionality ` prefix:["cmdlet", "Function-Advanced"]`

I did not know that! I might see about adding that to the VS Code documentation as well when I get the time. (it does not appear to be documented at the moment)

travis-c-lagrone:
FYI, I plan on submitting another PR soon to have the folder `.vscode` settings associate the "JSON with Comments" (`jsonc`) language for the `snippets/PowerShell.json` file. This is already how VS Code interprets snippet files anyway, but it will explicitly communicate commentability within the `vscode-powershell` project now too. This is important in order to be able to call out (at development time) overloaded snippets with ISE compatibility prefixes, _inter alia_.

PR #2070, opened 2019-07-06
Add snippet 'foreach-indexed'

travis-c-lagrone:
## PR Summary

Adds a snippet named `foreach-indexed`.

The `foreach-indexed` snippet enumerates a collection using a `foreach` loop, but also tracks the index of the current item in the process. Regarding *motivation*, the use cases of the `foreach-indexed` construct (which enumerates element-to-index) are somewhat the opposite of the `for` construct (which enumerates index-to-element): synthetically indexing an unordered collection (e.g. a `HashSet`), relatively indexing a live enumerator (vs. an absolutely-indexed enumerable), etc. Regarding *justification*, the logic sequencing required by the `foreach-indexed` construct to defend against naive uses of the `continue` keyword in the middle of the loop body (i.e. continuing without also incrementing the index) is both unconventional as well as more verbose than the conventional pattern (i.e. of incrementing the index at the end of the loop body).

## PR Checklist

- [x] Summarized changes
- [ ] `NA` ~~PR has tests~~

### DIFF
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -766,6 +766,17 @@
         ],
         "description": "foreach loop snippet"
     },
+    "foreach-indexed": {
+        "prefix": "foreach-indexed",
+        "body": [
+            "\\$private:_index = 0",
+            "foreach ($${1:item} in $${2:collection}) {",
+            "\t\\$index = (\\$_index++)",  // Instead of incrementing the index at the end of the loop body, incrementing it at the beginning (and therefore also assigning a proxy variable) defends against naive invocations of the `continue` statement in the middle of the loop body.
+            "\t${0:$TM_SELECTED_TEXT}",
+            "}"
+        ],
+        "description": "indexed foreach loop snippet"
+    },
     "function": {
         "prefix": "function",
         "body": [


rjmholt:
Thanks for your contribution @travis-c-lagrone.

Although this is generally applicable, I think this one is more advanced and specific than other snippets we enable by default and would be better in the [community snippets doc](https://github.com/PowerShell/vscode-powershell/blob/master/docs/community_snippets.md).

travisclagrone:
@rjmholt Would you plesae clarify what the criteria (even if open-ended) is for whether a snippet is appropriate to be included in the extension? I ask because while I do not understand how this proposed snippet is either more "advanced" or "specific" than many of the other snippets in the extension (e.g. "IfShouldProcess", "CalculatedProperty", "IArgumentCompleter Class", etc.), although I could see how it might be rejected on the basis of less frequent use.

rjmholt:
It's certainly case-by-case, but here the feeling was that an indexed foreach loop is an uncommon use-case given a standard for-loop and the usage of the private `$_index` as well as the `$index` variable made for a complex solution despite being nifty.

Because the JSON snippets represent part of the completions the extension offers, we really want to make sure that those snippets represent PowerShellisms most users would use; so essentially standard structures (like a loop or an advanced function or a switch) that you would find yourself using on a weekly basis but for which there's no completion.

travisclagrone:
Ok, I can grok the idea of "PowerShellisms". I'll modify this PR to add the `foreach-indexed` snippet to the Community Snippets instead.

Also, before I make a PR for it, how would you see a snippet like the following applying to the idea of regularly used PowerShellisms for which there's no completion (much less dedicated command)?

```json
"Test-Any loop": {
    "prefix": "any",
    "body": [
        "\\$any = \\$false",
        "foreach (\\$${1:object} in \\$${2:objects}) {",
        "\tif (${3:\\$$1}) {",
        "\t\t\\$any = \\$true",
        "\t\tbreak",
        "\t}",
        "}",
    ],
},
"Test-All loop": {
    "prefix": "all",
    "body": [
        "\\$all = \\$true",
        "foreach (\\$${1:object} in \\$${2:objects}) {",
        "\tif (-not ${3:\\$$1}) {",
        "\t\t\\$all = \\$false",
        "\t\tbreak",
        "\t}",
        "}",
    ],
},
```

B4Art:
Reading this thread I come to the conclusion that this is more a Compare issue than a Index issue.
As your example suggests:
2 Object lists.
Where any two elements are equal,
 OR
Where all elements are available in both lists.

godindustry:
foreach-indexed

andschwa:
@JustinGrote a snippet you might want to take a look at!

Issue #1611, opened 2018-11-13
Simplify generated launch.json 


isidorn:
Hey,

VSCode dev here. This milestone I am looking into simplifing generated `launch.json` for various extensions Microsoft/vscode#62851

The `launch.json` that Powershell generates is attached at the end. This is a bit too complex for the avarage user and I suggest to simplify it the following way:

- [ ] You are already using a `DebugConifugarionProvider` for resolving configuration. I suggest to aslo use it for providing debug conifiguration as it should make the following step possible
- [ ] `DebugConfigurationProvider` should use the `quickPick` to ask the user if he would like to launch or to attach. If he chooses launch I would ask a follow up question if he would like to launch a current file, Current File in Temporary Console, Current File w/Args Prompt or just an interactive powershell session. 
- [ ] Remove default fields which are not necessery like `args`, `runspaceId`
- [ ] You are already contributing `configuraitonSnippets` which is great. However these snippets also need to be simplified. I suggest to remove all attributes that have the default value. Example: `stopAtEntry`, `cwd`, `args`

If you agree with the suggestions I am making here I am also here to help with any potential questions you might have. The changes should not require a lot of work but will simplify the flow a lot imho. It should be much less complex and not too much like a wizard experience

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "PowerShell",
            "request": "launch",
            "name": "PowerShell Launch Current File",
            "script": "${file}",
            "args": [],
            "cwd": "${file}"
        },
        {
            "type": "PowerShell",
            "request": "launch",
            "name": "PowerShell Launch Current File in Temporary Console",
            "script": "${file}",
            "args": [],
            "cwd": "${file}",
            "createTemporaryIntegratedConsole": true
        },
        {
            "type": "PowerShell",
            "request": "launch",
            "name": "PowerShell Launch Current File w/Args Prompt",
            "script": "${file}",
            "args": [
                "${command:SpecifyScriptArgs}"
            ],
            "cwd": "${file}"
        },
        {
            "type": "PowerShell",
            "request": "attach",
            "name": "PowerShell Attach to Host Process",
            "processId": "${command:PickPSHostProcess}",
            "runspaceId": 1
        },
        {
            "type": "PowerShell",
            "request": "launch",
[...]

PR #2084, opened 2019-07-14
Implement #1611 - provide dynamic debug config

rkeithhill:
## PR Summary

This an overdue, first-pass attempt at implementing dynamic debug configurations as suggested by @isidorn in #1611.  I'm completely open to changing the text that appears in the various prompts.  I'm also open to the idea that "Attach to PowerShell Host Process" is rare enough that we could eliminate one level of prompting and only allow dynamic debug config for "Launch".  Or maybe we make the second level quick pick a multi-select option and then create the debug configs the user selects?  Then we could remove the first quick pick prompt.

BTW this functionality **only** comes into play when you don't have a launch.json file. The first time you configure launch.json is where you will see this.  Once launch.json exists, you are always choosing from the list of debug config snippets we provide in our package.json file.  One benefit of this approach is that when a user goes through this process now, they get just one debug config added to their new launch.json as opposed to everything we used to define in our package.json's initialDebugConfigurations - which was a lot of stuff.  

@isidorn I tried all the other debuggers on my system in this mode and none of them prompt (node, chrome, C++ gdb/Windows, .NET Core).  Are you sure this is OK for PowerShell to do?  Also, would you mind reviewing this PR?

Here is what this process looks like:

DebugLaunchConfig1

DebugLaunchConfig2

DebugLaunchConfig3

DebugAttachConfig1

Finally, I removed the `temp integrated console debug` config snippet.  That will still continue to work but I think we should remove this for now as its a bug factory.  Also, looking at this list of snippets, I wonder if we could cull it further:

image

I'm fuzzy on what the difference is between `PowerShell: Interactive Session` and `PowerShell: Attach Interactive Session`.  Also, not sure the `w/Args` is worth having as a snippet. I did add to the docs on `args` that you could specify `${command:SpecifyScriptArgs}` to be prompted for args.  Thoughts?

Also, one last thing.  We had the breakpoints configuration in a part of package.json that isn't valid anymore / deprecated.  So, I moved it to where the VSCode debug docs say it should go.  That said, I'm not sure when that switchover happened.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -2,7 +2,6 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import path = require("path");
 import vscode = require("vscode");
 import { CancellationToken, DebugConfiguration, DebugConfigurationProvider,
     ExtensionContext, WorkspaceFolder } from "vscode";
@@ -90,7 +89,9 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
                     cwd: "${file}",
                 },
             ];
-        } else if (launchSelection.id === launchScriptId) {
+        }
+
+        if (launchSelection.id === launchScriptId) {
             return [
                 {
                     name: "PowerShell: Launch Script",
@@ -100,7 +101,9 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
                     cwd: "${workspaceFolder}",
                 },
             ];
-        } else if (launchSelection.id === interactiveSessionId) {
+        }
+
+        if (launchSelection.id === interactiveSessionId) {
             return [
                 {
                     name: "PowerShell: Interactive Session",
@@ -111,7 +114,7 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
             ];
         }
 
-        // Return the "Attach to PowerShell Host Process" debug configuration
+        // Last remaining possibility is attach to host process
         return [
             {
                 name: "PowerShell: Attach to PowerShell Host Process",
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -47,82 +47,74 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
         folder: WorkspaceFolder | undefined,
         token?: CancellationToken): Promise<DebugConfiguration[]> {
 
-        const launchAttachItems = [
-            { label: "Launch",
-              description: "Launch the debugger with a specified script or for the interactive session" },
-            { label: "Attach",
-              description: "Attach the debugger to a running PowerShell Host Process" },
+        const launchCurrentFileLabel  = "Launch Current File";
+        const launchScriptLabel       = "Launch Script";
+        const interactiveSessionLabel = "Interactive Session";
+
+        const debugConfigPickItems = [
+            {
+                label: launchCurrentFileLabel,
+                description: "Launch and debug the file in the currently active editor window",
+            },
+            {
+                label: launchScriptLabel,
+                description: "Launch and debug the specified file or command",
+            },
+            {
+                label: interactiveSessionLabel,
+                description: "Debug commands executed from the Integrated Console",
+            },
+            {
+                label: "Attach",
+                description: "Attach the debugger to a running PowerShell Host Process",
+            },
         ];
 
-        const debugTypeSelection =
+        const launchSelection =
             await vscode.window.showQuickPick(
-                launchAttachItems,
-                { placeHolder: "Would you like to launch or attach the PowerShell debugger?" });
+                debugConfigPickItems,
+                { placeHolder: "Select a PowerShell debug configuration" });
 
-        let debugConfiguration = [];
-
-        if (debugTypeSelection.label === "Launch") {
-            const launchCurrentFileLabel = "Launch Current File";
-            const launchScriptLabel = "Launch Script";
-            const interactiveSessionLabel = "Interactive Session";
-
-            const launchItems = [
-                { label: launchCurrentFileLabel,
-                  description: "Debugs whichever script is in the active editor window" },
-                { label: launchScriptLabel,
-                  description: "Debugs the specified script" },
-                { label: interactiveSessionLabel,
-                  description: "Debugs scripts or modules executed from the Integrated Console" },
-            ];
-
-            const launchSelection =
-                await vscode.window.showQuickPick(
-                    launchItems,
-                    { placeHolder: "Select a launch option" });
-
-            if (launchSelection.label === launchCurrentFileLabel) {
-                debugConfiguration = [
-                    {
-                        name: "PowerShell: Launch Current File",
-                        type: "PowerShell",
-                        request: "launch",
-                        script: "${file}",
-                        cwd: "${file}",
-                    },
-                ];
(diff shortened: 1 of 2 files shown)


src/features/DebugSession.ts
```diff
@@ -42,6 +43,88 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
...
+        ];
+
+        const debugTypeSelection =
+            await vscode.window.showQuickPick(
+                launchAttachItems,
+                { placeHolder: "Would you like to launch or attach the PowerShell debugger?" });
+
+        let debugConfiguration = [];
+
+        if (debugTypeSelection.label === "Launch") {
```
rjmholt: This logic here I would write as:

```typescript
const powerShellType = 'PowerShell';

if (debugTypeSelection.label !== "Launch") {
    return [
        {
            name: 'PowerShell: Attach to PowerShell Host Process',
            type: powerShellType,
            request: 'attach',
            runspaceId: 1
        }
    ]
}

const launchCurrentFileLabel = 'Launch Current File';
const launchScriptLabel = 'Launch Script';
const interactiveSessionLabel = 'Interactive Session';

const launchItems = [
    {
        label: launchCurrentFileLabel,
        description: 'Debugs the script currently in the active editor window'
    },
    {
        label: launchScriptLabel,
        description: 'Debugs the specified script'
    },
    {
        label: interactiveSessionLabel,
        description: 'Debugs scripts or modules executed from the Integrated Console'
    }
];

const launchSelection = await vscode.window.showQuickPick(
    launchItems,
    { placeHolder: 'Select a debugger launch option'});

if (launchSelection.label === launchCurrentFileLabel) {
    return [
        {
            name: 'PowerShell: Launch Current File',
            type: powerShellType,
            request: 'launch',
            script: '${file}',
            cwd: '${file}'
        }
    ]
}

if (launchSelection.label === launchScriptLabel) {
    return [
        {
            name: 'Powershell: Launch Script',
            type: powerShellType,
            request: 'launch',
            script: 'Enter path or command to execute, e.g.: ${workspaceFolder}/src/foo.ps1 or Invoke-Pester'
        }
    ]
}

// Last remaining possibility is interactive launch
return [
    {
        name: 'PowerShell: Interactive Session',
        type: powerShellType,
        request: 'launch',
        cwd: ''
    }
];
```

rkeithhill: If the impl remains  the way it is then yeah, it should be broken up into mutiple methods like you show above.  At this point, I'm more interested in feedback on the experience.  I think two levels of prompts is probably one too many.  Yesterday, I was thinking that perhaps the multi-selection approach would be better.  I'm really interested in getting @isidorn feedback because none of the other debuggers except for well, Python. It does prompt with this.  

image

BTW with that Python debug prompt above, if I press <kbd>escape</kbd> to cancel the operation, I get all the Python configs dumped into my launch.json.   That seems like a bug.

rkeithhill:
This last commit reduces the number of prompts to one:

image

Still open to wording changes.  I updated the wording of the `Launch Script` debug configuration snippet to show how you can invoke arbitrary commands like `Invoke-Pester` e.g.:
```json
        {
            "name": "PowerShell: Launch Script",
            "type": "PowerShell",
            "request": "launch",
            "script": "enter path or command to execute e.g.: ${workspaceFolder}/src/foo.ps1 or Invoke-Pester",
            "cwd": "${workspaceFolder}"
        }
```
Given the above **and** the fact that you can use code lens to debug Pester tests (the most common way to debug Pester tests IMO), I've removed the `Launch Pester Tests` debug config snippet leaving these:

image

I'd even like to get rid of the `w/args` snippet.  When a user gets intellisense on the `args` parameter they see how they can make their debug configuration prompt for args:

image

Plus this prompt functionality applies to both `Launch Current File` as well as `Launch Script` so it's a bit weird only having this snippet for one and not the other.  Thoughts?

src/features/DebugSession.ts
```diff
@@ -42,6 +43,80 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
...
+                label: interactiveSessionLabel,
+                description: "Debug commands executed from the Integrated Console",
+            },
+            {
+                label: "Attach",
+                description: "Attach the debugger to a running PowerShell Host Process",
+            },
+        ];
+
+        const launchSelection =
```
TylerLeonhardt: Still can't get the index of the selection, huh?

rkeithhill: Nah, I think you get back the selected item object.  I could add an index/id to the item objects and then test against that.  Worth it?

rkeithhill:
That last commit whittles the debug configuration snippets to just these:

image

src/features/DebugSession.ts
TylerLeonhardt: Can you try it and let us know? I'm also curious for other situations as well.

rkeithhill: I looked at the API, it returns `Thenable<T | undefined>` where `T` is the type of the items array provided as input:

image

TylerLeonhardt: Ah ok so in theory if our type had an index property, we could take advantage of it

rkeithhill:
@rjmholt Are you OK with the latest changes?

src/features/DebugSession.ts
```diff
@@ -2,9 +2,10 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
+import path = require("path");
```
rjmholt:
```suggestion
import * as path from "path";
```

src/features/DebugSession.ts
```diff
@@ -42,6 +43,85 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
...
+            return [
+                {
+                    name: "PowerShell: Launch Current File",
+                    type: "PowerShell",
+                    request: "launch",
+                    script: "${file}",
+                    cwd: "${file}",
+                },
+            ];
+        } else if (launchSelection.id === launchScriptId) {
```
rjmholt:
```suggestion
        }
        
        if (launchSelection.id === launchScriptId) {
```

src/features/DebugSession.ts
```diff
@@ -42,6 +43,85 @@ export class DebugSessionFeature implements IFeature, DebugConfigurationProvider
...
+            return [
+                {
+                    name: "PowerShell: Launch Script",
+                    type: "PowerShell",
+                    request: "launch",
+                    script: "enter path or command to execute e.g.: ${workspaceFolder}/src/foo.ps1 or Invoke-Pester",
+                    cwd: "${workspaceFolder}",
+                },
+            ];
+        } else if (launchSelection.id === interactiveSessionId) {
```
rjmholt:
```suggestion
        }
        
        if (launchSelection.id === interactiveSessionId) {
```

src/features/DebugSession.ts
rkeithhill: Removed this since we weren't using `path` anywhere in the file.

isidorn:
@rkeithhill I appologise for not reviewing - I was out on vacation for 3 weeks.
Though I see this is merged and it looks good.
If you roll it out in insiders you should get immediate user feedback to potentially further polish this.

Thanks!

rkeithhill:
@isidorn No worries. I'm pretty happy with where we ended up.  And we can always course-correct based on user feedback.

PR #2094, opened 2019-07-18
Set Default PowerShell Version on Windows to Latest Available

SydneyhSmith:
## PR Summary

Updated the default PowerShell path on Windows to check for the latest version (previous behavior defaulted to Windows PowerShell). Also removed an SSL check that is no longer necessary. 

## PR Checklist

- [X] Summarized changes
- [x] PR has tests

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -73,7 +73,7 @@ export function getDefaultPowerShellPath(
     if (platformDetails.operatingSystem === OperatingSystem.Windows) {
         if (use32Bit) {
             psCoreInstallPath =
-            (platformDetails.isProcess64Bit ? process.env['ProgramFiles(x86)'] : process.env.ProgramFiles)
+            (platformDetails.isProcess64Bit ? process.env["ProgramFiles(x86)"] : process.env.ProgramFiles)
             + "\\PowerShell";
         } else {
 
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -98,7 +98,7 @@ export function getDefaultPowerShellPath(
                 return powerShellExePath = psCorePaths[0].exePath;
             }
         }
-        
+
         if (use32Bit) {
                 powerShellExePath =
                     platformDetails.isOS64Bit && platformDetails.isProcess64Bit
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -99,6 +99,7 @@ export function getDefaultPowerShellPath(
             }
         }
 
+        // No PowerShell 6+ detected so use Windows PowerShell.
         if (use32Bit) {
                 return platformDetails.isOS64Bit && platformDetails.isProcess64Bit
                         ? SysWow64PowerShellPath
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -109,7 +109,7 @@ export function getDefaultPowerShellPath(
                     !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
                         ? System32PowerShellPath
                         : SysnativePowerShellPath;
-            }
+        }
     } else if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
         // Always default to the stable version of PowerShell (if installed) but handle case of only Preview installed
         powerShellExePath = macOSExePath;
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -108,7 +108,7 @@ export function getDefaultPowerShellPath(
         return !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
                  ? System32PowerShellPath
                  : SysnativePowerShellPath;
-    }   if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
+    } if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
         // Always default to the stable version of PowerShell (if installed) but handle case of only Preview installed
         powerShellExePath = macOSExePath;
         if (!fs.existsSync(macOSExePath) && fs.existsSync(macOSPreviewExePath)) {
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -76,10 +76,10 @@ export function getDefaultPowerShellPath(
                 (platformDetails.isProcess64Bit ? process.env["ProgramFiles(x86)"] : process.env.ProgramFiles)
             + "\\PowerShell";
         } else {
-
             psCoreInstallPath =
                 (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";
         }
+        
         if (fs.existsSync(psCoreInstallPath)) {
             const arch = platformDetails.isProcess64Bit ? "(x64)" : "(x86)";
             const psCorePaths =
@@ -98,6 +98,7 @@ export function getDefaultPowerShellPath(
                 return powerShellExePath = psCorePaths[0].exePath;
             }
         }
+        
         if (use32Bit) {
                 powerShellExePath =
                     platformDetails.isOS64Bit && platformDetails.isProcess64Bit
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -229,7 +229,7 @@ export function getAvailablePowerShellExes(
         exePaths.forEach((exePath) => {
             if (fs.existsSync(exePath)) {
                 paths.push({
-                    versionName: "PowerShell " + (/-preview/.test(exePath) ? " Preview" : ""),
+                    versionName: "PowerShell" + (/-preview/.test(exePath) ? " Preview" : ""),
                     exePath,
                 });
             }
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -79,7 +79,7 @@ export function getDefaultPowerShellPath(
             psCoreInstallPath =
                 (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";
         }
-        
+
         if (fs.existsSync(psCoreInstallPath)) {
             const arch = platformDetails.isProcess64Bit ? "(x64)" : "(x86)";
             const psCorePaths =
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -69,17 +69,39 @@ export function getDefaultPowerShellPath(
 
     // Find the path to powershell.exe based on the current platform
     // and the user's desire to run the x86 version of PowerShell
-    if (platformDetails.operatingSystem === OperatingSystem.Windows) {
-        if (use32Bit) {
-            powerShellExePath =
-                platformDetails.isOS64Bit && platformDetails.isProcess64Bit
-                    ? SysWow64PowerShellPath
-                    : System32PowerShellPath;
+        if (platformDetails.operatingSystem === OperatingSystem.Windows) {
+        const psCoreInstallPath =
+        (!platformDetails.isProcess64Bit ? process.env.ProgramW6432 : process.env.ProgramFiles) + "\\PowerShell";
+        if (fs.existsSync(psCoreInstallPath)) {
+            const arch = platformDetails.isProcess64Bit ? "(x64)" : "(x86)";
+            const psCorePaths =
+                fs.readdirSync(psCoreInstallPath)
+                .map((item) => path.join(psCoreInstallPath, item))
+                .filter((item) => {
+                    const exePath = path.join(item, "pwsh.exe");
+                    return fs.lstatSync(item).isDirectory() && fs.existsSync(exePath);
+                })
+                .map((item) => ({
+                    versionName: `PowerShell Core ${path.parse(item).base} ${arch}`,
+                    exePath: path.join(item, "pwsh.exe"),
+                }));
+
+            if (psCorePaths) {
+                powerShellExePath = psCorePaths[0].exePath;
+            }
+
         } else {
-            powerShellExePath =
-                !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
-                    ? System32PowerShellPath
-                    : SysnativePowerShellPath;
+            if (use32Bit) {
+                powerShellExePath =
+                    platformDetails.isOS64Bit && platformDetails.isProcess64Bit
+                        ? SysWow64PowerShellPath
+                        : System32PowerShellPath;
+            } else {
+                powerShellExePath =
+                    !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
+                        ? System32PowerShellPath
+                        : SysnativePowerShellPath;
+            }
         }
     } else if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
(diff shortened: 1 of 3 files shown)


SydneyhSmith:
Looks like the error might be a result of the intended design-- the default PowerShell path is no longer Windows (as previously expected)
image

src/platform.ts
```diff
@@ -70,16 +70,38 @@ export function getDefaultPowerShellPath(
     // Find the path to powershell.exe based on the current platform
     // and the user's desire to run the x86 version of PowerShell
     if (platformDetails.operatingSystem === OperatingSystem.Windows) {
-        if (use32Bit) {
-            powerShellExePath =
-                platformDetails.isOS64Bit && platformDetails.isProcess64Bit
-                    ? SysWow64PowerShellPath
-                    : System32PowerShellPath;
+        const psCoreInstallPath =
+        (!platformDetails.isProcess64Bit ? process.env.ProgramW6432 : process.env.ProgramFiles) + "\\PowerShell";
```
corbob: Does `process.env.ProgramW6432` resolve properly on 32 bit Windows?

TylerLeonhardt: Not sure what you mean @corbob

src/platform.ts
```diff
@@ -70,16 +70,38 @@ export function getDefaultPowerShellPath(
...
+        if (fs.existsSync(psCoreInstallPath)) {
```
TylerLeonhardt: _Just_ in case someone went and created a `Program Files\PowerShell` path… we should refactor this if to say, if I’ve found a PowerShell Core, then return it, otherwise, use Windows PowerShell.

What I mean by this is:

```ts
if (fs.existsSync(psCoreInstallPath)) {

    // still do what you’re doing…

    if (psCorePaths) {
        return psCorePaths[0].exePath;
    }
}

// no ELSE, just go right to Windows PowerShell logic
if (use32Bit) {
    // …
```

src/platform.ts
TylerLeonhardt: negative ternary ops can be dangerous because the `!` is often overlooked. If you can flip this to be:
```suggestion
        (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";
```
it’s a bit more readable

corbob: Yes, it took me a good 5 minutes to wrap my head around what this was doing.

corbob: What I mean is that on 32 bit Windows, the environment variable `ProgramW6432` doesn't exist... So if someone is on 32 bit Windows with PowerShell Core installed (my first question is why?!?!?!?!) runs, it won't default to Core.

That being said: if I were to guess, I would say the number of people using vs code on 32 bit windows with PowerShell Core is probably really close to 0...

TylerLeonhardt: That's odd. What _is_ the 32bit program files location/is there an env var that gives us that?

corbob: on 32bit Windows it's just `ProgramFiles` and it resolves to C:\Program Files (usually)

I think if you just check for `ProgramFiles`, then the PowerShell will match the bitness of the process if it can find it...

TylerLeonhardt: Ah ok

So what we want to do is:

If 32 bit, check ProgramFiles
If 64 bit, check both ProgramFiles and ProgramW6432

?

SydneyhSmith: Something like this? 
```
    if (platformDetails.operatingSystem === OperatingSystem.Windows) {
        if (use32Bit){
            psCoreInstallPath = process.env.ProgramFiles + "\\PowerShell"
        }
        else {
            psCoreInstallPath =
            (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";
        }
```

corbob: Not quite...

```typescript
    if (platformDetails.operatingSystem === OperatingSystem.Windows) {
        if (use32Bit) {
           psCoreInstallPath =  (platformDetails.isProcess64Bit ? process.env.ProgramFiles(x86) : process.env.ProgramFiles) + "\\PowerShell";
           
        } else {
            psCoreInstallPath = (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";;
        }
```

The Program Files environment variables on my computer (in 64 bit PowerShell process):

```
ProgramFiles                   C:\Program Files // 64 bit Program Files
ProgramFiles(x86)              C:\Program Files (x86) // 32 bit Program Files
ProgramW6432                   C:\Program Files // 64 bit Program Files
```

Which might bring up a concern of the `(` and `)` in the variable name... might be worth pausing the debugger to inspect `process.env` to see what environment variables it has?

corbob: Not going to edit it, will just leave it here. To access the x86 Program Files Environment Variable, you seem to need to do `process.env['ProgramFiles(x86)']`

src/platform.ts
```diff
@@ -66,21 +66,49 @@ export function getDefaultPowerShellPath(
     use32Bit: boolean = false): string | null {
 
     let powerShellExePath;
+    let psCoreInstallPath;
 
     // Find the path to powershell.exe based on the current platform
```
bergmeister: This comment needs updating now
```suggestion
     // Find the path to the powershell executable based on the current platform
```

bergmeister:
This makes the assumption that the user has chosen the default MSI installation path. Ideally we should query the registry for it. However, currently the installer does not create a registry key for it, so we can't do that now. Going forward we should fix that though using something like this: https://stackoverflow.com/a/13451210/1810304
cc @TravisEz13

TylerLeonhardt:
Is it not here, in the registry @bergmeister?:

https://github.com/microsoft/vscode/blob/ffcb9e62b4365e652caeb05c80d4d7582833fdb8/src/vs/workbench/contrib/terminal/electron-browser/terminalService.ts#L120-L139

corbob:
> 
> 
> Is it not here, in the registry @bergmeister?:
> 
> https://github.com/microsoft/vscode/blob/ffcb9e62b4365e652caeb05c80d4d7582833fdb8/src/vs/workbench/contrib/terminal/electron-browser/terminalService.ts#L120-L139

Well that explains why Win+R > pwsh launches a different pwsh than cmd > pwsh. Windows Run looks to App Paths before it looks to the Path variable, and so it's launching 7 Preview for me, while cmd looks to the path only and thus launches 6.2.2.

I believe what @bergmeister is referring to is the `InstallLocation` value in `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{A55D1349-D294-43E8-B79A-1EC86DC3D18A}` for 6.2.2, and `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{EA126E54-972B-4EC0-A823-5D37B1284BCE}` for 7.0-preview. Personally the InstallLocation is where I would expect to find where the path actually is, but then you might need to maintain a list of GUIDs for each version?

TylerLeonhardt:
Yeah that makes sense - I think we should default to the install location (what Sydney already has) rather than switch to the App Paths.

src/platform.ts
```diff
@@ -66,21 +66,49 @@ export function getDefaultPowerShellPath(
...
+     // Find the path to the powershell executable based on the current platform
     // and the user's desire to run the x86 version of PowerShell
     if (platformDetails.operatingSystem === OperatingSystem.Windows) {
         if (use32Bit) {
-            powerShellExePath =
-                platformDetails.isOS64Bit && platformDetails.isProcess64Bit
-                    ? SysWow64PowerShellPath
-                    : System32PowerShellPath;
+            psCoreInstallPath =
+            (platformDetails.isProcess64Bit ? process.env["ProgramFiles(x86)"] : process.env.ProgramFiles)
```
rkeithhill: Minor-nit but to maintain the coding style this line and line 81 should be indented four more spaces.

src/platform.ts
```diff
@@ -66,21 +66,49 @@ export function getDefaultPowerShellPath(
...
+            + "\\PowerShell";
         } else {
-            powerShellExePath =
-                !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
-                    ? System32PowerShellPath
-                    : SysnativePowerShellPath;
+
```
rkeithhill: Remove blank line

src/platform.ts
```diff
@@ -66,21 +66,49 @@ export function getDefaultPowerShellPath(
...
+            psCoreInstallPath =
+            (platformDetails.isProcess64Bit ? process.env.ProgramFiles : process.env.ProgramW6432) + "\\PowerShell";
         }
+        if (fs.existsSync(psCoreInstallPath)) {
```
rkeithhill: Add blank line between separate if/else statements (also down on line 101)

src/platform.ts
```diff
@@ -66,21 +66,49 @@ export function getDefaultPowerShellPath(
...
+                powerShellExePath =
+                    platformDetails.isOS64Bit && platformDetails.isProcess64Bit
+                        ? SysWow64PowerShellPath
+                        : System32PowerShellPath;
+        } else {
+                powerShellExePath =
+                    !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
+                        ? System32PowerShellPath
+                        : SysnativePowerShellPath;
+            }
```
rkeithhill: This `}` doesn't seem to line up correctly.

test/platform.test.ts
```diff
@@ -46,7 +46,7 @@ suite("Platform module", () => {
 
             checkDefaultPowerShellPath(
                 platformDetails,
-                platform.System32PowerShellPath);
+                platform.getAvailablePowerShellExes(platformDetails, undefined)[0].exePath);
```
rkeithhill: It seems brittle to rely on a specific index in the array returned by `getAvailablePowerShellExes()`.  Someone might change that method in a way that reorders the elements in the array - which would seem like a safe thing to do. But it would break this code.  I would filter this by the versionName field or if that is inconvenient, perhaps add a new field to the objects returned by `getAvailablePowerShellExes` with the info you want to filter on.

This also applies to lines 79 and 109.

SydneyhSmith: I completely agree, just hadnt thought of a better way to do it...I will look into those options

rjmholt: How about something like:

```typescript
const psExePaths = platform.getAvailablePowerShellExes(platformDetails).reduce((acc, psExe) => {
    acc[psExe.versionName] = psExe
}, {});

...

checkDefaultPowerShellPath(platformDetails, psExePaths[<powershell name>]);
```

Ideally we'd name PowerShell versions by the version number and architecture (I doubt there's anyone using PowerShell 5.1 Core with VSCode)
```

File: src/platform.ts [hunk 9]
```diff
@@ -66,20 +66,49 @@ export function getDefaultPowerShellPath(
...
+            const arch = platformDetails.isProcess64Bit ? "(x64)" : "(x86)";
+            const psCorePaths =
+                fs.readdirSync(psCoreInstallPath)
+                .map((item) => path.join(psCoreInstallPath, item))
+                .filter((item) => {
+                    const exePath = path.join(item, "pwsh.exe");
+                    return fs.lstatSync(item).isDirectory() && fs.existsSync(exePath);
+                })
+                .map((item) => ({
+                    versionName: `PowerShell Core ${path.parse(item).base} ${arch}`,
```
TylerLeonhardt: Maybe this is a good time to drop the `Core` it's still technically correct to call it PowerShell 6... plus it future proofs us.

src/platform.ts
```diff
@@ -66,20 +66,49 @@ export function getDefaultPowerShellPath(
...
+                }));
+
+            if (psCorePaths) {
+                return powerShellExePath = psCorePaths[0].exePath;
+            }
+        }
+
+        if (use32Bit) {
+                powerShellExePath =
+                    platformDetails.isOS64Bit && platformDetails.isProcess64Bit
```
TylerLeonhardt: instead of storing this to a variable, you can return directly from here:

```suggestion
                return platformDetails.isOS64Bit && platformDetails.isProcess64Bit
```

src/platform.ts
```diff
@@ -66,20 +66,49 @@ export function getDefaultPowerShellPath(
...
+                        ? SysWow64PowerShellPath
+                        : System32PowerShellPath;
+        } else {
+                powerShellExePath =
+                    !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
```
TylerLeonhardt: instead of storing this to a variable, you can return directly from here:

```suggestion
                return !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
```

src/platform.ts
TylerLeonhardt: If you do this, make sure you also update `getAvailablePowerShellExes()` further down in the file

bergmeister:
> Is it not here, in the registry @bergmeister?:
> 
> https://github.com/microsoft/vscode/blob/ffcb9e62b4365e652caeb05c80d4d7582833fdb8/src/vs/workbench/contrib/terminal/electron-browser/terminalService.ts#L120-L139

Oh, I forgot we had this one as well. Normally apps, register a key for the uninstall location but the one that you references is perfectly fine and does the job as well :-)
From a standardization perspective, people would still expect to find the InstallLocation registry key as most (>50% of installers do), therefore I'd still recommend the PS installer to do that because people usually do something like this: https://stackoverflow.com/a/55229210/1810304

src/platform.ts
```diff
@@ -66,20 +66,47 @@ export function getDefaultPowerShellPath(
...
+                .map((item) => ({
+                    versionName: `PowerShell ${path.parse(item).base} ${arch}`,
+                    exePath: path.join(item, "pwsh.exe"),
+                }));
+
+            if (psCorePaths) {
+                return powerShellExePath = psCorePaths[0].exePath;
+            }
+        }
+
```
TylerLeonhardt:
```suggestion

        // No PowerShell 6+ detected so use Windows PowerShell.
```

src/platform.ts
```diff
@@ -66,20 +66,48 @@ export function getDefaultPowerShellPath(
...
+        // No PowerShell 6+ detected so use Windows PowerShell.
+        if (use32Bit) {
+                return platformDetails.isOS64Bit && platformDetails.isProcess64Bit
+                        ? SysWow64PowerShellPath
+                        : System32PowerShellPath;
+        } else {
```
rjmholt: No need for `else` here

src/platform.ts
```diff
@@ -66,20 +66,48 @@ export function getDefaultPowerShellPath(
...
+                return !platformDetails.isOS64Bit || platformDetails.isProcess64Bit
+                        ? System32PowerShellPath
+                        : SysnativePowerShellPath;
         }
     } else if (platformDetails.operatingSystem === OperatingSystem.MacOS) {
```
rjmholt: Or for the `else` here

PR #2266, opened 2019-11-01
[Ignore] Fix vsce changelog mangling (downgrade vsce)

rjmholt:
Temporary resolution for https://github.com/microsoft/vscode-vsce/issues/398.

Allows us to build the VSIX with the changelog working.


rjmholt:
Fix verified:

clog

PR #2502, opened 2020-02-27
Add new additions to .vscodeignore

rjmholt:
Looking into webpack (which looks unsuited to us unless it turns out to have a better way to manage assets that PowerShell needs to be in a certain folder configuration), I realised we're bundling stuff into the extension that we don't need or want.

This updates the `.vscodeignore` file to exclude them


.vscodeignore
```diff
@@ -15,3 +15,10 @@ out/test/**
 test/**
 sessions/**
 scripts/Install-VSCode.ps1
+tools/**
+.poshchan/**
+.github/**
+.vsts-ci/**
+build.ps1
+tsconfig.json
+tslint.json
```
rjmholt:
```suggestion
tslint.json

```

PR #2674, opened 2020-04-30
PSSA 1.19.0 settings

bergmeister:
## PR Summary

Pending PSSA `1.19.0` release and this PSES PR, but I tested it locally: https://github.com/PowerShell/PowerShellEditorServices/pull/1280

- Split the `powershell.codeFormatting.trimWhitespaceAroundPipe` setting into 2 settings since PSSA has changed the behaviour of the current setting to only add the missing white space around pipes but not remove any more. For trimming of redundant whitespace a new settings was added in PSSA (this way it would not be breaking if someone used PSSA 1.19.0 with the current version of the extension). I could've applied something similar here and only added the new setting but for better naming, I decided to rename the setting `powershell.codeFormatting.whitespaceAroundPipe` to `powershell.codeFormatting.trimWhitespaceAroundPipe`, this is a breaking change, can we help the user point to the new setting? The new setting is `powershell.codeFormatting.addWhitespaceAroundPipe` and is disabled by default)
- Add new setting `powershell.codeFormatting.whitespaceBetweenParameters` (disabled by default)
- Add new `None` option for `powershell.codeFormatting.pipelineIndentationStyle` and make it the default. This new setting means that the formatter will not change the user's pipeline indentation so that the user has to explicitly opt-in, this was due community feedback.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -656,10 +656,11 @@
           "enum": [
             "IncreaseIndentationForFirstPipeline",
             "IncreaseIndentationAfterEveryPipeline",
-            "NoIndentation"
+            "NoIndentation",
+            "None"
           ],
-          "default": "NoIndentation",
-          "description": "Multi-line pipeline style settings."
+          "default": "None",
+          "description": "Multi-line pipeline style settings (default: None)."
         },
         "powershell.codeFormatting.whitespaceBeforeOpenBrace": {
           "type": "boolean",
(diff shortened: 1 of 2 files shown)


bergmeister:
cc @adamdriscoll

PR #2689, opened 2020-05-10
🔧🚗 Migrate setting value of codeFormatting.whitespaceAroundPipe (if present) to new setting codeFormatting.addWhitespaceAroundPipe automatically.

bergmeister:
## PR Summary

This for the time of the preview with PSSA 1.19.0 to help migrate the setting automatically.
Because of settings sync, I do not want to remove the old setting yet, we should add removal of the old setting only for the RTM version of the PowerShell extension (where we will then also be able to give an informational window to the user)

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -323,7 +323,7 @@ export class SessionManager implements Middleware {
     }
 
     // During preview, populate a new setting value but not remove the old value.
-    // When the PowerShell extension releases the RTM version, then the old value can be safely removed.
+    // TODO: When the next stable extension releases, then the old value can be safely removed. Tracked in this issue: https://github.com/PowerShell/vscode-powershell/issues/2693
     private async migrateWhitespaceAroundPipeSetting() {
         const configuration = vscode.workspace.getConfiguration(utils.PowerShellLanguageId);
         const deprecatedSetting = 'codeFormatting.whitespaceAroundPipe'
--- a/src/session.ts
+++ b/src/session.ts
@@ -126,6 +126,8 @@ export class SessionManager implements Middleware {
 
         this.promptPowerShellExeSettingsCleanup();
 
+        this.migrateWhitespaceAroundPipeSetting();
+
         try {
             let powerShellExeDetails;
             if (this.sessionSettings.powerShellDefaultVersion) {
@@ -318,6 +320,18 @@ export class SessionManager implements Middleware {
             }
 
             return resolvedCodeLens;
+    }
+
+    // During preview, populate a new setting value but not remove the old value.
+    // When the PowerShell extension releases the RTM version, then the old value can be safely removed.
+    private async migrateWhitespaceAroundPipeSetting() {
+        const configuration = vscode.workspace.getConfiguration(utils.PowerShellLanguageId);
+        const deprecatedSetting = 'codeFormatting.whitespaceAroundPipe'
+        if (configuration.has(deprecatedSetting) && !configuration.has('codeFormatting.addWhitespaceAroundPipe')) {
+            const configurationTarget = await Settings.getConfigurationTarget(deprecatedSetting);
+            const value = configuration.get(deprecatedSetting, configurationTarget)
+            await Settings.change('codeFormatting.addWhitespaceAroundPipe', value, configurationTarget);
+        }
     }
 
     private async promptPowerShellExeSettingsCleanup() {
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -245,12 +245,30 @@ export function load(): ISettings {
...
+    else if (detail.workspaceValue !== undefined) {
+        configurationTarget = vscode.ConfigurationTarget.Workspace
+    }
+    else if (detail.globalValue !== undefined) {
+        configurationTarget = vscode.ConfigurationTarget.Global
+    }
+    return configurationTarget;
+}
+
+export async function change(
```
bergmeister, line 265: this function is functionally still the same , only the `configurationTarget` parameter has been optimized to match exactly the called VS-Code API underneath.

src/settings.ts
```diff
@@ -245,12 +245,30 @@ export function load(): ISettings {
     };
 }
 
-export async function change(settingName: string, newValue: any, global: boolean = false): Promise<void> {
-    const configuration: vscode.WorkspaceConfiguration =
-        vscode.workspace.getConfiguration(
-            utils.PowerShellLanguageId);
+export async function getConfigurationTarget(settingName: string): Promise<vscode.ConfigurationTarget> {
```
TylerLeonhardt, line 248: This is to get where the setting is currently set, yes?

A small comment would be appreciated

src/session.ts
```diff
@@ -320,6 +322,18 @@ export class SessionManager implements Middleware {
             return resolvedCodeLens;
     }
 
+    // During preview, populate a new setting value but not remove the old value.
```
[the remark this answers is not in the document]
bergmeister: Yes, that's exactly what I try to say: add new setting during preview but leave old setting still there (people who switch between preview/stable or use settings synchronization). Just before the next stable or in the last preview, I'd modify to to additionally remove the old setting.

src/settings.ts
bergmeister: Yes, it gets the scope (they call it `ConfigurationTarget`) from where the _effective_ setting is coming from (workspace settings override user settings, hence why the order is important here). If someone defined the old setting on both levels, we only look at the one that determines the _effective_ setting and change that one.

src/session.ts
```diff
@@ -320,6 +322,18 @@ export class SessionManager implements Middleware {
...
+    // When the PowerShell extension releases the RTM version, then the old value can be safely removed.
+    private async migrateWhitespaceAroundPipeSetting() {
```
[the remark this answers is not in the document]
bergmeister: I'd rather defer it to when it is needed as it might be a YAGNI.
Or maybe I should create an NPM package for it so that I can break the Internet later one 😂
https://www.zdnet.com/article/another-one-line-npm-package-breaks-the-javascript-ecosystem/

bergmeister:
@TylerLeonhardt Any ideas on the `Codacy` warnings? How can I run it locally. I've tried running eslint but it doesn't give me that warning?

src/settings.ts
```diff
@@ -251,21 +251,22 @@ export async function getEffectiveConfigurationTarget(settingName: string): Prom
 
     const detail = configuration.inspect(settingName);
     let configurationTarget = null;
-    if (typeof detail.workspaceFolderValue !== undefined) {
+    if (typeof detail.workspaceFolderValue !== "undefined") {
```
TylerLeonhardt, line 254: should this be this?
```suggestion
    if (detail.workspaceFolderValue !== undefined) {
```

bergmeister: No, `Codacy`/`ESLint` will complain because of not using `typeof` and not using double quotes around `undefined`:
https://eslint.org/docs/rules/no-undefined
Unfortunately, `Codacy` seems to lose its history but those were the complaints that made the build red in my previous commits.

TylerLeonhardt: I read the link. I think that advice is fine.

src/session.ts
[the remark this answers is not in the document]
TylerLeonhardt: Yes I think so. Can you capture this in an issue so we don't forget?

I'd even  suggest adding a TODO with the issue number... That way if we all disappear, someone could pickup where we left off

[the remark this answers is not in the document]
bergmeister: I opened issue #2693 and will add a comment

src/session.ts
```diff
@@ -320,6 +322,18 @@ export class SessionManager implements Middleware {
             return resolvedCodeLens;
     }
 
+    // During preview, populate a new setting value but not remove the old value.
+    // When the PowerShell extension releases the RTM version, then the old value can be safely removed.
```
bergmeister, line 326:
```suggestion
    // TODO: When the next stable extension releases, then the old value can be safely removed. Tracked in this issue: https://github.com/PowerShell/vscode-powershell/issues/2693
```

PR #2704, opened 2020-05-14
Add buttons for moving the terminal around.

MartinGC94:
## PR Summary

This adds 3 new buttons to the titlebar for moving the terminal around, like the ISE had.

## PR Checklist

- [X] Summarized changes
- [NA] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -265,6 +265,33 @@
         "command": "PowerShell.InvokeRegisteredEditorCommand",
         "title": "Invoke Registered Editor Command",
         "category": "PowerShell"
+      },
+      {
+        "command": "workbench.action.closePanel",
+        "title": "Close panel",
+        "category": "PowerShell",
+        "icon": {
+          "light": "resources/light/ClosePanel.svg",
+          "dark": "resources/dark/ClosePanel.svg"
+        }
+      },
+      {
+        "command": "workbench.action.positionPanelLeft",
+        "title": "Move panel left",
+        "category": "PowerShell",
+        "icon": {
+          "light": "resources/light/MovePanelLeft.svg",
+          "dark": "resources/dark/MovePanelLeft.svg"
+        }
+      },
+      {
+        "command": "workbench.action.positionPanelBottom",
+        "title": "Move panel to bottom",
+        "category": "PowerShell",
+        "icon": {
+          "light": "resources/light/MovePanelBottom.svg",
+          "dark": "resources/dark/MovePanelBottom.svg"
+        }
       }
     ],
     "menus": {
@@ -305,6 +332,21 @@
         }
       ],
       "editor/title": [
+        {
+          "when": "editorLangId == powershell",
+          "command": "workbench.action.positionPanelBottom",
+          "group": "navigation@97"
+        },
+        {
+          "when": "editorLangId == powershell",
+          "command": "workbench.action.positionPanelLeft",
+          "group": "navigation@98"
+        },
+        {
+          "when": "editorLangId == powershell",
+          "command": "workbench.action.closePanel",
+          "group": "navigation@99"
+        },
         {
           "when": "editorLangId == powershell",
           "command": "workbench.action.debug.start",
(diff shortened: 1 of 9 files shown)


rjmholt:
These should probably only be enabled in ISE mode to prevent complicating the current layout for users using the default, non-ISE-like functionality

TylerLeonhardt:
Can you add a screenshot for what this looks like?

MartinGC94:
Dark:
Dark
Light:
Light

Personally I don't think the Powershell extension has so many buttons that it feels too complicated/bloated but I do like the idea of giving users the option for hiding these buttons as well as the run buttons previously added.

rkeithhill:
That looks very nice!  But I agree with @rjmholt, I wouldn't add these except when in `ISE mode`.  And/or create a setting to control the visibility of these buttons. These three buttons aren't a lot but the problem is that other extensions also add buttons. My current button set up looks like this:

image

Having more buttons would cut further into space I have for editor tabs.

MartinGC94:
Good point about other extensions also being able to add their own buttons. I've added options for hiding all of the buttons and made the new buttons hidden by default.

rjmholt:
@MartinGC94 one last thing you might like to add is logic to activate all the buttons when ISE mode is set. That happens here:

https://github.com/PowerShell/vscode-powershell/blob/939379d13c1fbc1e1ce87cbc46606ba02e2b314d/src/features/ISECompatibility.ts#L20-L27

MartinGC94:
@rjmholt I did that for the new buttons I just forgot to mention it:
https://github.com/MartinGC94/vscode-powershell/blob/346f1b3a7770ff6e85d56ad4277fde4f6f8ac518/src/features/ISECompatibility.ts#L27-L29

Should I add the old buttons as well? I thought about doing that but didn't see any reason to do so as long as they are enabled by default.

rjmholt:
>  I did that for the new buttons I just forgot to mention it:

Oh! Sorry I didn't see it when I looked at the diff before.

> Should I add the old buttons as well?

No no, all good! Just meant the new buttons 🙂

package.json
```diff
@@ -800,6 +842,31 @@
...
+        },
+        "powershell.buttons.MovePanelLeftButtonVisibility": {
+          "type": "boolean",
+          "default": false,
+          "description": "Show the Move panel left button in the editor titlebar."
+        },
+        "powershell.buttons.MovePanelToBottomButtonVisibility": {
+          "type": "boolean",
+          "default": false,
+          "description": "Show the Move panel to bottom button in the editor titlebar."
```
TylerLeonhardt, line 869: This is really awesome and gives the user a ton of flexibility but I don't think some is going to want just 1 button or just 2 buttons from what appears to be two groups here:

`powershell.buttons.showRunButtons`
`powershell.buttons.showPanelMovementButtons`

Can you change it to these 2 settings and use these everyewhere? I discussed this with @rjmholt and @SydneyhSmith

MartinGC94: Sure, it still fits my needs so I don't mind limiting the options a bit. I'm just curious why limit it if the work has already been done? I can imagine there's people out there that never use the vertical split view or always use the run selection option that might not care for those buttons.

TylerLeonhardt: I'm very wary to add configuration because often times people get lost in it. There have been many instances where folks just don't realize that settings already exist or find themselves digging trying to figure out the right configuration for them.

I do want there to be _some_ way to toggle showing these buttons and that's kinda why we want 2 settings instead of 5.

I know the difference is only 3 but we try to have this mindset with all new settings.

>  I can imagine there's people out there that never use the vertical split view or always use the run selection option that might not care for those buttons.

Maybe. But those same people might be just as happy seeing all the buttons... in the off chance that they use the other.

It's in situations like this where I go "let's expose the least amount of stuff that makes sense (in other words 2 settings) and wait for a user to come around to ask for more."

MartinGC94: I understand and I've fixed it now.

TylerLeonhardt: Thanks @MartinGC94!

PR #2796, opened 2020-07-05
Add setting to change the cwd of the Powershell Integrated Console

### LINKED ISSUE #2227: Set start path
The start path of the integrated console is user's home path. It will be convenient if I can set the start path to be the current file folder. I found that this can be set in vs code terminal, but not in PowerShell Integrated Console. In vs code terminal, the configuration is
```
// Open the terminal in the currently opened file's directory
  "terminal.integrated.cwd": "${fileDirname}"
```
I just need a similar configuration for integrated console, such as 
`"powershell.cwd": "${fileDirname}"`

jwfx:
## PR Summary

This adds a new setting that will allow users to set the current working directory for the Powershell Integrated Console

Resolves #2227

## PR Checklist

- [x] Summarized changes
- [ ]  `NA` PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -643,7 +643,7 @@
         "powershell.cwd": {
           "type": "string",
           "default": null,
-          "description": "An explicit start path where the Powershell Integrated Console will be launched. Open workspace folders will take precedence over this setting. Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory)."
+          "description": "An explicit start path where the Powershell Integrated Console will be launched. Both the PowerShell process and the shell's location will be set to this directory. Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory)."
         },
         "powershell.scriptAnalysis.enable": {
           "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -781,6 +781,11 @@
           "default": false,
           "description": "Do not show the Powershell Integrated Console banner on launch"
         },
+        "powershell.integratedConsole.cwd": {
+          "type": "string",
+          "default": null,
+          "description": "An explicit start path where the Powershell Integrated Console will be launched. Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory"
+        },
         "powershell.debugging.createTemporaryIntegratedConsole": {
           "type": "boolean",
           "default": false,
--- a/package.json
+++ b/package.json
@@ -640,6 +640,11 @@
           "default": "BlockComment",
           "description": "Controls the comment-based help completion behavior triggered by typing '##'. Set the generated help style with 'BlockComment' or 'LineComment'. Disable the feature with 'Disabled'."
         },
+        "powershell.cwd": {
+          "type": "string",
+          "default": null,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -781,6 +781,11 @@
           "default": false,
           "description": "Do not show the Powershell Integrated Console banner on launch"
         },
+        "powershell.integratedConsole.cwd": {
```
rjmholt, line 784: The cwd of the process will affect more than just the integrated console, it will affect completions as well for example. So it might be better to name this less specifically for the integrated console. I believe the original issue suggested `"powershell.cwd"`.

jwfx:
I configured the new setting like so:
image

After opening a .ps1 file in a different location the Integrated Console presented itself like this:
image

Additionally I checked the current directory of the pwsh process:
image

Does this meet the expectations and I can proceed with relocating the setting to `powershell.cwd`?

TylerLeonhardt:
@jwfx when you open vscode, do you have a folder opened or does it say:

image

jwfx:
@TylerLeonhardt 

I have no folder open. As soon as I have an open folder, the folder path takes precedence and the Integrated Console is launched in that opened folder, no matter where the .ps1 is located I'm trying to open.

Maybe that behavior could be mentioned in the setting description?

rjmholt:
> Maybe that behavior could be mentioned in the setting description?

rjmholt:
The integrated console's location is set [here](https://github.com/PowerShell/PowerShellEditorServices/blob/8cfbf2b4eb636956c31ba6ad258bb3d1b09155eb/src/PowerShellEditorServices/Server/PsesLanguageServer.cs#L117-L124). Changing the behaviour needs to pull the setting from the configuration service and if it's not-null, use that

jwfx:
It seems like open workspace folders seem to override the location that is passed into createTerminal(). The process cwd is still applied though.

I've changed the PR so that it is at least consistent for now. The `powershell.cwd` setting is only applied if no folders are open.

I'd be thankful for pointers how to make the cwd setting also work for open workspace folders. I'm not sure if this even in the scope of this extension? Just guessing here.. I am new to this.

jwfx:
@rjmholt 

Ok, I think I got it.

I couldn't make the required changes [here](https://github.com/jwfx/PowerShellEditorServices/blob/63f5c0d30fd917e6a0df4be257320acbfb801dd0/src/PowerShellEditorServices/Server/PsesLanguageServer.cs#L95), because the configuration values are not available at this point.

Should I turn this into a PR on the PSES repo?
https://github.com/jwfx/PowerShellEditorServices/commit/63f5c0d30fd917e6a0df4be257320acbfb801dd0

TylerLeonhardt:
@jwfx you need changes in both to make this all come together. This makes sense because vscode-powershell and PSES work together.

jwfx:
Not sure if I can do anything about those failed checks..?

rjmholt:
> Not sure if I can do anything about those failed checks..?

They're bugs in PowerShell, being addressed there. Hopefully we can merge that soon and get the checks passing

PR #2828, opened 2020-07-23
Update version in package.lock.json to 2020.7.0 and resolve npm audit warnings

bergmeister:
## PR Summary

When you bump versions, do you have a script that does it? Because it should do it for both `package.json` and `package.lock.json`.
This keeps coming up that `package.lock.json` gets behind but is a tad annoying as it shows as a diff when doing a clean re-compile...

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1325,12 +1325,20 @@
       "dev": true
     },
     "mkdirp": {
-      "version": "0.5.1",
-      "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.1.tgz",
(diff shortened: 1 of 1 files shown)


rjmholt:
We should actually be running `npm` when `package.json` is updated to ensure all packages are updated properly. The problem isn't that a script is updating the package.json, it's that we're not running the thing that updates the package-lock.json

bergmeister:
The build itself uses npm, which updates the package-lock.json but only in the case of clean build where there is no local `node_modules` cache

rjmholt:
> The build itself uses npm, which updates the package-lock.json

Oh I mean that we're not running the build when package.json is updated, and we should

PR #2871, opened 2020-08-06
Prevent `Update Notebook dts` action from triggering on forked repositories

corbob:
Similar to https://github.com/PowerShell/PowerShell/pull/12763

## PR Summary

Update GitHub Actions Workflow to only trigger on the main vscode-powershell repository.

Note: I have not tested this, merely modeled it after the [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-contexts) and the PowerShell PR.

## PR Checklist

- [x] Summarized changes
- [NA] PR has tests

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -11,6 +11,7 @@ jobs:
 
     # CodeQL runs on ubuntu-latest and windows-latest
     runs-on: ubuntu-latest
+    if: github.repository == 'PowerShell/vscode-powershell'
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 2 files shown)


TylerLeonhardt:
I was kinda hoping this would prevent the entire action from running. In the other PR the `if` is at a higher level.

Can you follow that?

corbob:
I've moved the if up to the steps level. I tried it in the on level, but it didn't like that. I also added it to the codeql workflow so it too only runs on the main repository. From my testing it works:
image

PR #2972, opened 2020-09-24
Ctrl + Shift + P

jeffpatton1971:
That's how you launch the command palette

## PR Summary

Correcting what appears to be an old way of calling the Command Palette, the other option would be to use Ctrl + Shift + B to invoke the Builder.

## PR Checklist

- [ ] Summarized changes
- [ ] PR has tests

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -24,7 +24,7 @@ to before you can proceed.
 
 #### From Visual Studio Code:
 
-Press <kbd>Ctrl</kbd>+<kbd>P</kbd> and type `task build`
+Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> and type `task build`
 
 This will compile the TypeScript files in the project to JavaScript files.
 


TylerLeonhardt:
The steps are correct already:

`ctrl`+`P` -> `task build` 

image

TylerLeonhardt:
`ctrl`+`P` is different than `ctrl`+`shift`+`P`

jeffpatton1971:
@TylerLeonhardt Ctrl + P doesn't do anything when I type task build, nothing shows up. However, the command palette i get those options...I'm not sure what the difference is between your UI and mine, I have the latest update.

TylerLeonhardt:
I have the lastest version of vscode and I have the `vscode-powershell` repo opened. Just tried on a second machine:

image

jeffpatton1971:
Ok, so what it looks like I'm seeing is an extension issue that is causing the Ctrl + P to take over a minute before it figures out what I want after I type. So when I first attempting to do this Ctrl + P was doing nothing, and I didn't bother waiting around, when I used the command palette I had no issue at all. Feel free to reject this if you like, I'm not really married to it now that I see it's not opening because one of the PowerShell extensions appears to be slowing things down.

TylerLeonhardt:
if you type `ctrl`+`shift`+`p` and then hit backspace (to get rid of the `>`) then type "task" that's another way to get to that. But yes the tasks like a tiny bit to show up - sounds  like it's longer for you. That's something that would be a performance issue of vscode.

Issue #3195, opened 2021-02-22
Release Automation Improvements


SydneyhSmith:
We should clean up/optimize the release automation for the VSCode extension in order to reduce the time dedicated to each release so that we can optimize our time for improving the extension and release at a more frequent cadence. 

cc: @rjmholt @andschwa

andschwa:
- [ ] Replace VSES vars with trigger
- [ ] Force-update branch (so changelog generator can be re-run)
- [ ] Handle “v” prefixes and “-preview” suffixes
- [ ] Sub out GitHubTools with PowerShell for GitHub
- [ ] Handle merge PRs (not just squashes)
- [ ] Misuses URL when it can’t find the right one
- [ ] Delete old script
- [ ] Fix version scripts (package.json, lock file, psd1 all need to be bumped)
- [ ] Fix publishGH to handle emoji changelog and find correct repo’s changelog
- [x] Ignore more bots
- [x] Open draft PR

rjmholt:
We should slurp up any relevant parts of https://github.com/PowerShell/vscode-powershell/issues/2286 into this issue and close that one

andschwa:
@rjmholt I have like 3 remaining bullet points to do, I'll get those in before merging the PR.

andschwa:
Uhh wtf my bulleted list of tasks are gone. Well, I made a new issue for the remaining piece, plus there's to delete the 'vses' scripts after we're happy with the pipeline.

andschwa:
I think GitHub's having issues, adding this as I found it in an unrefreshed tab:

 Handle merge PRs (not just squashes)
  Sub out GitHubTools with PowerShell for GitHub
  Fix version scripts (package.json, lock file, psd1 all need to be bumped)
  Fix publishGH to handle emoji changelog
  Force-update branch (so changelog generator can be re-run)
  Handle “v” prefixes and “-preview” suffixes
  Misuses URL when it can’t find the right one
  Replace VSES vars with trigger (after testing)
  Fix publishGH to find correct repo’s changelog
  Delete old script
  Add pipeline trigger so PSES triggers VSCE
  Ignore more bots
  Open draft PR

PR #3197, opened 2021-02-23
Some improvements to the release scripts

andschwa:
Part of #3195


tools/GitHubTools.psm1
```diff
@@ -85,7 +85,7 @@ function GetGitHubHeaders
         $Accept
     )
 
-    $headers = @{}
+    $headers = @{"Content-Type" = "UTF-8"}
```
rjmholt, line 88: The right fix for this is for us to move over to [PowerShellForGitHub](https://github.com/microsoft/PowerShellForGitHub), but:

```suggestion
    $headers = @{"Content-Type" = "application/json; charset=utf-8"}
```

rjmholt (approved):

tools/GitHubTools.psm1
rjmholt: See also:
- [MDN: Content-Type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
- [MDN: Common MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)

andschwa: OMG, thank you.

Issue #3181, opened 2021-02-13
microsoft store installed powershell can not be recognized


Wongboo:
I used to use powershell delivered by github release, now as I favor the convenience of microsoft-store's integrated update, I switch to powershell in microsoft store.
But then bugs appear:
1. *vscode-powershell* cannot **automatically** recognize microsoft store installed powershell
2. even if I changed “PowerShell Additional Exe Paths” in JSON with the value
  $PSHOME\pwsh.exe which is
 "C:\\Program Files\\WindowsApps\\Microsoft.PowerShellPreview_7.2.2.0_x64__8wekyb3d8bbwe\\pwsh.exe"
 *vscode-powershell* cannot recognize microsoft store installed powershell **as well**

Wongboo:
I tried to change   "terminal.integrated.shell.windows" as well,but no use

Softgrider:
This happens since updated to Visual Studio Code Version 1.53. 
Uninstall Store Version and install msi Version brings the powershell 7.1.x back

rjmholt:
The logic to automatically pick up a store installation is here:

https://github.com/PowerShell/vscode-powershell/blob/7f0b9696d74a6813253ff54f3d64eabd3465b68e/src/platform.ts#L262-L290

It's not clear why this would suddenly fail, especially due to a VSCode update (since that shouldn't update anything within the extension itself).

It's also mysterious why giving an explicit absolute path would also fail -- that's effectively a bypass for the PowerShell-finding logic and should work no matter where it's pointed (as long as that's a valid PowerShell executable).

SydneyhSmith:
Thanks @Softgrider just to clarify did this issue not occur for you before 1.53? Thanks @Wongboo for the bug report

rjmholt:
So we've now had a second report of this, and I'm wondering if there's a permissions issue or similar going on -- hopefully we can reproduce this locally to investigate

wjvii:
@rjmholt I can confirm the same behavior here as well.

rjmholt:
So it looks like this is due to [`fs.existsSync` returning false for symlinks](https://stackoverflow.com/questions/23253332/check-if-a-symlink-exists-in-node-js), causing this to be false here:

https://github.com/PowerShell/vscode-powershell/blob/ce41617eb6eb10cb41b9d6076593490617267e59/src/platform.ts#L494

It's not clear why this previously worked, but either node or Windows/the store has changed its behaviour here.

[...]

PR #3202, opened 2021-02-23
Fix PowerShell MSIX (Store) detection

rjmholt:
## PR Summary

Fixes https://github.com/PowerShell/vscode-powershell/issues/3181.

The MSIX exe is a symlink and node's file test API returns false for those.

This fixes that so we now detect the PowerShell MSIX installation properly again.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -268,7 +268,7 @@ export class PowerShellExeFinder {
         // Find the base directory for MSIX application exe shortcuts
         const msixAppDir = path.join(process.env.LOCALAPPDATA, "Microsoft", "WindowsApps");
 
-        if (!fs.existsSync(msixAppDir)) {
+        if (!fileExistsSync(msixAppDir)) {
             return null;
         }
 
@@ -310,7 +310,13 @@ export class PowerShellExeFinder {
         const powerShellInstallBaseDir = path.join(programFilesPath, "PowerShell");
 
         // Ensure the base directory exists
-        if (!(fs.existsSync(powerShellInstallBaseDir) && fs.lstatSync(powerShellInstallBaseDir).isDirectory())) {
+        try {
+            const powerShellInstallBaseDirLStat = fs.lstatSync(powerShellInstallBaseDir);
+            if (!powerShellInstallBaseDirLStat.isDirectory())
+            {
+                return null;
+            }
+        } catch {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


src/platform.ts
```diff
@@ -268,7 +268,7 @@ export class PowerShellExeFinder {
         // Find the base directory for MSIX application exe shortcuts
         const msixAppDir = path.join(process.env.LOCALAPPDATA, "Microsoft", "WindowsApps");
 
-        if (!fs.existsSync(msixAppDir)) {
```
andschwa, line 271: Good call to remove this anyway:

> [fs.existsSync() will be deprecated.](https://node.readthedocs.io/en/latest/api/fs/#fsexistssyncpath)

src/platform.ts
```diff
@@ -469,6 +475,17 @@ export function getWindowsSystemPowerShellPath(systemFolderName: string) {
         "powershell.exe");
 }
 
+function fileExistsSync(filePath: string): boolean {
```
andschwa, line 478: The node docs aren't wrong about this kinda being an anti-pattern.

> fs.exists() is an anachronism and exists only for historical reasons. There should almost never be a reason to use it in your own code.
> 
> In particular, checking if a file exists before opening it is an anti-pattern that leaves you vulnerable to race conditions: another process may remove the file between the calls to fs.exists() and fs.open(). Just open the file and handle the error when it's not there.
> 
> fs.exists() will be deprecated.

We're essentially making our own `fs.existsSync` and I question if we need it. As the code is written right now, this does seem to be the best option, but I'd want to review why we're checking if a file exists, then checking if a folder exists, etc.

src/platform.ts
```diff
@@ -469,6 +475,17 @@ export function getWindowsSystemPowerShellPath(systemFolderName: string) {
...
+    try {
+        // This will throw if the path does not exist,
+        // and otherwise returns a value that we don't care about
+        fs.lstatSync(filePath);
```
andschwa, line 482: This probably ought to just be `fs.statSync` since we don't care about the information, but we _may_ care that the target file exists. I don't really care though, as the code as written is (presumably) tested, and it's not really our problem if the symlink exists but its target does not.

andschwa (approved):
I don't think we'll make it any worse 😂

andschwa:
Weird, @rjmholt the CI tests passed on this PR but then failed on the merge to master.

PR #3222, opened 2021-03-04
Fix hardcoded extension name string

### LINKED ISSUE #3208: Extension name is hardcoded, leading to failing tests
When the extension name is switched from `powershell-preview` to just `powershell` (such as when releasing a stable version like in #3207), the tests fail because of hardcoded constants like `const testExtensionId = "ms-vscode.powershell-preview";`. We should instead read the current extension ID from the `package.json` file, which is already done in [`main.ts`](https://github.com/PowerShell/vscode-powershell/blob/f22659720ffd1a11e85416554031071adb4aebef/src/main.ts#L40).

andschwa:
Fixes #3208.

### DIFF
--- a/test/features/ExternalApi.test.ts
+++ b/test/features/ExternalApi.test.ts
@@ -6,7 +6,9 @@ import * as vscode from "vscode";
 import { before, beforeEach, afterEach } from "mocha";
 import { IExternalPowerShellDetails, IPowerShellExtensionClient } from "../../src/features/ExternalApi";
 
-const testExtensionId = "ms-vscode.powershell-preview";
+// tslint:disable-next-line: no-var-requires
+const PackageJSON: any = require("../../../package.json");
+const testExtensionId = `${PackageJSON.publisher}.${PackageJSON.name}`;
 
 suite("ExternalApi feature - Registration API", () => {
     let powerShellExtensionClient: IPowerShellExtensionClient;
@@ -82,7 +84,7 @@ suite("ExternalApi feature - Other APIs", () => {
     });
 
     beforeEach(() => {
-        sessionId = powerShellExtensionClient.registerExternalExtension("ms-vscode.powershell-preview");
+        sessionId = powerShellExtensionClient.registerExternalExtension(testExtensionId);
     });
 
     afterEach(() => {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,10 +1,10 @@
...
-  "name": "powershell-preview",
-  "displayName": "PowerShell Preview",
+  "name": "powershell-test",
+  "displayName": "PowerShell",
   "version": "2021.2.3",
-  "preview": true,
+  "preview": false,
   "publisher": "ms-vscode",
-  "description": "(Preview) Develop PowerShell scripts in Visual Studio Code!",
+  "description": "Develop PowerShell scripts in Visual Studio Code!",
```
rjmholt, line 7: It makes total sense to me to use a third name for dev builds of the extension.

- Possible alternative to `powershell-test`: `powershell-dev`
- My suggestion would be for `"preview": true`, since anything that isn't stable is essentially a preview of some kind
- Are we still changing the description based on the channel of the extension? If so, might be worth having a note or tag for the dev build
- We probably want something in place to prevent publication of the dev build to the marketplace

test/features/ExternalApi.test.ts
```diff
@@ -6,7 +6,9 @@ import * as vscode from "vscode";
 import { before, beforeEach, afterEach } from "mocha";
 import { IExternalPowerShellDetails, IPowerShellExtensionClient } from "../../src/features/ExternalApi";
 
-const testExtensionId = "ms-vscode.powershell-preview";
+// tslint:disable-next-line: no-var-requires
+const PackageJSON: any = require("../../../package.json");
```
rjmholt, line 10: We should make sure that this still works when the extension is bundled to a VSIX (or possibly webpacked)

rjmholt (approved):

package.json
andschwa: This was just a test commit, I'm rolling it back lol.

test/features/ExternalApi.test.ts
andschwa: It should, since this is just test code. Are the tests included in the VSIX?

rjmholt: I don't think so, but I'm not sure. Possible to find out by decompressing the VSIX and inspecting it.

andschwa: LOL it's just a zip, who would have thought 🤦

andschwa: Yup, we're good!

PR #3275, opened 2021-04-05
Update vsce to 1.87.1

andschwa:
Replaces #3273.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "sinon": "~10.0.0",
         "tslint": "~6.1.3",
         "typescript": "~4.2.3",
-        "vsce": "~1.85.1",
+        "vsce": "~1.87.1",
         "vscode-test": "~1.5.2"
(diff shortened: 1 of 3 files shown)


rjmholt (approved):

rjmholt:
We should also open an issue on vsce just to register that they broke us again

andschwa:
This brings in https://github.com/microsoft/vscode-vsce/commit/83163ed09e26b2218832f60acbe7aad215c77a02 which fixes a subtle break in our changelog where, on the marketplace, all the #xyzs are auto-linked directly and don't obey the actual link we put in the changelog (which was supposed to be fixed, but there was a weird implicit behavior of their [commander dependency](https://github.com/tj/commander.js/#other-option-types-negatable-boolean-and-booleanvalue) that caused it not to be, but now it's fixed, https://github.com/microsoft/vscode-vsce/issues/542).

PR #3286, opened 2021-04-09
Fix release build pipeline

### LINKED ISSUE #3282: 2021.4.1-preview Includes an old version of PowerShell Editor Services
The latest Powershell extension preview changelog includes updates about editor services but those changes aren't actually in this release.  
If the file version of Microsoft.PowerShell.EditorServices.dll can be trusted then the version used is 2.3.0 whereas the version I get when I manually build the extension from the source is 2.4.0-preview.1 so either there's some sort of build/release issue or the patch notes are wrong.

andschwa:
Fixes #3282.


andschwa:
From commit:

> A backlog task of removing the `findPsesBuild` script and associated logic
> caused us to end up with a release with the wrong artifacts. The tagged PSES
> build was partially succeeded, which the download task then ignored as the
> trigger, and instead of failing, fell back to this script which gave it the
> latest fully succeeded build.
> 
> That PSES build was only partially successful because the "Install PowerShell
> Daily" step failed due to a transient network error, and so this was
> intentionally ignored as it did not affect the build. Or so we thought.

.vsts-ci/templates/release-general.yml
```diff
@@ -1,40 +1,20 @@
...
   condition: succeededOrFailed()
 
-# TODO: Use modern resources for these variables.
-- task: PowerShell@2
-  displayName: 'Set environment variables for VSTS (Phase 1)'
-  inputs:
-    targetType: filePath
-    filePath: ./vscode-powershell/tools/releaseBuild/setVstsVariables.ps1
-
-# TODO: Remove this when satisfied with added trigger.
```
andschwa, line 20: @rjmholt This `TODO` bit us good.

andschwa:
Ok so the next preview still has the incorrect version field in the PSD1 file, but that's bug https://github.com/PowerShell/PowerShellEditorServices/issues/1455,  I confirmed the DLLs are correct and double-checked the build IDs that were included by the pipeline:

```
> monodis --assembly ./Microsoft.PowerShell.EditorServices.dll

Assembly Table
Name:          Microsoft.PowerShell.EditorServices
Hash Algoritm: 0x00008004
Version:       2.4.0.0 
```

andschwa:
@rjmholt I pulled the version bump out, I'd like to just merge this, and then do the release.

rjmholt (approved):

PR #3319, opened 2021-04-28
Fix triggers for CodeQL

### LINKED ISSUE #3316: Fix CodeQL with Dependabot
Now that Dependabot is opening PRs again (per #3310), I can't seem to merge them because CodeQL fails with this error:

> Error: Workflows triggered by Dependabot on the "push" event run with read-only access. Uploading Code Scanning results requires write access. To use Code Scanning with Dependabot, please ensure you are using the "pull_request" event for this workflow and avoid triggering on the "push" event for Dependabot branches. See https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#scanning-on-push for more information on how to configure these events.

See #3313 for an example.

andschwa:
The push event should only be registered for the master branch and for tags. It was erroneously running on every branch, breaking Dependabot.  For PRs we exclude documentation changes only. The schedule was set to run once a week, but we may as well run every morning.

Resolves #3316.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,9 +2,17 @@ name: "Code scanning - action"
 
 on:
   push:
+    branches:
+      - master
+    tags:
+      - v*
   pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
   schedule:
-    - cron: '0 19 * * 0'
+    - cron: '00 14 * * *' # Every morning at 7:00am PDT
 
 jobs:
   CodeQL-Build:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -2,9 +2,17 @@ name: "Code scanning - action"
...
+    tags:
+      - v*
   pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
   schedule:
-    - cron: '0 19 * * 0'
+    - cron: '00 02 * * *'
```
rjmholt, line 15: Can you add a comment describing when this means -- mainly because I'm too lazy to look up crontab syntax

rjmholt (approved):

.github/workflows/codeql.yml
andschwa: I was also off 12 hours, fixed!

PR #3321, opened 2021-04-28
Update CodeQL GitHub workflow

andschwa:
This updates the workflow to be consistent with the newly setup workflow in PowerShellEditorServices, which is much simpler. See https://github.com/PowerShell/PowerShellEditorServices/pull/1464


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,42 @@
...
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
```
rjmholt, line 24: Ditto

rjmholt (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,42 @@
...
+      matrix:
+        language: [ 'javascript' ]
```
andschwa, line 26: I'm double checking that this isn't supposed to be 'typescript'.

andschwa: I guess it could be either lol https://github.com/github/codeql-action/blob/03f029c2a15a75961a2724371dcbbab4bd33ce47/src/languages.ts#L16

andschwa: Testing...

PR #3359, opened 2021-05-21
Exclude `-` and `$` from PowerShell `wordSeparators`

andschwa:
This is only a supplied default, so users can easily override it.

This suggestion came from a "Getting Started with VS Code for PowerShell" presentation by @FriedrichWeinmann, thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -895,6 +895,11 @@
         }
       }
     },
+    "configurationDefaults": {
+      "[powershell]": {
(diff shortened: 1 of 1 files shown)


rjmholt (approved):

jhoneill:
This is mentioned in the release notes for the latest extension, it's the right default (I prefer the non default behaviour)

What I found frustrating is the terminal window has its own behaviour. Extensions shouldn't change this (because each one might change it a different way) but if the release notes mentioned that Terminal › Integrated: Word Separators in settings may be set to match what have for your editor that would save other people hunting for it.

PR #3372, opened 2021-05-26
Note that ISE theme is included in this extension

### LINKED ISSUE #1943: Consider moving the PowerShell ISE theme into a separate extension
* install this extension
* :bug: theme switches

This is really uncool - while I can understand your excitement about your theme you should understand that not everyone is equally excited about it. The extension already pops up a notification message and that would be a much more subtle place for "selling" your theme 

Screenshot 2019-05-03 at 12 39 41

andschwa:
Resolves #1943.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -3,23 +3,30 @@
 This document contains troubleshooting steps for commonly reported issues when using the
 [PowerShell Extension] for Visual Studio Code.
 
-## Script Analysis is Reporting False Errors
+## How do I debug my PowerShell script?
 
-Script analysis is provided by the [PSScriptAnalyzer] project on GitHub.
-Please [open an issue there] if you are getting fault script diagnostics
-(red and green squiggly lines under PowerShell in scripts).
+This topic is best covered in the "Debugging PowerShell Script in Visual Studio Code"
+Scripting Guys blog posts (thanks community!):
 
-## Problems with Syntax Highlighting
+* [Part 1](https://blogs.technet.microsoft.com/heyscriptingguy/2017/02/06/debugging-powershell-script-in-visual-studio-code-part-1/)
+* [Part 2](https://blogs.technet.microsoft.com/heyscriptingguy/2017/02/13/debugging-powershell-script-in-visual-studio-code-part-2/)
 
-PowerShell syntax highlighting is not performed by the [PowerShell Extension].
-Instead syntax highlighting for VSCode, Atom, SublimeText and even GitHub is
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -3,13 +3,21 @@
...
+* [Part 1](https://blogs.technet.microsoft.com/heyscriptingguy/2017/02/06/debugging-powershell-script-in-visual-studio-code-part-1/)
+* [Part 2](https://blogs.technet.microsoft.com/heyscriptingguy/2017/02/13/debugging-powershell-script-in-visual-studio-code-part-2/)
+
+## Script analysis is reporting false errors
+
+Script analysis is provided by the [PSScriptAnalyzer] project on GitHub. If the warning
+message starts with `[PSScriptAnalyzer]` or if you are getting faulty script diagnostics
+(red and green squiggly lines under PowerShell in scripts) please [open an issue there].
+
+## Problems with syntax highlighting
```
rjmholt, line 20: Ah, this is no longer true

docs/troubleshooting.md
```diff
@@ -30,7 +38,19 @@ were, set:
 If you leave it enabled, you can customize the colors used for the various
 tokens. See [#3221] for more info and to leave suggestions.
 
-## VSCode is not working like the ISE
+## PSReadline features aren't working
+
+Integration of [PSReadline] into the PowerShell Integrated Console is under active
```
rjmholt, line 43: Also no longer true

docs/troubleshooting.md
```diff
@@ -45,7 +65,7 @@ VSCode itself and can't be changed by the extension.
 The VSCode maintainers are quite reasonable though,
 and you can ask for new features [in their repository](https://github.com/Microsoft/vscode).
 
-## Known Issues in the Extension
+## Known issues in the extension
```
rjmholt, line 68: We should review these

rjmholt (commented):
The rest of this doc is still fairly out of date. We could change it in this PR or do it more incrementally, as you prefer.

docs/troubleshooting.md
andschwa: PSSA or syntax highlighting?

docs/troubleshooting.md
andschwa: Definitely 😄

docs/troubleshooting.md
andschwa: Updated the latter with further clarification, since I'm quite sure script analysis is still PSSA's domain.

rjmholt (approved):

PR #3458, opened 2021-07-13
Automate more of the release

andschwa:
This makes the `ReleaseTools` module almost entirely automated, which sets us up for modifying our release pipelines to create the GitHub releases and upload the assets/binaries directly.

### DIFF
--- a/tools/ReleaseTools.psm1
+++ b/tools/ReleaseTools.psm1
@@ -202,8 +202,7 @@ function Update-Changelog {
         Where-Object { -not $_.user.UserName.EndsWith("[bot]") } |
         Where-Object { "Ignore" -notin $_.labels.LabelName } |
         Where-Object { -not $_.title.StartsWith("[Ignore]") } |
-        Where-Object { -not $_.title.StartsWith("Update CHANGELOG") } |
-        Where-Object { -not $_.title.StartsWith("Bump version") } |
+        Where-Object { -not $_.title.StartsWith("Release ``v") } |
         Get-Bullets -RepositoryName $RepositoryName
 
     $NewSection = switch ($RepositoryName) {
@@ -242,6 +241,8 @@ function Update-Changelog {
     }
 
     Pop-Location
+
+    Update-Version -RepositoryName $RepositoryName
 }
 
 <#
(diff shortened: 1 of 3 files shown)


andschwa:
@rjmholt I'd like to get the build pipeline stuff using this first. Not sure how to convert back to draft PR.

andschwa:
Ok actually this PR is now ready, doing the rest in a separate PR.

tools/ReleaseTools.psm1
```diff
@@ -344,23 +401,29 @@ function New-DraftRelease {
...
+        Tag            = "v$Version"
+        Committish     = "release/v$Version"
+        Name           = "v$Version"
+        Body           = $ChangeLog
+        Draft          = $true
+        PreRelease     = [bool]$Version.PreReleaseLabel
+        OwnerName      = "PowerShell"
         RepositoryName = $RepositoryName
+        WhatIf         = $WhatIfPreference
+        Confirm        = $ConfirmPreference
```
andschwa, line 420: Actually, @rjmholt this seems to cause `New-GitHubRelease` to _always_ act like `-Confirm` was passed.

rjmholt: I think `-Confirm` is a switch but `$ConfirmPreference` is an enum which never takes a falsey value, so it will behave as you describe.

I'm guessing we don't really want to confirm here, but instead are more interested in the WhatIf functionality

tools/ReleaseTools.psm1
```diff
@@ -344,23 +401,29 @@ function New-DraftRelease {
...
     }
 
     $Release = New-GitHubRelease @ReleaseParams
-    $Assets | New-GitHubReleaseAsset -Release $Release.Id
+    if ($Release) {
+        Write-Output "Draft release URL: $($Release.html_url)"
+        Write-Output "Uploading assets..."
```
rjmholt, line 426: These should probably be either `Write-Host` or `Write-Verbose`

tools/ReleaseTools.psm1
```diff
@@ -322,11 +328,62 @@ function Update-Version {
...
+    Push-Location -Path "$PSScriptRoot/../../$RepositoryName"
+
+    $Version = Get-Version -RepositoryName $RepositoryName
+    $Branch = "release/v$Version"
+
+    Update-Branch -Version $Version
+
+    if ($PSCmdlet.ShouldProcess("$RepositoryName/$Branch", "git push")) {
+        Write-Output "Pushing branch ``$Branch``..."
+        git push origin $Branch
```
rjmholt, line 365: I thought we had a nice invocation checker for this so that if the command fails we can throw

andschwa: Not that I know of in this module.

tools/ReleaseTools.psm1
andschwa: Oh, sure.

tools/ReleaseTools.psm1
andschwa: The build module has something, but I think it's too much for this script. It's no huge deal if it fails, it'll just lead to the PR not being opened and a failure from there.

tools/ReleaseTools.psm1
andschwa: I forgot `[CmdletBinding(SupportsShouldProcess)]`, that was the issue here 😅

rjmholt (approved):

PR #3493, opened 2021-08-03
Update extension icon (and use a special preview icon)

### LINKED ISSUE #3422: Add "Preview" indicator to Preview extension's marketplace icon
Right now it is exactly the same as the stable extension. I just want to take it and plaster the text "PREVIEW" on it 😅

andschwa:
This aligns us with the icons used for PowerShell 7 and sufficiently
differentiates the PowerShell, PowerShell Preview, and Azure PowerShell
extensions on the marketplace.

Resolves #3422.


andschwa:
@rjmholt do you have a way to test this? It seems like since the extensions already exist on the marketplace, at least on my Mac when I build and install the VSIX I see the old icon, but I check the package contents and manifest and they're correct. I also verified and like the Python extension's icon isn't square and is ~600 pixels wide PNG so the icon itself should be fine.

rjmholt:
> @rjmholt do you have a way to test this?

No, but I think it should be ok

rjmholt (approved):

PR #3506, opened 2021-08-11
Update test runner

andschwa:
It got renamed. Also pulled in new versions of the sample runner scripts.


andschwa:
Realized I can add that regression test probably straight to PSES, so opening this for review as just a clean-up and dependency update.

rjmholt (approved):

PR #3549, opened 2021-09-07
Fix `ReleaseTools` issue matcher regex

andschwa:
The simpler "anything but a digit" logic failed on an edge case with a recent PR that had another URL with digits in it after the word "Fix" so we had to make this more explicit.

### DIFF
--- a/tools/ReleaseTools.psm1
+++ b/tools/ReleaseTools.psm1
@@ -108,7 +108,9 @@ function Get-Bullets {
             'resolved'
         )
 
-        $IssueRegex = '(' + ($CloseKeywords -join '|') + ')\s+(?<repo>\D+)(?<number>\d+)'
+        # NOTE: The URL matcher must be explicit because the body of a PR may
+        # contain other URLs with digits (like an image asset).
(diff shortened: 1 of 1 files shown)


andschwa:
I had to work at this a little more to fix it 😅 My first guess was wrong once I started testing with real API data.

tools/ReleaseTools.psm1
```diff
@@ -108,7 +108,9 @@ function Get-Bullets {
             'resolved'
         )
 
-        $IssueRegex = '(' + ($CloseKeywords -join '|') + ')\s+(?<repo>\D+)(?<number>\d+)'
+        # NOTE: The URL matcher must be explicit because the body of a PR may
+        # contain other URLs with digits (like an image asset).
+        $IssueRegex = '(' + ($CloseKeywords -join '|') + ')\s+((https://github.com/PowerShell/(?<repo>(' + ([RepoNames]::Values -join '|') + '))/issues/)|#)(?<number>\d+)'
```
StevenBucher98, line 113: what case does the `|#` account for after the /issues/?

andschwa: The case where a URL was not used an instead just `#1234`. So it's either `the/url/issues/1234` OR `#1234`.

StevenBucher98 (approved):

PR #3554, opened 2021-09-08
WIP: Upgrade `node-fetch`

andschwa:
Just leaving this around so we can take it when VS Code supports ES Modules (https://github.com/microsoft/vscode/issues/130367)


andschwa:
And ES Module support does not look like it'll come any time soon, closing!

PR #3555, opened 2021-09-08
Bundle the extension

### LINKED ISSUE #2450: Bundle the PowerShell extension
See https://code.visualstudio.com/api/working-with-extensions/bundling-extension

andschwa:
This uses `esbuild` as outlined in the [VS Code docs](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#using-esbuild) to bundle the PowerShell extension.

Resolves #2450.


andschwa:
Wow, so this is really annoying. The tooling is absolutely not ready for tests yet.

andschwa:
@rjmholt and @SeeminglyScience please test this (and especially the launch configurations). Everything _seems_ to be working right but it wasn't trivial.

andschwa:
Also filed #3561 as a next step.

tsconfig.json
```diff
@@ -1,16 +1,18 @@
...
-            "DOM"
-        ],
-        "sourceMap": true
-    },
-    "exclude": [
-        "node_modules"
-    ]
+  "compilerOptions": {
+    // NOTE: The TypeScript compiler is only used for building the tests (and
+    // the sources which the tests need). The extension is built with `esbuild`.
```
rjmholt, line 4: Does this file configure the typescript compiler rather than `esbuild`?

rjmholt (approved):

tsconfig.json
andschwa: Yes, T(ype)S(cript)config.json. While `esbuild` _supports_ transpiling typescript to javascript, it _does not_ do any sort of type checking etc., and it also will _only_ produce a bundled file. So we still need `tsc` around to type-check the sources and transpile the test code.

Issue #3586, opened 2021-09-24
deactivate() fails to execute


CrendKing:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.

### Summary

In the exported `deactivate()` function, it [first](https://github.com/PowerShell/vscode-powershell/blob/0ec48b0542618657a062b8a619e7e8158883c756/src/main.ts#L212) calls to all `languageClientConsumers`' `dispose()`. I noticed that two of them are throwing error, and preventing the rest of the `deactivate()` logic from running (e.g. the session manager's `stop()` is never run).

The two offenders are ShowHelp and DebugSession. The common reason of their failure is that, in the ts file a command variable is declared but never assigned. When the ts is compiled to js file, these variables got omitted. Therefore, when the consumer's dispose() tries to dispose these command variables, variable can't be found.

For example, ShowHelpFeature has a [`deprecatedCommand`](https://github.com/PowerShell/vscode-powershell/blob/0ec48b0542618657a062b8a619e7e8158883c756/src/features/ShowHelp.ts#L14). It is accessed [here](https://github.com/PowerShell/vscode-powershell/blob/0ec48b0542618657a062b8a619e7e8158883c756/src/features/ShowHelp.ts#L37). Here is the relevant part of the js file from vsix:
```javascript
...
    constructor(log) {
        super();
        this.log = log;
        this.command = vscode.commands.registerCommand("PowerShell.ShowHelp", (item) => {
            if (!item || !item.Name) {
...
    }
    dispose() {
        this.command.dispose();
        this.deprecatedCommand.dispose();
    }
}
```
The one from DebugSession is [this one](https://github.com/PowerShell/vscode-powershell/blob/0ec48b0542618657a062b8a619e7e8158883c756/src/features/DebugSession.ts#L55).

If I comment out the `dispose()` calls in the installed js file, I can verify from log file that things are back to work (these two lines are currently missing):
```
9/24/2021 6:49:23 AM [NORMAL] - Shutting down language client...
9/24/2021 6:49:23 AM [NORMAL] - Terminating PowerShell process...
```

[...]

PR #3592, opened 2021-09-28
Remove unused feature commands

CrendKing:
## PR Summary

Issue report is https://github.com/PowerShell/vscode-powershell/issues/3586. These two commands are unused except a `dispose()` call (I suppose these are the remnants of some historical refactoring?) In the generated js file, the declarations are gone, but the calls to `dispose()` remain, which causes the extension's `dispose()` to fail.

I suppose alternatively we could tweak how js files are generated and not optimize out unused variables, but that could have way more footprint than removing them. However, if these variables are still used in some way, you guy decide how to proceed.

## PR Checklist

- [x] Summarized changes
- [`NA`] PR has tests

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -25,7 +25,6 @@ export class DebugSessionFeature extends LanguageClientConsumer
     implements DebugConfigurationProvider, vscode.DebugAdapterDescriptorFactory {
 
     private sessionCount: number = 1;
-    private command: vscode.Disposable;
     private tempDebugProcess: PowerShellProcess;
     private tempSessionDetails: utils.IEditorServicesSessionDetails;
 
@@ -49,10 +48,6 @@ export class DebugSessionFeature extends LanguageClientConsumer
         debugAdapter.start();
 
         return new vscode.DebugAdapterInlineImplementation(debugAdapter);
-    }
-
-    public dispose() {
-        this.command.dispose();
     }
 
     public setLanguageClient(languageClient: LanguageClient) {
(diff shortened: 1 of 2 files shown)


src/features/DebugSession.ts
```diff
@@ -51,10 +50,6 @@ export class DebugSessionFeature extends LanguageClientConsumer
         return new vscode.DebugAdapterInlineImplementation(debugAdapter);
     }
 
-    public dispose() {
-        this.command.dispose();
-    }
```
andschwa, line 56: Thanks for the contribution! Looks like due to the abstraction this class is inheriting, we'll need to leave `public dispose()` implemented (but empty):

```
src/features/DebugSession.ts(24,14): error TS2515: Non-abstract class 'DebugSessionFeature' does not implement inherited abstract member 'dispose' from class 'LanguageClientConsumer'.
```

andschwa (approved):
Also please sign the CLA! I'd love to take this PR today. Thanks so much!

src/features/DebugSession.ts
CrendKing: Done

andschwa: Well dang, that triggered:

> ERROR: 53:22  no-empty  block is empty

Are you able to build this locally? TSC should complain in the same way as CI here.

CrendKing: You are right. I could either temporarily disable the tslint or add harmless variable:
```
/* tslint:disable:no-empty */
public dispose() {
}
/* tslint:enable:no-empty */
```
public dispose() {
    const noop = undefined;
}
```

andschwa: Sweet!

CrendKing: Done. CI passed.

andschwa:
Thanks so much @CrendKing!

PR #3593, opened 2021-09-28
Remove `mocha-junit-reporter` package by relying on Mocha's xUnit module

andschwa:
This important because the removed package (`mocha-junit-reporter`) is quite out-of-date and pulls in a vulnerable dependency. Funnily, we're still producing JUnit-style XML test results, which is a very curious bug in Mocha itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,6 @@
         "@vscode/test-electron": "~1.6.2",
         "esbuild": "^0.13.3",
         "mocha": "~9.1.2",
-        "mocha-junit-reporter": "~2.0.0",
         "mocha-multi-reporters": "~1.5.1",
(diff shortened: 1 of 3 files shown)


andschwa:
Unfortunately we also still need to do something about `rewire` which is also woefully three full years from its latest update and pulls in the same vulnerable package.

andschwa:
Ugh the ADO test reporter can't read this xUnit output.

andschwa:
Filed https://github.com/mochajs/mocha/issues/4758 upstream, because I'm confused.

rjmholt (approved):

andschwa:
Looks like rewire's dependency was soft enough that this PR mitigated the issue. 🥳

PR #3643, opened 2021-10-25
Improve test stability with `ensureEditorServicesIsConnected`

andschwa:
This uses a newly added `ExternalApi` function `waitUntilStarted` that ensures the LSP server (PowerShell session) is started after manually activating the extension.


daxian-dbw (approved):
LGTM

PR #3735, opened 2021-12-21
Fix `EnableISEMode` for Azure Data Studio

### LINKED ISSUE #3709: Enable ISE Mode command fails in Azure Data Studio
### Prerequisites

- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.

### Summary

After installing the PowerShell Extension for Azure Data Studio and then attempting to enable the ISE mode an error comes up with  "Command 'PowerShell: Enable ISE Mode' resulted in an error (Running the contributed command: "PowerShell:EnableISEMode' failed.)". This is preventing ISE mode from fully loading and being useable. All that needs to be done to reproduce this with the PowerShell Extension installed is running the PowerShell: Enable ISE Mode. 

image

I would expect this to be similar to Visual Studio Code after launching. The version being reported by **code --list-extensions --show-versions | Select-String powershell** in Azure Data Studio is the same as the latest for Visual Studio Code, but the download reports 2021.8.2.

### PowerShell Version

```console
PSVersion                      7.2.0
GitCommitId                    7.2.0
OS                             Microsoft Windows 10.0.19044
```

```console
1.62.3
ccbaa2d27e38e5afa3e5c21c1c7bef4657064247
x64
```

### Extension Version

```console
ms-vscode.powershell@2021.10.2
```

### Steps to Reproduce

1. Install Azure Data Studio (latest version 1.33.1). 
2. Does` not matter if it is the insider build or not
3. Add the Extension PowerShell v2021.8.2 (current).
4. After adding the PowerShell Extension press F1.
5. Then type >PowerShell: Enable ISE Mode. 

Result:
1. While the ISE mode attempts to start and few of the objects in Azure Data Studio load it does not complete successfully
2. The below error is what is produced.
3. Command 'PowerShell: Enable ISE Mode' resulted in an error (Running the contributed command: "PowerShell:EnableISEMode' failed.)

### Visuals

_No response_

### Logs

_No response_

andschwa:
Because ADS stripped out VS Code's debugger, the `debug` settings don't
exist, causing our `update()` call for the debug setting to throw an
error to the user when they run `EnableISEMode`. I couldn't find a
reliable way to test if the extension is running in the ADS fork instead
of the upstream VS Code, but a simple `try {} catch {}` that allows the
funciton to continue is a reliable fix, as we can't do anything about
this setting regardless.

Resolves #3709.

### DIFF
--- a/src/features/ISECompatibility.ts
+++ b/src/features/ISECompatibility.ts
@@ -42,7 +42,15 @@ export class ISECompatibilityFeature implements vscode.Disposable {
 
     private async EnableISEMode() {
         for (const iseSetting of ISECompatibilityFeature.settings) {
-            await vscode.workspace.getConfiguration(iseSetting.path).update(iseSetting.name, iseSetting.value, true);
+            try {
+                await vscode.workspace.getConfiguration(iseSetting.path).update(iseSetting.name, iseSetting.value, true);
+            } catch {
+                // The `update` call can fail if the setting doesn't exist. This
+                // happens when the extension runs in Azure Data Studio, which
+                // doesn't have a debugger, so the `debug` setting can't be
+                // updated. Unless we catch this exception and allow the
+                // function to continue, it throws an error to the user.
+            }
         }
 
         // Show the PowerShell Command Explorer


Charles-Gagnon:
Thanks for fixing @andschwa ! 

There's a couple options for determining if it's running in ADS : 

1. You can check at runtime whether the azdata API object exists - such as what we do here : https://github.com/microsoft/azuredatastudio/blob/main/extensions/data-workspace/src/common/utils.ts#L54 
1. `vscode.env.appName` is another option (it should be `Azure Data Studio` for ADS) Note though that this may not be the best option as other versions of the app such as insiders may have a slightly different name.

Although unless you find another case that you can't easily solve like this one it's probably easiest to just leave it as this. Although I would suggest adding a log statement or something in the catch so you can see what settings were ignored in case a setting is erroring that you don't expect.

andschwa:
Thanks so much @Charles-Gagnon, you're always so helpful!

Issue #3710, opened 2021-12-02
Allow running/debugging a file in a workspace without adding an entry in launch.json


parched:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Summary

I am new to using the powershell extension, but have previously used the python extension. The python extension adds an editor button to run or debug the current file, ignoring anything from launch.json. See ![the button](https://code.visualstudio.com/assets/docs/python/tutorial/run-python-file-in-terminal-button.png) from [here](https://code.visualstudio.com/docs/python/python-tutorial#_run-hello-world).

I incorrectly assumed that when I pressed the similar looking button that the powershell extension provides on a ps1 file that it would just run that file. Instead, it runs the current configuration that I have selected in the debugger view (which happened to be some python thing completely unrelated to the powershell file open in the editor).

### Proposed Design

My preference would be for the button to work like the python extension, although I imagine that would confuse people already used to be current behaviour. At the very least changing the tooltip from "Run" to something like "Run selected configuration" would have saved me some confusion.

JustinGrote:
@andschwa I don't think this is particular unreasonable, it's pretty easy to "embed" a default launch configuration, we already do it via the pester "Run Tests" codelens.

andschwa:
Yup, I agree. This is something that bugs me too, would love to see it fixed. Not even sure if this is a feature request so much as a bug fix as I think it's an error that the extension just uses the current launch configuration. My guess is the vast majority of use cases for a PowerShell file would be (and are) the default "run this file" configuration.

andschwa:
It took a bit to figure out what exactly was going on here since I was making assumptions based on other erroneous behavior I was seeing. What we're actually looking to do in order to fix this is modify this command:

https://github.com/PowerShell/vscode-powershell/blob/44d8f403feff544d7aa95641223097c0691d4777/src/features/ExtensionCommands.ts#L226-L227

So instead of calling `workbench.action.debug.start` (which leads to the bad behavior you're seeing where it just launches the selected configuration with no regard as to what it is) with a custom command that just runs the equivalent of a `PowerShell: Launch Current File` debug session.

PR #3765, opened 2022-01-15
Migrate folder settings to workspace

andschwa:
Way better!


andschwa:
Next up is this: https://github.com/PowerShell/vscode-powershell/issues/3710

It _really_ bothers me when testing because I'm often launching the extension, then opening a PowerShell file, and not having a "PowerShell Launch Configuration" available. I agree, it really should just work!

SeeminglyScience (approved):

PR #3518, opened 2021-08-23
Update build and task configurations

andschwa:
Small mostly developer oriented fixes.

PR #3772, opened 2022-01-19
Fix `PowerShell.Debug.Start` to just launch current file

### LINKED ISSUE #3710: Allow running/debugging a file in a workspace without adding an entry in launch.json

andschwa:
This command previously used a private API
`workbench.action.debug.start` which led to bad behavior. Namely it
meant that while a PowerShell file was opened, if the triangular "Start"
button was pressed, it would start Code's currently selected launch
configuration which is often not what the user intended.

1. If there was no `launch.json` this worked accidentally in that we
   resolved a default configuration to launch the current file.
2. If a working PowerShell configuration was selected, it may work as
   intended, unless that configuration was to attach.
3. If any other configuration was selected, the user would be left
   bewildered as to why, say, a Python debugger was started for a
   PowerShell file.

Instead we call the public API to start the debugger and give it a copy
of our "Launch Current File" configuration, which is what the user
intended do when clicking the "Start" button on a PowerShell file.

This may introduce some breaking behavior if the user was relying on
this button to start their current correctly configured (and selected)
launch configuration with possible extra customizations. However, in
that the case the user can use Code's built-in options to call the
private API we were calling preivously, namely F5 or the triangular
start button in the debugger workbench (instead of our button).

Fixes #3710, cleans up some files, and reverts a bug introduced in #3518.

### DIFF
--- a/package.json
+++ b/package.json
@@ -442,7 +442,7 @@
               "type": "PowerShell",
               "request": "launch",
               "script": "^\"\\${file}\"",
-              "cwd": "^\"\\${workspaceFolder}\""
+              "cwd": "^\"\\${file}\""
             }
           },
           {
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -75,29 +75,31 @@ export class DebugSessionFeature extends LanguageClientConsumer
         folder: WorkspaceFolder | undefined,
         token?: CancellationToken): Promise<DebugConfiguration[]> {
 
-        const launchCurrentFileId  = 0;
-        const launchScriptId       = 1;
-        const interactiveSessionId = 2;
-        const attachHostProcessId  = 3;
+        enum DebugConfig {
+            LaunchCurrentFile,
+            LaunchScript,
+            InteractiveSession,
+            AttachHostProcess,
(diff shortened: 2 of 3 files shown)


andschwa:
See https://github.com/Microsoft/vscode/issues/59689 for rationale as to why we shouldn't have been calling the formerly used API.

SeeminglyScience (approved):
LGTM!

Issue #3770, opened 2022-01-18
Using a launch configuration with interactive arguments does not work correctly in Powershell preview extension


Hetshu:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.

### Summary

Using a launch configuration like
```json
{
            "name": "Current File w/Args",
            "type": "PowerShell",
            "request": "launch",
            "script": "${file}",
            "args": [
                "${command:SpecifyScriptArgs}"
            ],
            "cwd": "${file}"
},
``` 
and starting the debugger in a powershell script tab with F5 and adding the following arguments in the following popup  
![image](https://user-images.githubusercontent.com/50502257/150013337-b4f0b337-0f91-4b97-aa19-e16d18042f6c.png)

results in the following terminal output
![image](https://user-images.githubusercontent.com/50502257/150013439-7da40277-9d6f-452b-969d-99711c1ea135.png)
if running this with Powershell Preview extension.

This prevents the powershell script to parse the supplied parameters correctly.

### Erroneous  behavior:
![image](https://user-images.githubusercontent.com/50502257/150014854-cb4c7b95-f1bd-466b-acfe-d8c944e564e2.png)

### Expected behavior: (screenshot from working non preview powershell extension)
![image](https://user-images.githubusercontent.com/50502257/150014602-515c796b-4f67-49a9-961f-93810efc7349.png)

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.2.1
PSEdition                      Core
GitCommitId                    7.2.1
OS                             Microsoft Windows 10.0.19043
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.63.2
[...]

PR #3819, opened 2022-02-09
Add magic marker to `specifyScriptArgs` to disable argument escaping

andschwa:
Coupled with a change to PSES, this enables us to workaround VS Code's
API limitations where our interactive "get user's arguments" command can
only return a string, not an array of strings, which is then passed as a
single argument to PSES. Because it likely contains spaces (and multiple
arguments), we need to instruct PSES to not quote the string in this
special case. It would be easier if we could just pass a list of string
arguments, but so it goes.

Coupled with https://github.com/PowerShell/PowerShellEditorServices/pull/1702, this fixes https://github.com/PowerShell/vscode-powershell/issues/3770.


andschwa:
We're going to fix this by reverting the argument escaping.

Issue #3456, opened 2021-07-13
Word Separator configuration for "$" ignored


FriedrichWeinmann:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.

### Summary

After modifying the VSCode settings for `Editor: Word Separators`, removing both `-` and `$` from the default list will have VSCode ignore some of that setting _for the PowerShell extension only_ :

> In PowerShell script file

+ Removing the `-` is honored - doubleclicking on a cmdlet selects the full name
+ Removing the `$` is _ignored_ - doubleclicking on a variable does not select the $ of the variable

> In a Markdown file

+ Removing the `-` is honored - doubleclicking on a cmdlet selects the full name
+ Removing the `$` is honored - doubleclicking on a variable selects the full name

> Configuration value

The full current value of the `Editor: Word Separators` setting is:

```text
`~!@#%^&*()=+[{]}\|;:'",.<>/?
```

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.1.0-preview.6
PSEdition                      Core
GitCommitId                    7.1.0-preview.6
OS                             Microsoft Windows 10.0.19043
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.58.0
2d23c42a936db1c7b3b06f918cde29561cc47cd6
x64
```

### Extension Version

```console PS> code --list-extensions --show-versions | Select-String powershell
ms-vscode.powershell@2021.5.1
ms-vscode.powershell@2021.6.2
ms-vscode.powershell-preview@2021.6.1
```

### Steps to Reproduce

Set Word Separators to:

```text
`~!@#%^&*()=+[{]}\|;:'",.<>/?
```

Then doubleclick on a variable in a PowerShell script

andschwa:
[...]

Issue #3404, opened 2021-06-08
Snippets doesn't follow the new word separator standard


Schillman:
System Details Output

```
### VSCode version: 1.56.2 054a9295330880ed74ceaedda236253b4f39a335 x64

### VSCode extensions:
antfu.icons-carbon@0.2.4
eamodio.gitlens@11.4.1
formulahendry.code-runner@0.11.4
hashicorp.terraform@2.11.0
mohsen1.prettify-json@0.0.3
ms-python.python@2021.5.842923320
ms-toolsai.jupyter@2021.6.832593372
ms-vscode-remote.remote-ssh@0.65.4
ms-vscode-remote.remote-ssh-edit@0.65.6
ms-vscode-remote.vscode-remote-extensionpack@0.21.0
ms-vscode.azure-account@0.9.8
ms-vscode.azurecli@0.5.0
ms-vscode.powershell-preview@2021.5.0
ms-vsliveshare.vsliveshare@1.0.4360
msjsdiag.debugger-for-chrome@4.12.12
msjsdiag.debugger-for-edge@1.0.15
PKief.material-icon-theme@4.7.0
quicktype.quicktype@12.0.46
rangav.vscode-thunder-client@1.4.0
zhuangtongfa.material-theme@3.10.14


### PSES version: 2.4.2.0

### PowerShell version:

Name                           Value                                                                                                          
----                           -----                                                                                                          
PSVersion                      5.1.19041.1023                                                                                                 
PSEdition                      Desktop                                                                                                        
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                        
BuildVersion                   10.0.19041.1023                                                                                                
CLRVersion                     4.0.30319.42000                                                                                                
WSManStackVersion              3.0                                                                                                            
PSRemotingProtocolVersion      2.3                                                                                                            
SerializationVersion           1.1.0.1                                                                                                        

```

### Issue Description

The Snippet behaviour still matches the old separators. It doesn't select the $ to be replaced when inserting a snippet.
![image](https://user-images.githubusercontent.com/19289252/121164229-bcc60e00-c84f-11eb-897c-f4a297728ca5.png)

#### Expected Behaviour
I would expect it to match the behaviour of the word separators by default for PowerShell files.

#### Actual Behaviour

When inserting a snippet, it should not only select the word in the variables, but the entire variable, including '$' as that follows the new standard.

andschwa:
[...]

PR #3839, opened 2022-02-19
Modernize built-in snippets

JustinGrote:
Evaluate and modernize existing snippets. This will be a breaking change in some cases, we should provide guidance for restoring existing snippets for any cranky people :).

After this  PR, snippets may be further improved by moving them out of JSON definition and into typescript, the primary improvement being that we can define the description with a MarkdownString and include help links, etc. but how much of that should just be returned by the LSP vs. a snippet is TBD.

- [ ] ~~ArgumentCompleterAttribute ScriptBlock~~ - Redundant and not useful since you cant use a variable with an ArgumentCompleter
- [x] ArgumentCompleterAttribute ~~with ScriptBlock~~ - Renamed and more descriptive help added. Added final tabstop after argumentcompleter
- [x] CalculatedProperty - description and  help reference
- [x] Class - description and help reference
- [x] comment block
- [ ] ~~Comment-Help~~ Replaced with `Function Help`
- [x] Constructor - description and help reference
- [x] do-until - description and help reference
- [x] do-while - description and help reference
---
## Should DSC move to community snippets?
- [ ] DSC Ensure Enum
- [ ] DSC Resource Provider (class-based)
- [ ] DSC Resource Provider (function-based)
---
- [x] else - description and help reference
- [x] elseif  - description and help reference
- [x] Enum - description and help reference
- [ ] ~~Example-Advanced Workflow~~ - Workflows are deprecated, we should encourage parallel instead.
---
## Should we still have examples or should the examples be with placeholders?
- [ ] Example-Class
- [ ] Example-Cmdlet
- [ ] Example-DSC Configuration
- [ ] Example-DSC Resource Provider (class-based)
- [ ] Example-DSC Resource Provider (function based)
- [ ] Example-Path Processing for No Wildcards Allowed
- [ ] Example-Path Processing for Non-Existing Paths
- [ ] Example-Path Processing for Wildcards Allowed
- [ ] Example-Splatting
- [ ] Example-Switch
---

- [x] for - description and help reference,
- [ ] for-reversed
- [x] foreach
- [x] function
- [x] Function-Advanced- description and help reference, update aliases
- [x] Function-Help- description and help reference, update aliases
- [x] Function-Inline- description and help reference, update aliases
- [x] Hashtable- description and help reference, update aliases
- [x] Hidden Property- description and help reference, update aliases
- [x] IArgumentCompleter Class - description and help reference, update aliases
- [x] if - description and help reference,
- [x] IfShouldProcess  - description and help reference, update aliases
- [ ] Method
- [ ] ModuleManifest
- [x] Parameter  - description and help reference, update aliases
- [x] Parameter_Block  - description and help reference, update aliases
- [x] Parameter-LiteralPath  - description and help reference, update aliases
- [x] Parameter-Path  - description and help reference, update aliases
- [x] Parameter-Path-Wildcards  - description and help reference, update aliases
---
## Should Pester Move to Community Snippets since it is no longer bundled with Powershell?
- [ ] PesterContext
- [ ] PesterContextIt
- [ ] PesterDescribeBlock
- [ ] PesterDescribeContextIt
- [ ] PesterIt
---

- [ ] Property
- [ ] PSCustomObject
- [ ] Region Block
---
## Intellisense now handles the requires, there was a comment that these should be removed when it does.
- [ ] ~~Requires Assembly~~
- [ ] ~~Requires Assembly Path~~
- [ ] ~~Requires Assembly Version~~
- [ ] ~~Requires Module~~
- [ ] ~~Requires Module RequiredVersion~~
- [ ] ~~Requires Module Version~~
- [ ] ~~Requires PSEdition~~
- [ ] ~~Requires PSSnapin~~
- [ ] ~~Requires PSSnapin Version~~
- [ ] ~~Requires RunAsAdministrator~~
- [ ] ~~Requires ShellId~~
- [ ] ~~Requires Version~~
---
- [x] Suppress PSScriptAnalyzer Rule - rewritten to multiline, description, and help reference
- [x] Suppress PSScriptAnalyzer Rule in Scope - rewritten to multiline, description, and help reference
- [x] Suppress PSScriptAnalyzer Rule on Parameter - rewritten to multiline, description, and help reference
- [x] switch  - description and help reference
- [x] try-catch - description and help reference
- [x] try-catch-finally - description and help reference
- [x] try-finally - description and help reference
- [x] while - description and help reference
---
## Workflows are deprecated and not supported in 6+, we should not be encouraging their use.
- [ ] ~~Workflow~~ 
- [ ] ~~Workflow ForEachParallel~~
- [ ] ~~Workflow InlineScript~~
- [ ] ~~Workflow Parallel~~
- [ ] ~~Workflow Sequence~~
---
# NEW
* Function Help - Better Comment based help with tabstopes
* Pipeline Function - A simple definition of a function that is pipelineable
* Parallel Pipeline Function - A simple definition of a function that uses foreach-parallel to process items in the pipeline in parallel.
* Ternary - build a ternary condition
* Splat - Select a function name and it will build a splat for it. This is *not* a refactor for an existing function.
* Here String
* Literal Here String


JustinGrote:
@andschwa first pass done. There's a lot of things here that I think are more community questions: Should we retain legacy snippets for deprecated and discouraged features like workflows, should we retain pester now that it's no longer bundled with PS, etc.

JustinGrote:
@andschwa Build failure is not me:
```
ERROR: The script 'PowerShellEditorServices.build.ps1' cannot be run because the following modules that are specified by the "#requires" statements of the script are missing: platyPS.
```

andschwa:
> @andschwa Build failure is not me:
> 
> ```
> ERROR: The script 'PowerShellEditorServices.build.ps1' cannot be run because the following modules that are specified by the "#requires" statements of the script are missing: platyPS.
> ```

Fixed it! Sorry about that. 😅 Give it a rebase and it should work. This is looking great!

andschwa:
I rebased it!

JustinGrote:
Should be good to go. A lot of these snippets support the vscode "snippet in place" method now. For example, type `Invoke-Restmethod`, highlight it, and then overwrite it with `splat` and hit tab or select the completion snippet.

andschwa:
@JustinGrote do you know what this issue means: https://github.com/PowerShell/vscode-powershell/issues/3404 ? 😂

JustinGrote:
@andschwa it's a symptom of an issue closed for workaround that we may want to reopen
https://github.com/PowerShell/vscode-powershell/issues/3456

JustinGrote:
Disregard, I was mistaken, this is part of the snippet itself. It can be configured either way, but we should decide if we are going to change the consensus on whether $ is considered a word separator or not (e.g. double clicking $this if it highlights the whole thing or just "this"), whatever that decision, the snippets can follow suit. As of today I have them matching the existing behavior ($ is excluded), a user can always provide their own overriding user snippets for the other mode.

Ultimately I'd like to see the snippets converted to completions for multiple reasons:
1.  Markdown descriptions can be used for better help (e.g. links)
2. This behavior could be configurable or could auto-detect wordseparators

JustinGrote:
Also I personally am team "$ is not a word separator" mostly because it makes double-click + ctrl-f2 to rename all symbols more error prone, unless we come up with a way to do a proper rename symbol in Powershell

andschwa:
> Should DSC move to community snippets?

Eh, sure, if we have that setup.

> Should we still have examples or should the examples be with placeholders?

I'm down to remove them. They could be in a help file if we wanted.

> Should Pester Move to Community Snippets since it is no longer bundled with Powershell?

You'd be the expert on this! Perhaps they should? But also everyone uses Pester with PowerShell, it's the default test framework really.

> Intellisense now handles the requires, there was a comment that these should be removed when it does.

Sweet!

> Workflows are deprecated and not supported in 6+, we should not be encouraging their use.

Sounds good to me!

> NEW

Loving it!

P.S. Sorry this took me a while to get to! Testing them out right now.

andschwa (approved):

andschwa:
Thanks so much @JustinGrote!

Issue #3820, opened 2022-02-10
`Get-Credential` throws exception instead of prompting for input in PowerShell 5.1


csandfeld:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.

### Summary

When using preview v2022.2.0 with Windows PowerShell 5.1, this exception is throw in the integrated console when running Get-Credential.
```PowerShell
PS C:\> Get-Credential
Get-Credential : Cannot process command because of one or more missing mandatory parameters: Credential.
At line:1 char:1
+ Get-Credential
+ ~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-Credential], ParameterBindingException
    + FullyQualifiedErrorId : MissingMandatoryParameter,Microsoft.PowerShell.Commands.GetCredentialCommand
 
PS C:\> 
```

This is not the case when using PowerShell Extension v2021.12.0, or when using PowerShell 7.2.1 instead of Windows PowerShell.

### PowerShell Version

```console
PS C:\> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.19041.1320
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.19041.1320
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

### Visual Studio Code Version

```console
PS C:\> code --version
1.64.1
d6ee99e4c045a6716e5c653d7da8e9ae6f5a8b03
x64
PS C:\>
```

### Extension Version

```console
PS C:\> code --list-extensions --show-versions                           
ms-vscode.powershell@2021.12.0
ms-vscode.powershell-preview@2022.2.0
PS C:\>
```

### Steps to Reproduce

- Open Visual Studio Code
[...]

PR #3847, opened 2022-02-22
Remove `-NonInteractive` from PowerShell startup sequence

andschwa:
This command-line flag sets a low-level configuration within PowerShell
that prevents Windows PowerShell from using interactive parameters in
scripts (such as in `Get-Credential`). It does not appear to be
necessary for us to use this flag, so we're going to test without it.

Fixes https://github.com/PowerShell/vscode-powershell/issues/3820.

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -73,7 +73,6 @@ export class PowerShellProcess {
         }
 
         powerShellArgs.push("-NoProfile");
-        powerShellArgs.push("-NonInteractive");
 
         // Only add ExecutionPolicy param on Windows
         if (utils.isWindows) {


andschwa:
@dependabot rebase

andschwa:
@SeeminglyScience On PowerShell 5.1 and 7.2.1, the previous bug where the wrong PSReadLine was loaded does not appear to be recurring with this PR. They both loaded the module from the extension path, despite other versions of PSReadLine being available (and loaded in the standalone console).

andschwa:
Screen Shot 2022-02-22 at 3 47 20 PM
Seems to solve the problem!

SeeminglyScience (approved):
Ahh that's right. `-Command`/`-File` should do the same as long as `-NoExit` isn't present. LGTM!

PR #3853, opened 2022-02-28
Update `engines.vscode` and run `npm update`

andschwa:
While Dependabot is great, it's not quite a replacement for periodically letting `npm` update to our semver specifications.


SeeminglyScience (approved):
LGTM!

PR #3854, opened 2022-02-28
Use new `isTransient` API to prevent duplicate integrated consoles

andschwa:
This new API available today in VS Code 1.65.0 prevents Code from saving
and reloading the integrated console when the extension is reloaded.
This is necessary because the extension always spawns a new terminal to
host PowerShell Editor Services when it starts, and Code's newish
feature to save and restore terminals caused these to be duplicated.
When we update the VS Code engine, we can remove the cast to `any`.

Fixes https://github.com/PowerShell/vscode-powershell/issues/3360.

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -2,9 +2,7 @@
 // Licensed under the MIT License.
 
 import cp = require("child_process");
-import fs = require("fs");
-import net = require("net");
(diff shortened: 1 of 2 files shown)


SeeminglyScience (approved):
LGTM!

PR #3865, opened 2022-03-04
Replace custom graphics with updated Codicons

andschwa:
These look more like the rest of Code since they're built-in.


andschwa:
Some of the icons are a bit newer (particularly `layout-*` that just came out in 1.65.0), but _theoretically_ everyone should have up-to-date VS Code, and the worst manifestation of this is just missing icons. Going to manually test that assumption though.

## Before:
Screen Shot 2022-03-04 at 2 26 28 PM
Screen Shot 2022-03-04 at 2 26 43 PM

## After:
Screen Shot 2022-03-04 at 2 25 07 PM
Screen Shot 2022-03-04 at 2 25 21 PM

package.json
```diff
@@ -93,8 +93,8 @@
       "activitybar": [
         {
           "id": "PowerShellCommandExplorer",
-          "title": "(Preview) PowerShell Command Explorer",
-          "icon": "media/pwsh.svg"
+          "title": "PowerShell Command Explorer",
+          "icon": "$(terminal-powershell)"
```
andschwa, line 97: Since 1.56!

package.json
```diff
@@ -155,19 +155,13 @@
       {
         "command": "PowerShell.RefreshCommandsExplorer",
         "title": "Refresh Command Explorer",
-        "icon": {
-          "light": "media/resources/light/sync.svg",
-          "dark": "media/resources/dark/sync.svg"
-        },
+        "icon": "$(sync)",
```
andschwa, line 158: Since <1.56

package.json
```diff
@@ -155,19 +155,13 @@
...
         "category": "PowerShell"
       },
       {
         "command": "PowerShell.InsertCommand",
         "title": "Insert Command",
-        "icon": {
-          "light": "media/resources/light/pencil.svg",
-          "dark": "media/resources/dark/pencil.svg"
-        },
+        "icon": "$(pencil)",
```
andschwa, line 164: Since <1.56

package.json
```diff
@@ -178,29 +172,20 @@
       {
         "command": "PowerShell.ShowHelp",
         "title": "Get Help for Command",
-        "icon": {
-          "light": "media/resources/light/question.svg",
-          "dark": "media/resources/dark/question.svg"
-        },
+        "icon": "$(question)",
```
andschwa, line 175: Since <1.56

package.json
```diff
@@ -178,29 +172,20 @@
...
       },
       {
         "command": "PowerShell.Debug.Start",
         "title": "Run",
         "category": "PowerShell",
-        "icon": {
-          "light": "media/resources/light/run.svg",
-          "dark": "media/resources/dark/run.svg"
-        }
+        "icon": "$(run-all)"
```
andschwa, line 182: Since <1.56. Could also just use `$(run)` (the single arrow, used next) and instead use `$(debug-line-by-line)` for "Run Selection" which would look like this:
Screen Shot 2022-03-04 at 2 42 07 PM

package.json
```diff
@@ -276,28 +261,19 @@
         "command": "PowerShell.ClosePanel",
         "title": "Close panel",
         "category": "PowerShell",
-        "icon": {
-          "light": "media/resources/light/ClosePanel.svg",
-          "dark": "media/resources/dark/ClosePanel.svg"
-        }
+        "icon": "$(layout-activitybar-left)"
```
andschwa, line 264: This and the next two are new and in 1.65.0. They're only seen in the ISE-mode. We could delay changing them until then. I guess Azure Data Studio users that use ISE-mode wouldn't see these, it'd look like this:
Screen Shot 2022-03-04 at 2 40 05 PM

Ok ok, we'll stash this part of the change for later.

package.json
andschwa: This is the current iteration. It looks more like what we had.

SydneyhSmith:
These changes look good to me!

SydneyhSmith (approved):

PR #3866, opened 2022-03-05
Fix title of PowerShell Command Explorer pane

andschwa:
It's no longer in preview, and the title was repetitive.

This _probably_ won't break anyone (unless they bound a keyboard shortcut to this particular view, which both seems unlikely and I didn't see anything on a search of GitHub indicating anyone was doing this).

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,17 +92,17 @@
     "viewsContainers": {
       "activitybar": [
         {
-          "id": "PowerShellCommandExplorer",
-          "title": "(Preview) PowerShell Command Explorer",
+          "id": "PowerShell",
+          "title": "PowerShell",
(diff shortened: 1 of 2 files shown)


andschwa:
It was poorly named and it made things crowded in the UI. Technically we could change the title without changing the name but then they'd be confusingly out of sync, and it's better to do it sooner than later (especially since the title has included "Preview" in it).

### Before:
Screen Shot 2022-03-04 at 4 33 24 PM

### After:

Screen Shot 2022-03-04 at 4 34 49 PM

SeeminglyScience (approved):
Yeah I think that's safe. At the very least it'll be easy to fix for the small handful who might have something set up.

LGTM

PR #3869, opened 2022-03-10
Use Code's new Language Status API

### LINKED ISSUE #2499: Use new language status items API for status bar info
PowerShell uses the very right of the right status bar to show both initializing and version number:

image

image

This is bad since it makes the status bar unstable; all items left of the PS item will shift to the left and again to the right. I suggest aligning with the TS extension and show the initializing item on the right side of the left status bar:

image

image

Also getting closer to the wording of the TS extension might be a good idea; "Initializing PS language features" or "Initializing PS session" (no "...")?

Related: https://github.com/PowerShell/vscode-powershell/issues/2436

andschwa:
Also use `@ts-ignore` instead of `any` workaround, and replace old hack with new API. Much cleaner!

Fixes #2499.

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -110,9 +110,9 @@ export class PowerShellProcess {
             cwd: this.sessionSettings.cwd,
         };
 
-        // This API is available only in newer versions of VS Code.
         if (semver.gte(vscode.version, "1.65.0")) {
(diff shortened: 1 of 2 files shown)


SydneyhSmith (approved):
LGTM

PR #3897, opened 2022-03-29
Add PowerShell icon to Integrated Console

andschwa:
Saw this option in the terminal creation options struct, tested it out, quite nice!

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -106,8 +106,9 @@ export class PowerShellProcess {
             name: this.title,
             shellPath: this.exePath,
             shellArgs: powerShellArgs,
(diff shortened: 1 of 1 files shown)


JustinGrote:
Only issue I potentially have is that I used the new icon in my terminal profiles and made it easy to distinguish the PSIC from the "regular" terminals.

andschwa:
And I think that's actually a great reason _to_ do this, since the default `pwsh` terminals opened by Code do not use the PowerShell icon. We can do so, and differentiate the PowerShell extension's integrated terminal from the others visually. (The opposite of what you'd done manually.)

SeeminglyScience (approved):
Love it 🎉

PR #3900, opened 2022-03-31
Fix snippet prefixes (especially `#region`)

### LINKED ISSUE #3276: Typing `<-#-<RET>` results in `<#endregion#>` instead of `<#\n`
We have a snippet for `#endregion#` that seems to be blocking me from just opening PowerShell's comment block syntax like:

```powershell
<#
  .SYNOPSIS
...
#>
```

Because hitting enter after `<#` accepts the suggestion to expand that completion/snippet instead of entering a newline.

@rjmholt any ideas?

andschwa:
The snippets' prefixes should all be consistently lower-cased, and
moreover, the region/endregion snippet must not start with a hashtag,
because this triggers the completion when users are trying to type
comments. VS Code also has two built-in snippets with this issue, that
should be removed upstream.

Coupled with https://github.com/microsoft/vscode/pull/146459, this fixes #3276.

### DIFF
--- a/snippets/PowerShell.json
+++ b/snippets/PowerShell.json
@@ -234,7 +234,7 @@
     ]
   },
   "Function-Inline": {
-    "prefix": "Function-Inline",
+    "prefix": "function-inline",
     "description": "Function definition snippet that does not contain a param block, but defines parameters inline. This syntax is commonly used in other languages. More: Get-Help about_Functions",
     "body": [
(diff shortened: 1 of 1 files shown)


JustinGrote:
LGTM, probably mostly muscle memory on my part, thanks!

JustinGrote (approved):

PR #3939, opened 2022-04-26
Send `p` instead of `\0` for `sendKeyPress`

andschwa:
This sends a less-risky character in case PSReadLine does _not_ consume it (due to a race condition). The `p` can be deleted and does ring the console bell.


SeeminglyScience (approved):
LGTM!

Issue #3950, opened 2022-05-04
Temporary integrated consoles stuck waiting for key press


ohaak2:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/master/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/master/SECURITY.md).

### Summary

With vscode version 1.66.2 and PowerShell extension v2022.5.0 I'm suddenly unable to start any PowerShell session, if  the powershell.debugging.createTemporaryIntegratedConsole parameter is set to 'true'.

PowerShell session isn't even started in run mode. It's just hanging somewhere before actually processing any custom code. No code will be executed, no breakpoint reached. I can reproduce the isse even with a single file containing 'hello world' output.

Switching this setting to 'false' and everything is fine.

### PowerShell Version

```console
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.2.3
PSEdition                      Core
GitCommitId                    7.2.3
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
PS > code --version

1.66.2
dfd34e8260c270da74b5c2d86d61aee4b6d56977
x64
```

### Extension Version

```console
PS > code --list-extensions --show-versions | Select-String powershell

ms-vscode.powershell@2022.5.0
```

### Steps to Reproduce

- Open PowerShell file in vscode.
- set powershell.debugging.createTemporaryIntegratedConsole configuration parameter to 'true'
- try to start run or debug session.

### Logs

DEBUG: Logging started
DEBUG: Beginning EndProcessing block
VERBOSE: Removing PSReadLine
[...]

PR #3951, opened 2022-05-05
Handle `sendKeyPress` events for temporary integrated consoles

andschwa:
The first commit removes our custom named pipe debug adaptor. It was used primarily for logging, but until we actually need to log at these logs, let's reduce code complexity and rely on VS Code's built-in implementation.

The second commit registers a handler to send a key press to the spawned temporary integrated console. Note that the DAP specification is slightly different from the LSP specification. In https://github.com/PowerShell/PowerShellEditorServices/pull/1791, we use the same `sendNotification`, which comes across DAP as a debug session custom event (not a notification).

Note that we seem to only support one temporary integrated console / process at a time, though I don't see anything explicitly calling that out.

Fixes https://github.com/PowerShell/vscode-powershell/issues/3950.


SeeminglyScience (approved):
LGTM!

PR #3953, opened 2022-05-05
Fix disposal of temporary console and event handler

andschwa:
Fixes a bug where the first patch would register a new event handler every time, leading to multiple 'p's being sent. We currently only support one process at a time, and so this patch consolidates the process's and the event handler's disposal.


SeeminglyScience (approved):
LGTM!

PR #3992, opened 2022-05-19
Fix how we check extension mode

andschwa:
Well, this was dumb, and was the reason for red-herring logs in user reports that their extension was in development mode, and why our telemetry had stopped coming through.


SeeminglyScience (approved):
LGTM!

SeeminglyScience (approved):
LGTM!

PR #4071, opened 2022-07-11
Use `context.storageUri` for logs and support `None` level

andschwa:
Resolves #4067 (and uses the _correct_ location for logs, as reported via Discord, our extension was defaulting to the extension folder which is not where they be stuck).

PR #4088, opened 2022-07-20
Use `context.storageUri` for session file (and refactor)

andschwa:
This is a decent cleanup of our session file logic in the client, so it needs a close review. Most important note is that I completely removed `writeSessionFile`...and everything works as expected in my testing. Which I think makes sense, as we pass the session file path to the server and expect it to write the session file. I don't know why we had the client ever writing a session file. Furthermore, the only reference I can find to it in the server is with a writer, no reader. 🤷

Related to https://github.com/PowerShell/vscode-powershell/pull/4071.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -16,13 +16,15 @@ import utils = require("./utils");
 import {
     CloseAction, DocumentSelector, ErrorAction, LanguageClientOptions,
     Middleware, NotificationType, RequestType0,
-    ResolveCodeLensSignature, RevealOutputChannelOn } from "vscode-languageclient";
+    ResolveCodeLensSignature, RevealOutputChannelOn
+} from "vscode-languageclient";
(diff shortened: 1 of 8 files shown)


src/session.ts
```diff
@@ -34,6 +36,20 @@ export enum SessionStatus {
...
+    detail: string;
+    powerShellVersion: string;
+    channel: string;
+    languageServicePort: number;
+    debugServicePort: number;
+    languageServicePipeName: string;
+    debugServicePipeName: string;
+}
+
+export type IReadSessionFileCallback = (details: IEditorServicesSessionDetails) => void;
```
andschwa, line 51: I can't find what this was used for...delete it?

SeeminglyScience (approved):
LGTM!

I think originally it used to write the session file first and wait for it to be populated... something like that? Can't remember if it was a strange limitation of the time or just accidental over engineering but either way can probably just be dropped 😁

Issue #2523, opened 2020-03-05
When showOnStartup is off, integrated console only starts during debugging


ThubLives:
Issue Description
=====

If powershell.integratedConsole.showOnStartup is set to false, the workbench.action.terminal.toggleTerminal command starts the default terminal instead of the integrated PowerShell terminal.  In version 2020.1.0 and earlier, the integrated PowerShell terminal would be started when using toggleTerminal, rather than the default terminal.

The only obvious way to start the integrated terminal is to start debugging (F5) or close any open terminals and use F8 to run a line of code from the editor.

Attached Logs
=====

[vscode-PS-2020.3.0-logs.zip](https://github.com/PowerShell/vscode-powershell/files/4294823/vscode-PS-2020.3.0-logs.zip)

Environment Information
=====

Visual Studio Code
-----

| Name | Version |
| --- | --- |
| Operating System | Windows_NT x64 10.0.17134 |
| VSCode | 1.42.1|
| PowerShell Extension Version | 2020.3.0 |

PowerShell Information
-----

|Name|Value|
|---|---|
|PSVersion|5.1.17134.858|
|PSEdition|Desktop|
|PSCompatibleVersions|1.0 2.0 3.0 4.0 5.0 5.1.17134.858|
|BuildVersion|10.0.17134.858|
|CLRVersion|4.0.30319.42000|
|WSManStackVersion|3.0|
|PSRemotingProtocolVersion|2.3|
|SerializationVersion|1.1.0.1|

Visual Studio Code Extensions
-----

<details><summary>Visual Studio Code Extensions(Click to Expand)</summary>

|Extension|Author|Version|
|---|---|---|
|powershell|ms-vscode|2020.3.0|

rjmholt:
Yes that was https://github.com/PowerShell/vscode-powershell/pull/2437, which fixed https://github.com/PowerShell/vscode-powershell/issues/2434.

It sounds like we want to break this out into a configuration rather than bake the setting in.

ThubLives:
I would suggest replacing the `powershell.integratedConsole.showOnStartup` config option with a tri-state option something like `powershell.integratedConsole.showConsole` with states:  `OnStartup`, `AsDefault`, and `OnCodeExecution`.

ThubLives:
[...]

PR #4096, opened 2022-07-28
Don't hide extension terminal entirely

andschwa:
When `showOnStartup` is false, we currently completely hide the terminal from the user via an API that essentially backgrounds the terminal. This was in response to a user request, but has caused more problems than it's worth. This PR reverts the behavior so that it simply won't show the terminal pane on startup, but opening it reveals the connected extension terminal.

This resolves https://github.com/PowerShell/vscode-powershell/issues/2523 and https://github.com/PowerShell/vscode-powershell/issues/3918, and reverts https://github.com/PowerShell/vscode-powershell/pull/2437 which was a response to https://github.com/PowerShell/vscode-powershell/issues/2434. However, we've decided that behavior is undesirable.

Future work: we're going to flight switching the default of `showOnStartup`, and add to the upcoming PowerShell Walkthrough a step which shows how to re-enable it.

### DIFF
--- a/src/process.ts
+++ b/src/process.ts
@@ -109,7 +109,6 @@ export class PowerShellProcess {
             shellPath: this.exePath,
             shellArgs: powerShellArgs,
             cwd: this.sessionSettings.cwd,
-            hideFromUser: !this.sessionSettings.integratedConsole.showOnStartup,
             iconPath: new vscode.ThemeIcon("terminal-powershell"),
         };
 


SeeminglyScience:
Maybe a third option?

There's definitely some folks who never want to see it and often spam close terminals instead of hiding the pane. No idea *how many*, maybe it'll just be a vocal few. But I think changing the way this setting currently works will result in another surge or false positive "it doesn't work sometimes" issues.

Mostly just wanted to voice that concern, I'm cool with trying it out in a preview if you still want to.

SeeminglyScience:
> Future work: we're going to flight switching the default of `showOnStartup`, and add to the upcoming PowerShell Walkthrough a step which shows how to re-enable it.

We *want* folks to use the PSIC though. If they don't, `F8` stops working properly, intellisense doesn't reflect the context of the terminal they're in, when they start debugging they will all of a sudden be in a different context.

That default makes sense if you're only occasionally using PowerShell but if it's your primary workload that's going to be detrimental.

andschwa:
@SeeminglyScience I'll chat with you virtually, I think there's some crossed wires here.

andschwa:
Sounds like maybe we'll add a "Background Process" option.

SeeminglyScience (approved):
Hittin' approve in case you wanna trial run this in a preview before we make a "as background process" setti

andschwa:
Sounds good, let's trial it.

wsmelton:
> We want folks to use the PSIC though. If they don't, F8 stops working properly, intellisense doesn't reflect the context of the terminal they're in, when they start debugging they will all of a sudden be in a different context.

I know this is a bit dated but I'm confused on this area as I intentionally use Code terminal over integrated just for stability and "it runs the same on my machine". Since it more closes mimics running scripts outside of VS Code. I also had issues with integrated using dbatools module and developing in large workspaces like dbatools.

Either way, F8 and Intellisense worked fine for me using Code terminal when I have that setting set to false. I import a module into the terminal and had full Intellisense of that module with no issues.

SeeminglyScience:
> I know this is a bit dated but I'm confused on this area as I intentionally use Code terminal over integrated just for stability and "it runs the same on my machine". Since it more closely mimics running scripts outside of VS Code. I also had issues with integrated using dbatools module and developing in large workspaces like dbatools.

A large focus of the extension is to keep the console working the same as an external one. Have you reported the issues you encountered?

> Either way, F8 and Intellisense worked fine for me using Code terminal when I have that setting set to false. I import a module into the terminal and had full Intellisense of that module with no issues.

If the module is in your `PSModulePath` then intellisense will work, but it's not related to the code you executed with <kbd>F8</kbd>. We get intellisense only from the Extension terminal, any others will have no effect on intellisense (aside from things like *installing* a module ofc).

wsmelton:
> A large focus of the extension is to keep the console working the same as an external one. Have you reported the issues you encountered?

Most of it was known issues way back when, when you loaded modules that had a high number of DLL libraries. This was well before all the rewrite that Andy started working on. If I come across it on latest release I will be sure to submit issues.

Issue #4098, opened 2022-07-29
Multi-root workspace prompts twice for `cwd` selection


andschwa:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

Thought I'd already fixed this, but sadly it made into stable. If you _use_ multi-root workspaces with the PowerShell extension, please react to this issue so I know how to prioritize this and the remaining work for full support of it!

### PowerShell Version

```console
7.2.5
```

### Visual Studio Code Version

```console
1.69.2
```

### Extension Version

```console
v2022.7.2
```

### Steps to Reproduce

Open a multi-root workspace without `cwd` set. Choose which folder in the prompt, then sadly it prompts again.

andschwa:
Other work: https://github.com/PowerShell/vscode-powershell/pull/3796 and the issue https://github.com/PowerShell/vscode-powershell/issues/2153.

PR #4099, opened 2022-07-29
Fix `checkIfDirectoryExists()` so `validateCwdSetting()` works

andschwa:
Oops, Node APIs are weird. Moved to VS Code APIs.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4098.


SeeminglyScience (approved):
LGTM!

Issue #4102, opened 2022-08-02
Extension doesn't start with when workspace is defined with zero folders


opmdsm:
### Prerequisites

- [X] I have written a descriptive issue title.
- [ ] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [ ] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [ ] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [ ] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

Hi,

Just updated PowerShell extension to v2022.7.2 and all powershell files within an existing workspace does not work anymore.

For instance:
1. Switching from PowerShell x86 and x64 does not work anymore (command PowerShell: ShowSessionMenu not found).
2. Executing only a selection of commands also does not work (command PowerShell: RunSelection not found).

If I open those files separately (open file, without open workspace) the above commands runs successfully.

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.2.5
PSEdition                      Core
GitCommitId                    7.2.5
OS                             Microsoft Windows 10.0.19044
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.69.2
3b889b090b5ad5793f524b5d1d39fda662b96a2a
x64
```

### Extension Version

```console
ms-vscode.powershell@2022.7.2
```

### Steps to Reproduce

Open an existing workspace and try to execute what's described in the issue.

### Visuals

Commands executed within an opened workspace:

Command PowerShell: ShowSessionMenu not found:
![image](https://user-images.githubusercontent.com/107676477/182338130-dc7e158c-334f-4609-8eed-169471bec03a.png)

PowerShell: RunSelection not found:
[...]

PR #4104, opened 2022-08-03
Fix edge case for workspaces defined with zero folders

andschwa:
Wasn't catching the case where `workspaceFolders` itself is defined, but with a length of zero. Just didn't expect this to happen.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4102.


SeeminglyScience (approved):
LGTM!

PR #4117, opened 2022-08-09
Add `ToggleISEMode` command with tests

andschwa:
So it can be turned on and off more easily, especially if wired to a button like in a walkthrough.


SeeminglyScience (approved):
LGTM!

PR #4121, opened 2022-08-09
Use `vscode.workspace.fs` and suppress startup banner for `dotnet` installs of PowerShell

andschwa:
This was some yak-shaving. I added `supportsProperArguments` to `PossiblePowerShellExe` so we can special-case the `dotnet` installer and drop the startup banner (since it doesn't support proper argument parsing).

Fixes https://github.com/PowerShell/vscode-powershell/issues/4111.


andschwa:
I bit off a bit more than I intended to with this one @SeeminglyScience.

SeeminglyScience (approved):
LGTM!

PR #4138, opened 2022-08-17
Add walkthrough telemetry command

andschwa:
This exposes a command that could be used in a walkthrough to send a level (presumably 1, 2, or 3) indicating satisfaction. It also refactors slightly to de-duplicate checking the extension mode before sending telemetry.


SeeminglyScience (approved):
LGTM!

PR #4141, opened 2022-08-18
Improve language client library close action message

andschwa:
The updated language client library now displays an error when the connection to the server terminates. Thing is, we already have our own restart experience where we inform the user that the terminal should be restarted, and give them a Yes/No option to do so. Unfortunately, we can't suppress this error message, and I don't think that we want to auto-restart the terminal, so I've at least made the message a little more clear.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -621,7 +621,10 @@ Type 'help' to get help.
                 },
                 closed: (): CloseHandlerResult => {
                     // We have our own restart experience
-                    return { action: CloseAction.DoNotRestart };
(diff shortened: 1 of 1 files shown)


andschwa:
@SeeminglyScience @SydneyhSmith I've asked the upstream library to either add a `PromptForRestart` option as a `CloseAction` or to allow us to suppress their error message. For now, things look like this:

Screen Shot 2022-08-17 at 5 58 07 PM

SydneyhSmith (approved):

PR #4161, opened 2022-08-31
Lock `SessionManager.start()` so only one session is started

### LINKED ISSUE #4160: Extension terminal can accidentally be spawned twice
### Prerequisites

- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

As reported on the Discord by @PrzemyslawKlys, it is possible to spawn the extension terminal twice on accident. If the extension is in the process of starting, and another event initiates a start, two terminals will be spawned. The `start()` function of the session manager is _not_ re-entrant, but as there are multiple events that may enter it, it needs a simple lock that waits for the original call to finish if it's called while already starting.

### PowerShell Version

```console
N/A
```

```console
N/A
```

### Extension Version

```console
v2022.8.5
```

### Steps to Reproduce

While extension is still starting, do something like run a PowerShell debug config which also starts the extension.

### Visuals

_No response_

### Logs

_No response_

andschwa:
Also correctly wait for disposal on extension deactivation, and more carefully dispose and set to undefined in `SessionManager.dispose()`.
 
Fixes #4160.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -181,7 +181,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<IPower
     };
 }
 
-export function deactivate(): void {
+export async function deactivate(): Promise<void> {
     // Clean up all extension features
     for (const languageClientConsumer of languageClientConsumers) {
         languageClientConsumer.dispose();
@@ -192,11 +192,11 @@ export function deactivate(): void {
     };
 
     // Dispose of the current session
-    sessionManager.dispose();
+    await sessionManager.dispose();
 
     // Dispose of the logger
     logger.dispose();
 
     // Dispose of telemetry reporter
-    telemetryReporter.dispose();
+    await telemetryReporter.dispose();
 }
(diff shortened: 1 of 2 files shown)


andschwa:
@PrzemyslawKlys thanks for the repro steps, this seems to take care of it!

SeeminglyScience (approved):
Very nice ❤️ LGTM!

Issue #4163, opened 2022-09-02
Workspace folder pick pops up if a predefined variable is used for the `powershell.cwd` setting


lewis-yeung:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

The description for the `powershell.cwd` setting says:

> Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory).

However, even if a predefined variable (e.g., `${fileDirname}` or `${fileWorkspaceFolder}`) is used for `powershell.cwd`, whenever I open a PowerShell (.ps1) file in a **multi-root workspace for the first time**, it prompts me to choose a workspace folder as the working directory:

> Select a folder to use as the PowerShell extension's working directory.

This is so annoying and should not happen. While if I use a literal path such as `C:\`, it works fine and the workspace folder pick does not pop up.

Other possibly related settings I use:

```json
{
  "powershell.startAutomatically": true,
  "powershell.integratedConsole.showOnStartup": false,
  "powershell.integratedConsole.startInBackground": true
}
```

Possibly related code:

https://github.com/PowerShell/vscode-powershell/blob/c3c857db669d97908df93267c6d3763468594e80/src/settings.ts#L326-L331

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.2.6
PSEdition                      Core
GitCommitId                    7.2.6
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.71.0
[...]

PR #4171, opened 2022-09-13
Fix incorrect docstring for `powershell.cwd`

andschwa:
This variable never could have accepted VS Code predefined variables because support for that is still an open feature request upstream. We have no API to resolve predefined variables in values given to contributed settings, and so cannot support this.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4163

### DIFF
--- a/package.json
+++ b/package.json
@@ -644,7 +644,7 @@
         "powershell.cwd": {
           "type": "string",
           "default": null,
-          "description": "An explicit start path where the PowerShell Extension Terminal will be launched. Both the PowerShell process and the shell's location will be set to this directory. Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory)."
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -644,7 +644,7 @@
         "powershell.cwd": {
           "type": "string",
           "default": null,
-          "description": "An explicit start path where the PowerShell Extension Terminal will be launched. Both the PowerShell process and the shell's location will be set to this directory. Predefined variables can be used (i.e. ${fileDirname} to use the current opened file's directory)."
+          "description": "An explicit start path where the PowerShell Extension Terminal will be launched. Both the PowerShell process's and the shell's location will be set to this directory. A real path must be provided!"
```
SeeminglyScience, line 647: Consider:

```suggestion
          "description": "An explicit start path where the PowerShell Extension Terminal will be launched. Both the PowerShell process's and the shell's location will be set to this directory. A fully resolved path must be provided!"
```

SeeminglyScience (approved):
LGTM, one minor wording suggestion

PR #4187, opened 2022-09-28
Re-implement indicator when running registered editor commands

### LINKED ISSUE #3954: Add back "PowerShell" text indicator when running a "Editor Command"
### Prerequisites

### Summary

In previous versions of the PowerShell extension, when I ran an Editor Command associated with PS1 file, it would show the text "PowerShell" in the bottom left corner.  When the command finished running, the PowerShell text would go away.  This was a nice indicator letting me when the command had finished, and I knew I could go look for the results.  In my case, my command (by use of Microsoft.VSCode_profile.ps1 with the script block registered as an Editor Command) signs the script, compiles the PS1 into an EXE format (for both .NET 4.8 and .NET 6.x formats), and then signs the executables.  This whole process takes a bit of time...long enough that the "PowerShell" text indicator "going away" really helped let me know when it was done.  With the current version, I have to wait "until I think it should be done" to check it.

image

As of version v2022.5.0, this feature no longer exists.

### Proposed Design

Please add the "PowerShell" text (or some other indicator) to the bottom (or other obvious location) of VSCode when an Editor Command is running and then hide it when the command finishes.  For an example of this, check out any previous version of the extension prior to v2022.5.0.  Hopefully this is an easy add, since it already worked in previous versions.

andschwa:
Resolves #3954 and uses the new language status bar item. Essentially just adds a spinner when the editor command is running.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -33,6 +33,7 @@ export enum SessionStatus {
     NotStarted,
     Initializing,
     Running,
+    Busy,
     Stopping,
     Failed,
 }
@@ -75,10 +76,6 @@ export const SendKeyPressNotificationType =
 export const PowerShellVersionRequestType =
     new RequestType0<IPowerShellVersionDetails, void>(
         "powerShell/getVersion");
-
-export const RunspaceChangedEventType =
-    new NotificationType<IRunspaceDetails>(
-        "powerShell/runspaceChanged");
 
 export class SessionManager implements Middleware {
     public HostName: string;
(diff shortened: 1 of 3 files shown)


SeeminglyScience (approved):
LGTM!

src/session.ts
```diff
@@ -76,10 +77,6 @@ export const PowerShellVersionRequestType =
     new RequestType0<IPowerShellVersionDetails, void>(
         "powerShell/getVersion");
 
-export const RunspaceChangedEventType =
-    new NotificationType<IRunspaceDetails>(
-        "powerShell/runspaceChanged");
```
andschwa, line 81: @SeeminglyScience this notification isn't wired up, and I wasn't planning to rewire it, so I deleted it. What do you think?

SeeminglyScience: If it's not used I'm with ya for sure. It's in git history if we need it again 👍

shangles:
I want to thank everyone involved in restoring this functionality to the PowerShell plug-in in Visual Studio Code.  I hope you all have a wonderful weekend.  Cheers!

PR #4220, opened 2022-10-21
Add assertion to build that PSES bits are built in release configuration

andschwa:
Because it's way too easy to break that! We were building in release configuration, then running the tests...and the tests were then building (and overwriting the bits) in debug configuration.


SeeminglyScience (approved):
LGTM!

Issue #4223, opened 2022-10-22
v2022.10.0 may cause previous "script" values in launch.json fail to execute


changbowen:
### Prerequisites

- [X] I have written a descriptive issue title.
- [ ] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [ ] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [ ] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

A configuration like below in launch.json is valid in the previous version 2022.8.5:
```
{
    "name": "Run PS Command",
    "type": "PowerShell",
    "request": "launch",
    "cwd": "${cwd}",
    "script": "Invoke-WebRequest www.bing.com",
}
```

With the latest 2022.10.0 it throws an error that says `PowerShell does not support debugging this file type: 'Invoke-WebRequest www.bing.com'`.

I know when run like this debugging might not work but it has been a convenient feature for quickly running something using F5 instead of going into tasks and select one to run. And it has invalidated all configurations dependent on this "feature".

Looks like it's related to changes [here](https://github.com/PowerShell/vscode-powershell/blob/193660bd9ccd25627f0ff895389fd3f0794be385/src/features/DebugSession.ts#L254).

### PowerShell Version

```console
PSVersion                      5.1.19041.1682
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.19041.1682
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

### Visual Studio Code Version

```console
Version: 1.72.2
Commit: d045a5eda657f4d7b676dedbfa7aab8207f8a075
Date: 2022-10-12T22:15:18.074Z (1 wk ago)
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
[...]

PR #4227, opened 2022-10-24
Only check a script's extension if a script file was given

andschwa:
Since it could be an inline script.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4223.


SeeminglyScience (approved):
LGTM!

PR #4235, opened 2022-10-31
Refactor `settings.ts`

andschwa:
This adds logging to our `changeSetting()` function so we can track what settings the extension makes changes to (mostly because we're really confused how `editorServicesWaitForDebugger` keeps getting turned on). It also finishes a clean up started in the strict-mode refactor to make the settings fields non-optional and uses a recursive function to populate `ISettings` from the `WorkspaceConfiguration`, complete with a regression test (which will fail if you run the tests from the launch config and have modified settings).

### DIFF
--- a/package.json
+++ b/package.json
@@ -584,7 +584,6 @@
         },
         "powershell.powerShellExePath": {
           "type": "string",
-          "default": "",
           "scope": "machine",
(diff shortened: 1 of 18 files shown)


andschwa:
Hm, this might break `developer.featureFlags` since that's a `string[]`.

SeeminglyScience (approved):
LGTM!

andschwa:
Ok let's do it!

PR #4242, opened 2022-11-02
Show warning when additional PowerShell is not found

andschwa:
Now we show a popup warning if an additional PowerShell was configured, but not found.


SeeminglyScience (approved):
LGTM!

Issue #4202, opened 2022-10-10
VSCode Extension v2022.8.5 initializes wrong user profile path


Adwaenyth:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

On a system with folder redirection for `$env:AppData` Folder to a UNC Path, the PowerShell extension does not resolve that path correctly on startup.

### PowerShell Version

```console
$PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.2.6
PSEdition                      Core
GitCommitId                    7.2.6
OS                             Microsoft Windows 10.0.19044
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
code --version
1.72.0
64bbfbf67ada9953918d72e1df2f4d8e537d340e
x64
```

### Extension Version

```console
code --list-extensions --show-versions | select-string powershell

ms-vscode.powershell@2022.8.5
```

### Steps to Reproduce

The Extension creates the initial session files in  `C:\<shareName>\<UserName>\AppData\Roaming\Code\User\globalStorage\ms-vscode.powershell\sessions\PSES-VSCode-<ID>.json` and the intial log Folders in `C:\<shareName>\<UserName>\AppData\Roaming\Code\User\globalStorage\ms-vscode.powershell\logs\<RandomID>` on the local harddrive.

[...]

PR #4276, opened 2022-11-21
Make `Logger.WriteLine` thread-safe and fix bug with UNC paths

andschwa:
This adds a simple lock to `Logger.WriteLine` since it's not re-entrant, hence we were losing logs. Now concurrent calls to it wait for each other to finish. It also _should_ fix https://github.com/PowerShell/vscode-powershell/issues/4202, as the bug appears to be with the scheme used in `globalStorageUri` being `vscode-userdata` instead of `file`, and `vscode.Uri.fsPath` claiming it does not care about the scheme, but in fact essentially requires it to be `file`.

### DIFF
--- a/src/features/ExternalApi.ts
+++ b/src/features/ExternalApi.ts
@@ -169,7 +169,9 @@ export class ExternalApiFeature extends LanguageClientConsumer implements IPower
     }
 
     public getStorageUri(): vscode.Uri {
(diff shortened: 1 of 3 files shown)


SeeminglyScience (approved):
LGTM! Well done getting this figured out ❤️

Issue #4129, opened 2022-08-12
Additional PowerShell executables: Warning if versionName cannot be resolved


sofa1780:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Summary

If the PowerShell setting for assigning additional execution paths (powershell.powerShellAdditionalExePaths) is used with an invalid versionName parameter shown within the snippet down below no warning appears.

This can lead to an unnoticed use of a different PowerShell version than defined seemingly within the settings since the value "Windows PowerShell **5.1** (x86)" for the parameter VersionName is not valid.
Instead, the correct value should be "Windows PowerShell (x86)".

```
"powershell.powerShellAdditionalExePaths": {
    "Windows PowerShell 5.1 (x64)": "%SystemRoot%\\syswow64\\WindowsPowerShell\\v1.0\\powershell.exe",
    "Windows PowerShell 5.1 (x86)": "%SystemRoot%\\system32\\WindowsPowerShell\\v1.0\\powershell.exe",
},
"powershell.powerShellDefaultVersion": "Windows PowerShell 5.1 (x86)",
```
### Proposed Design

This manner could be accomplished by either informing the user or accepting the user-defined value for parameter versionName.

PR #4295, opened 2022-11-30
Show warning if `powerShellDefaultVersion` is set but not found

andschwa:
Resolves https://github.com/PowerShell/vscode-powershell/issues/4129.


SeeminglyScience (approved):
LGTM!

Issue #3435, opened 2021-05-29
Extension incorrectly attempts x64 update on Arm64 platform


pb5050:
I use nothing but native arm64..... 

I get this alert in visual studio code (arm64) an alert about an updated version of powershell.

![Powershell Issue](https://user-images.githubusercontent.com/58855947/120072346-0dbc5200-c059-11eb-946e-80d5d2071bd9.png)

when clicking it tries to install x64 version  instead of arm. would be nice to have an arm installer or recognize im using arm and not x64.

andschwa:
We (unfortunately) have our own logic in the extension to suggest updating PowerShell. Perhaps we'll fix this bug, but frankly I've been contemplating removing that logic entirely. PowerShell _already_ has its own logic to prompt users to update; I don't think the extension has any business adding duplicative logic on top of that, especially (as evidenced by this bug, among several) that it's _buggy_. It's not worth the maintenance cost.

PR #4298, opened 2022-12-02
Skip auto-update on unsupported Windows architectures

andschwa:
While PowerShell now works on ARM64 Windows, it does not yet have an MSI installer, so for now instead of erroneously attempting to install the x64 version, we just recommend that the user update (like we do for Linux). Note that the string is now capitalized due to a server-side change that simply returns .NET's
`System.Runtime.InteropServices.Architecture` enum as a string.

Paired with https://github.com/PowerShell/PowerShellEditorServices/pull/1964, this resolves https://github.com/PowerShell/vscode-powershell/issues/3435.

### DIFF
--- a/src/features/UpdatePowerShell.ts
+++ b/src/features/UpdatePowerShell.ts
@@ -15,7 +15,7 @@ import { LanguageClient } from "vscode-languageclient/node";
 import { Logger } from "../logging";
 import { SessionManager } from "../session";
 import { changeSetting } from "../settings";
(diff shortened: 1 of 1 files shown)


SeeminglyScience (approved):
LGTM!

andschwa:
Screenshot 2022-12-02 at 11 17 38 AM

With an Arm64 `pwsh` on Windows, still get the update, but just with guidance!

Issue #3226, opened 2021-03-05
PowerShell update checker ignores `POWERSHELL_UPDATECHECK` environment variable


trolleyboy:
### System Details

System Details Output
```
### VSCode version: 1.53.2 622cb03f7e070a9670c94bae1a45d78d7181fbd4 x64

### VSCode extensions:
azsdktm.SecurityIntelliSense@0.8.0     
AzurePolicy.azurepolicyextension@0.1.1 
DotJoshJohnson.xml@2.5.1
eamodio.gitlens@11.2.1
everettjf.filter-line@2.0.1
jinsihou.diff-tool@0.0.1
ms-azure-devops.azure-pipelines@1.183.0
ms-azuretools.vscode-azureappservice@0.21.0
ms-azuretools.vscode-azureresourcegroups@0.3.0
ms-dotnettools.vscode-dotnet-runtime@1.0.0
ms-vscode.azure-account@0.9.6
ms-vscode.powershell@2021.2.2
msazurermtools.azurerm-vscode-tools@0.14.1
ryu1kn.text-marker@1.11.0
Summer.azure-event-hub-explorer@0.1.2


### PSES version: 2.3.0.0

### PowerShell version:

Name                           Value
----                           -----
PSVersion                      7.0.5
PSEdition                      Core
GitCommitId                    7.0.5
OS                             Microsoft Windows 10.0.19042
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}       
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Issue Description

I am experiencing a problem with...
VS Code notifies as follows...
```
You have an old version of PowerShell (7.0.5). The current latest release is v7.1.2. Would you like to update the version? (This will close ALL pwsh terminals running in this Visual Studio Code session)
```
However I have set $Env:POWERSHELL_UPDATECHECK to "LTS" per https://docs.microsoft.com/en-us/powershell/scripting/whats-new/what-s-new-in-powershell-70?view=powershell-7.1#new-version-notification

#### Expected Behaviour

I'd expect the notification to only appear when updates to the LTS version of Powershell 7 are available. EG: 7.0.6.

#### Actual Behaviour

VS Code notifies as follows...
```
[...]

PR #4306, opened 2022-12-05
Rewrite `UpdatePowerShell` feature

andschwa:
It's way better now, relying on GitHub less and instead mimicking what PowerShell itself does, and resolves https://github.com/PowerShell/vscode-powershell/issues/3226.

Requires https://github.com/PowerShell/PowerShellEditorServices/pull/1965.

### DIFF
--- a/.vsts-ci/templates/ci-general.yml
+++ b/.vsts-ci/templates/ci-general.yml
@@ -70,6 +70,7 @@ steps:
   inputs:
     targetType: inline
     script: |
+      Get-ChildItem env:
       Get-Module -ListAvailable Pester
       Install-Module InvokeBuild -Scope CurrentUser -Force
       Install-Module platyPS -Scope CurrentUser -Force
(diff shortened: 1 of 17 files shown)


andschwa:
@SeeminglyScience you can pretty much ignore the change from `Logger` to `ILogger`, that was so I could instantiate a `TestLogger` class that doesn't do anything and pass it along to the constructor of `UpdatePowerShell` (and probably other things when we get more robust client unit tests).

For Linux, macOS, and unsupported (mostly Arm64) architectures on Windows, the update prompt now opens the user's configured URL handler to the GitHub release for the new release. 🥳 While we never supported auto-updating on Linux, this does mean I took out the `brew cask update` script for macOS. It _can_ work, but it should _only_ be run if we can assert first that the `pwsh` we're updating was installed by `brew`, which we weren't doing and isn't always the case. For instance, I install straight from the `.pkg` installer and so had turned off the update prompt. Also, we should do it using `vscode.env.ShellExecution` instead of relying on the language server to run a script (to update its own executable). So I've punted that work because it's complicated and requires an update to do right.

SeeminglyScience (approved):
LGTM!

PR #4308, opened 2022-12-06
Restore original settings after disabling ISE mode

andschwa:
We were previously setting them to `undefined` which mostly worked, unless the user had customized the setting. This was most apparent and commonly done with `colorTheme`. Now we save when we enable, and restore to that value. Note that will only work for the original session, but it does allow the user to test ISE mode out without repercussions.

We could probably save the settings into the global storage context, but that might get hairy!

I don't know if we have an issue for this, but it _constantly_ bothered me when running the tests.

### DIFF
--- a/src/features/ISECompatibility.ts
+++ b/src/features/ISECompatibility.ts
@@ -28,6 +28,7 @@ export class ISECompatibilityFeature implements vscode.Disposable {
 
     private _commandRegistrations: vscode.Disposable[] = [];
     private _iseModeEnabled: boolean;
+    private _originalSettings: Record<string, boolean | string | undefined> = {};
 
     constructor() {
         // TODO: This test isn't great.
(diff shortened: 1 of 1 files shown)


SydneyhSmith (approved):

gerneio:
@andschwa Not sure why, but just installed the [v2022.12.1-preview](https://github.com/PowerShell/vscode-powershell/releases/tag/v2022.12.1-preview) version and this fix didn't seem to work. Toggling back and forth from ISE mode does NOT retain the `colorTheme`.

andschwa:
Hm, thanks for the report @gerneio. I'll take a second look.

andschwa:
@gerneio I can't repro, this appears to be working as intended for me. I have my theme set to "Solarized Light." I run "Toggle ISE Mode," it applies the ISE theme, and then I run the same command again and it's restoring Solarized Light for me.

PR #4331, opened 2022-12-15
Automatically find "PowerShell Daily" at its common path

andschwa:
Resolves https://github.com/PowerShell/vscode-powershell/issues/4329 by negating the need to manually add the PowerShell daily path.


andschwa:
Ok @TylerLeonhardt all done.

andschwa:
@SeeminglyScience I kind of wonder if we should just copy/paste that function without the `-daily` to support finding the `pwsh` installed by this script _without_ the `-Daily` parameter...but what would we call it, "PowerShell Install Script"?

SeeminglyScience (approved):
LGTM!

andschwa:
@TylerLeonhardt available in the preview now!

PR #4357, opened 2023-01-03
Fix `@vscode/vsce` dependency after its rename

andschwa:
Also fix the documentation to recommend Node v16, as that's what VS Code currently uses, and v14 is too old for dependencies such as `@vscode/test-electron`, and the badge links in the readme since the upstream URL changed without notice.

This should fix the release process so we can get another preview out.


SeeminglyScience (approved):
LGTM!

PR #4425, opened 2023-02-18
Hide last line in folded ranges for ISE Mode

fflaten:
## PR Summary

PowerShell ISE folds ranges to a single line unlike the current default in this extension:
image
image

This PR sets the `powershell.codeFolding.showLastLine` option to `false` in ISE mode to be more similar with ISE behavior.
image

There is one major difference: ISE appended the closing char for most ranges which VSCode doesn't atm. That might be a dealbreaker, so submitting as a suggestion without discussion. Feel free to close if you don't agree.

## PR Checklist

- [x] Summarized changes
- [x] PR has tests


andschwa:
TBH I really like how the ISE handles this, but yeah even without the end characters I think this is good!

andschwa (approved):

PR #4464, opened 2023-03-17
Remove `preview` label from marketplace

andschwa:
This was a lot of annoying work just to remove the `preview` label from the marketplace web page, since it shows from the "latest" (inclusive of pre-release) releases. Our only other way to determine at package and release time that a version is a pre-release is to re-use the version-from-changelog logic of the release tools. But that meant splitting out the relevant parts so the build script didn't take a dependency on the `PowerShellForGitHub` module, and splitting the CI into a "build and package" step separate from a "test" step. The latter happened because in CI we test against PowerShell 5.1...and that doesn't have the `semver` type, so it _couldn't_ use the version logic, hence the "build and package" step now requires PowerShell Core. 🙃

But hey, at least I snuck in a fix for a long-standing `TODO` to better validate the version number we hand to the release tools when doing a release.

### DIFF
--- a/.vsts-ci/templates/ci-general.yml
+++ b/.vsts-ci/templates/ci-general.yml
@@ -7,13 +7,6 @@ parameters:
     default: false
 
 steps:
-- task: PowerShell@2
-  displayName: PowerShell version
-  inputs:
-    targetType: inline
-    script: $PSVersionTable
-    pwsh: ${{ parameters.pwsh }}
-
 - checkout: self
 
 # NOTE: We either checkout the Git repo for PowerShellEditorServices, or we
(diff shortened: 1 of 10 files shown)


.vsts-ci/templates/ci-general.yml
```diff
@@ -62,19 +55,32 @@ steps:
...
-      Get-ChildItem env:
-      Get-Module -ListAvailable Pester
       Install-Module InvokeBuild -Scope CurrentUser -Force
       Install-Module platyPS -Scope CurrentUser -Force
-      Invoke-Build -Configuration Release
+      Invoke-Build -Configuration Release Package
       $PackageJson = Get-Content -Raw package.json | ConvertFrom-Json
       Write-Host "##vso[task.setvariable variable=vsixPath]$(Resolve-Path powershell-$($PackageJson.version).vsix)"
     workingDirectory: $(Build.SourcesDirectory)/vscode-powershell
+    pwsh: true
```
SeeminglyScience, line 70: smart lol

tools/ReleaseTools.psm1
```diff
@@ -6,40 +6,26 @@
...
-        switch ($RepositoryName) {
-            "vscode-powershell" {
-                Push-Location -Path "$PSScriptRoot/../"
-            }
-            "PowerShellEditorServices" {
-                Push-Location -Path "$PSScriptRoot/../../PowerShellEditorServices"
+    Use-Repository -RepositoryName $RepositoryName -Script {
+        $Branch = git branch --show-current
+        if ($Branch -ne "release") {
+            if ($PSCmdlet.ShouldProcess("release", "git checkout -B")) {
```
SeeminglyScience, line 25: ❤️ love a good should process impl

tools/VersionTools.psm1
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#requires -Version 7.0
+
+using namespace System.Management.Automation
+
+class RepoNames: IValidateSetValuesGenerator {
```
SeeminglyScience, line 8:
```suggestion
class RepoNames : IValidateSetValuesGenerator {
```

nit

.vsts-ci/templates/ci-general.yml
andschwa: Turns out that the GitHub PowerShell action defaults to 5.1 on Windows and 7 elsewhere 😳

tools/ReleaseTools.psm1
andschwa: No idea if it actually still works, I never use it lmao.

SeeminglyScience (approved):
LGTM!

PR #4481, opened 2023-03-27
Remove problematic `[RepoNames]` PowerShell class

andschwa:
It was admittedly over-engineered and caused problems since PowerShell classes aren't imported across modules.

Instead just use `ValidateSet` directly wherever the command is either use-facing or switches on the string directly.


SeeminglyScience (approved):
LGTM!

Issue #4521, opened 2023-04-16
Better error message when using unsupported PowerShell


saeedt2:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [ ] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

I am getting this error every time I start VS code
4/16/2023 1:10:06 PM [NORMAL] - pwsh.exe started.

4/16/2023 1:10:06 PM [NORMAL] - Waiting for session file...

4/16/2023 1:10:36 PM [WARNING] - Loading the PowerShell extension is taking longer than expected. If you're using privilege enforcement software, this can affect start up performance.

4/16/2023 1:14:08 PM [NORMAL] - Timed out waiting for session file to appear!

4/16/2023 1:14:08 PM [ERROR] - PowerShell process failed to start: 

4/16/2023 1:14:08 PM [ERROR] - Timed out waiting for session file to appear!

4/16/2023 1:14:08 PM [ERROR] - Unknown session status 'undefined' with reason 'undefined

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.1.4
PSEdition                      Core
GitCommitId                    7.1.4
OS                             Microsoft Windows 10.0.19044
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
ms-vscode.powershell@2023.3.3

1.77.3
704ed70d4fd1c6bd6342c436f1ede30d1cff4710
x64
```

### Extension Version

```console
ms-vscode.powershell@2023.3.3
```

### Steps to Reproduce

[...]

PR #4532, opened 2023-04-21
Handle end-of-support PowerShell with error message

andschwa:
Significantly improves our handling of PowerShell failing to start because it is unsupported.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4521.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -24,7 +24,7 @@ import {
     OperatingSystem, PowerShellExeFinder
 } from "./platform";
 import { LanguageClientConsumer } from "./languageClientConsumer";
-import { SemVer } from "semver";
+import { SemVer, satisfies } from "semver";
 
 export enum SessionStatus {
     NeverStarted,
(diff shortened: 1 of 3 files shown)


andschwa:
This can probably be improved, but at least it works!

andschwa:
I'm still working on adding more actions, like actual actionable prompts, to our various possible reasons to fail at startup. Should help users a ton, I hope.

src/session.ts
```diff
@@ -458,35 +458,57 @@ export class SessionManager implements Middleware {
...
+            // We should kill the process in case it's stuck.
+            void languageServerProcess.dispose();
+
+            // PowerShell never started, probably a bad version!
+            const version = await languageServerProcess.getVersionCli();
+            let shouldUpdate = true;
+            if (satisfies(version, "<5.1.0")) {
+                void this.setSessionFailedGetPowerShell(`PowerShell ${version} is not supported, please update!`);
+            } else if (satisfies(version, ">=5.1.0 <6.0.0")) {
+                void this.setSessionFailedGetPowerShell("It looks like you're trying to use Windows PowerShell, which is supported on a best-effort basis. Can you try PowerShell 7?");
```
JustinGrote, line 470: Worth adding an aka.ms link to the install instructions here maybe?

JustinGrote (commented):
LGTM tho I don't have an ability to actually test the code right now. I like the additional link to open an issue for crashes in general!

src/session.ts
andschwa: So with `this.setSessionFailedGetPowerShell` this actually now shows an error window popup with a button "Open PowerShell Install Documentation" that opens the link https://aka.ms/get-powershell-vscode!

SeeminglyScience (approved):
LGTM!

PR #4541, opened 2023-04-24
Fix Dependabot pull requests limit

andschwa:
There seems to be a bug in the backend that thinks we're hitting the default limit of 5 even when no PRs are open.


SeeminglyScience (approved):
LGTM!

PR #4543, opened 2023-04-24
Update startup logic to handle session failure reasons

andschwa:
These used to be sent given that the logic existed, but the server was no longer sending them at all. Now that it does, we can handle them. Also see https://github.com/PowerShell/PowerShellEditorServices/pull/2018.


SeeminglyScience (approved):
LGTM!

PR #4547, opened 2023-04-26
Fix flaky test

andschwa:
I believe this test was flaky due to race conditions caused by the events. Awaiting promises instead should eliminate those races.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -6,6 +6,7 @@
     ".jsx"
   ],
   "require": "source-map-support/register",
-  "timeout": 10000,
+  "timeout": 30000,
+  "slow": 2000,
   "spec": "out/test/**/*.test.js"
 }
--- a/test/core/settings.test.ts
+++ b/test/core/settings.test.ts
@@ -6,7 +6,6 @@ import * as vscode from "vscode";
 import { Settings, getSettings, getEffectiveConfigurationTarget, changeSetting, CommentType } from "../../src/settings";
 
 describe("Settings E2E", function () {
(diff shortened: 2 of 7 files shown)


andschwa:
Let's see if they all pass.

JustinGrote:
> I believe this test was flaky due to race conditions caused by the events. Awaiting promises instead should eliminate those races.

Events are so hard to test. I think it might still race flake so we can try using the done() callback in the event to make sure it doesn't end until all the assertions are complete. I'll take a look

andschwa:
I did also increase the timeout in the release PR while I was adding a skip:

    Binary Modules
      ✔ Debugs a binary module script (11974ms)
      ✔ Stops at a binary module breakpoint (10853ms)

But IIRC it was set to 20m000ms and they're coming in around 10,000ms.

Anyway, I'm down to merge this and let upcoming work test the flakiness. They passed on the first try this time!

andschwa:
Re-running checks to see if they pass again.

andschwa:
> Events are so hard to test. I think it might still race flake so we can try using the done() callback in the event to make sure it doesn't end until all the assertions are complete. I'll take a look

That's essentially what the `Promise` does. By doing the assignment within the event as a resolution to the promise, the latter test code which needs the value waits for that resolved promise (and thus assignment) to actually happen (which is in the background asynchronously via the event). So instead of racing the event to its completion, the test code `await`s for that logic to happen.

SeeminglyScience (approved):
LGTM!

PR #4548, opened 2023-04-27
Use named debug configurations instead of duplicating them

andschwa:
Now that we have a record of debug configurations we can fix these TODOs.


SeeminglyScience (approved):
LGTM!

PR #4584, opened 2023-05-20
Robustify startup, error handling, and logging

andschwa:
This was a couple late nights that resulted in an overhaul of our startup, error handling, and logging logic. Mostly because I realized how dang easy it was to get things into a funky state when I added `editorServicesWaitForDebugger` to the list of settings that will prompt for a restart. This made it really easy to test, especially in the "still starting up" state. It is still possible to generate perhaps more error messages than I'd like, but they mostly come from the LSP client library. In fact, the last bug I tracked down is that if the client failed to start, it can't be stopped (which throws an exception) AND it can't be disposed, the latter of which I think is an upstream bug I'll need to investigate.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -6,7 +6,7 @@
     ".jsx"
   ],
   "require": "source-map-support/register",
-  "timeout": 30000,
+  "timeout": 60000,
   "slow": 2000,
   "spec": "out/test/**/*.test.js"
 }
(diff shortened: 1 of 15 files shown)


andschwa:
This got big enough I split it out into sensible commits.

andschwa:
Wow it really was just the timeout 😬

SeeminglyScience (commented):
LGTM!

SeeminglyScience (approved):
LGTM!

Issue #4522, opened 2023-04-17
Replace custom logger with tslog


JustinGrote:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Summary

Prerequisite to: https://github.com/PowerShell/vscode-powershell/issues/4441

To implement this feature, this is an opportune time to look at taking a dependency on a proper logging system to allow for more sinks and a more robust operation.

### Proposed Design

I have a lot of love for [tslog](https://tslog.js.org/#/), which is extremely extensible and can be tied into the existing methods to avoid breaking changes.

@andschwa would you have an appetite for taking this dependency or would you prefer to just overhaul the existing implementation?

andschwa:
I'd be totally fine with tslog, it looks nice! Just want to point out that for #4441 we have to wait until our engine is updated to 1.76...and we're a long way from that at 1.67.0. We are tied to the version that Azure Data Studio uses, so don't expect it to change quickly!

Issue #4441, opened 2023-03-03
Support VS Code 'Set Log Level'


juvtib:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Summary

VS Code version 1.76 has a "Developer: Set Log Level..." command. I tried using the command while I had a PowerShell project loaded. But I did not see an explicit option for the PowerShell extension. I'm not sure if the extension outputs to a generic log or if the PowerShell extension should be represented on this list. 

![Screen capture of VS Code log menu. There is no menu item for the PowerShell extension.](https://user-images.githubusercontent.com/51273929/222677600-a76c44a9-88e5-404d-a117-1bbc3258985d.png)

'Set Log Level' seems like a useful command. Please take a look at it and see if the PowerShell extension can support this VS Code command. 

## References 

* [Set Log Level | Visual Studio Code February 2023](https://code.visualstudio.com/updates/v1_76#_configure-default-log-level)
* [Logs | Troubleshooting Guide](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md#logs)

JustinGrote:
Thanks for the Issue Submit!

So this is a fairly new API that logging has to be explicitly written for:
https://code.visualstudio.com/updates/v1_73#_log-output-channel

The PowerShell extension has its own logger implementation and would need to be updated for this new API. 
https://vscode.dev/github/powershell/vscode-powershell/blob/019a9f8cf8509eaffff994af82b804b39e19d883/src/logging.ts#L49 

And then wired up so that `writeAtLevel` and `writeLine` and related code paths submit properly to the output channel api, then the existing log level setting would need to be retired.

juvtib:
Thanks for taking a look at it. 

For reference, here is that logging line without Visual Studio Code for the Web. 
https://github.com/PowerShell/vscode-powershell/blob/019a9f8cf8509eaffff994af82b804b39e19d883/src/logging.ts#L49

JustinGrote:
Working on this in https://github.com/PowerShell/vscode-powershell/pull/5065 but there's some upstream vscode bugs I need addressed.

PR #4641, opened 2023-06-27
Update VS Code engine to 1.79.0

andschwa:
Now that Azure Data Studio is doing the same in:
https://github.com/microsoft/azuredatastudio/pull/23482

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "publisher": "ms-vscode",
   "description": "Develop PowerShell modules, commands and scripts in Visual Studio Code!",
   "engines": {
-    "vscode": "^1.67.0"
+    "vscode": "^1.79.0"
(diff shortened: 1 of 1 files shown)


andschwa:
@JustinGrote this will unblock https://github.com/PowerShell/vscode-powershell/issues/4441 and https://github.com/PowerShell/vscode-powershell/issues/4522!

andschwa:
FWIW I don't think this is jumping the gun to merge today as they are clearly updating and we still need to put this through preview and into stable.

SeeminglyScience (approved):
LGTM!

JustinGrote:
Huzzah! We'll give this a bit to make sure there's no regressions and maybe I'll get started on the updated logger.

PR #4661, opened 2023-07-13
Update readme

andschwa:
Fixed issues with badges, removed Gitter (not actively monitored by us so let's prefer advertising Discord), lots of updates and reorganizing, more useful links (like to the official guide), removed confusing or not useful information (such as how to install from the script or open the examples, neither of which we actively keep updated), and sorry Patrick but after changing my name now alphabetically it's first.


SydneyhSmith (approved):
LGTM thanks for the updates

PR #4686, opened 2023-08-04
Enhance `additionalPowerShellExes` setting

andyleejordan:
Now the warning that's emitted when an additional PowerShell executable isn't found can be suppressed with a setting. If the setting's value has single or double quotes, it handles it. If the executable name wasn't appended, it handles it. Smarter!

Work in progress because it needs unit tests... @JustinGrote what do you think of this?

### DIFF
--- a/test/core/platform.test.ts
+++ b/test/core/platform.test.ts
@@ -11,6 +11,7 @@ import * as sinon from "sinon";
 import * as platform from "../../src/platform";
 import * as fs from "fs"; // NOTE: Necessary for mock-fs.
 import * as vscode from "vscode";
+import { stripQuotePair } from "../../src/utils";
 
 // We have to rewire the platform module so that mock-fs can be used, as it
 // overrides the fs module but not the vscode.workspace.fs module.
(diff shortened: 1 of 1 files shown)


andyleejordan:
It might not have been bothering anyone but me, but it was really bothering me.

src/platform.ts
```diff
@@ -223,9 +229,39 @@ export class PowerShellExeFinder {
...
+                yield new PossiblePowerShellExe(exePath, versionName);
+
+                // Also search for `pwsh[.exe]` and `powershell[.exe]` if missing
+                const args: [string, undefined, boolean, boolean]
+                    // Must be a tuple type and is suppressing the warning
+                    = [versionName, undefined, true, true];
+
+                // Handle Windows where '.exe' and 'powershell' are things
+                if (this.platformDetails.operatingSystem === OperatingSystem.Windows) {
+                    if (!exePath.endsWith("pwsh.exe") && !exePath.endsWith("powershell.exe")) {
```
JustinGrote, line 253: Agree probably need some tests to validate all these potential permutations, looks good to me though.

andyleejordan: I realized last night I should also be able to handle the common case of `~/`, and share that and the quote-stripping with the `cwd` setting, and then do a preview and _then_ work on saving the `cwd` in a cache instead of a setting.

andyleejordan: Ok so I decided that since the Python extension uses `untildify` I'm fine bringing it too, which means I'll get that taken care of...in the next PR. @JustinGrote could you give this a review? It's got tests now!

JustinGrote:
Sorry for the delay in testing, it LGTM to me too even tho it just got merged :)

PR #4687, opened 2023-08-05
Support `~`, `./` and named workspace folders in `cwd`

### LINKED ISSUE #4557: Allow means to avoid adding an absolute Cwd path to the workspace settings
### Prerequisites

### Summary

Our team's repositories all track our `.code-workspace` files to make sure people are working with the same settings. However, we obviously can't include an absolute path in the `powershell.cwd` setting because we can't ensure it will be the same between users. Unfortunately, there seems to be no way to avoid this issue, and worst case, we keep having to remove the automatically appended setting.

### Proposed Design

I can think of 3 ways to circumvent this problem:

1. Add an option that, when opening a file in PS-language, recursively scan in the file's directory for a workspace setting file, going up a folder if nothing could be found. If one is found, pick the workspace directory the file is located in, (choosing the nearest if there are nested folders) and use that as the Cwd for the session.
2. Add an option that reads the Cwd path from a different file that can be included in the `.gitignore`. You could enforce it to be located in the same directory as the workspace settings.
3. Add an option that doesn't write the Cwd path into the settings, and just caches them instead, prompting the user to specify a folder every time after the cache is cleared.

andyleejordan:
This makes them way less annoying. I'm ok taking a dependency on untildify as it's a very simple package, and the Python extension for VS Code also uses it. However, it must remain at v4.0.0, as the latest version, v5.0.0, is pure ESM and therefore cannot be loaded by VS Code.

https://github.com/sindresorhus/untildify/releases/tag/v5.0.0

Now also handles paths relative to a single workspace folder, or understands the name of a workspace folder in multi-root workspaces. Resolves #4557.

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -8,6 +8,7 @@ import vscode = require("vscode");
 import { integer } from "vscode-languageserver-protocol";
 import { ILogger } from "./logging";
 import { changeSetting, getSettings, PowerShellAdditionalExePathSettings } from "./settings";
+import untildify from "untildify";
 
 // This uses require so we can rewire it in unit tests!
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-var-requires
@@ -232,10 +233,12 @@ export class PowerShellExeFinder {
     private *enumerateAdditionalPowerShellInstallations(): Iterable<IPossiblePowerShellExe> {
         for (const versionName in this.additionalPowerShellExes) {
             if (Object.prototype.hasOwnProperty.call(this.additionalPowerShellExes, versionName)) {
-                const exePath = utils.stripQuotePair(this.additionalPowerShellExes[versionName]);
+                let exePath = utils.stripQuotePair(this.additionalPowerShellExes[versionName]);
                 if (!exePath) {
                     continue;
                 }
+
+                exePath = untildify(exePath);
 
                 // Always search for what the user gave us first
                 yield new PossiblePowerShellExe(exePath, versionName);
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ updates:
         - "esbuild"
         - "eslint"
         - "@typescript-eslint/*"
+  ignore:
+    - dependency-name: "untildify"
+      versions: ["5.x"]
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -6,7 +6,7 @@
     ".jsx"
   ],
   "require": "source-map-support/register",
-  "timeout": 60000,
+  "timeout": 600000,
   "slow": 2000,
   "spec": "out/test/**/*.test.js"
 }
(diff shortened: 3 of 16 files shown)


andyleejordan:
I _definitely_ thought we had an issue (possibly closed) for this...it's come up at least, a number of times.

src/platform.ts
```diff
@@ -232,11 +233,13 @@ export class PowerShellExeFinder {
     private *enumerateAdditionalPowerShellInstallations(): Iterable<IPossiblePowerShellExe> {
         for (const versionName in this.additionalPowerShellExes) {
             if (Object.prototype.hasOwnProperty.call(this.additionalPowerShellExes, versionName)) {
-                const exePath = utils.stripQuotePair(this.additionalPowerShellExes[versionName]);
+                let exePath = utils.stripQuotePair(this.additionalPowerShellExes[versionName]);
                 if (!exePath) {
                     continue;
                 }
 
+                exePath = untildify(exePath);
```
andyleejordan, line 241: @JustinGrote toss a `path.resolve()` around this?

JustinGrote: Assuming it's handled later on, the only reason I can think not is if you *wanted* the relative path later on if it happened to show up that way. I'm not sure how safe `untildify` is or if it maybe even calls resolve internally.

JustinGrote:
> I _definitely_ thought we had an issue (possibly closed) for this...it's come up at least, a number of times. @JustinGrote any idea where that went?

Maybe you are thinking of https://github.com/PowerShell/vscode-powershell/issues/3827 which was closed?

src/platform.ts
andyleejordan: This is the entirety of `untildify`:

```typescript
import os from 'node:os';

const homeDirectory = os.homedir();

export default function untildify(pathWithTilde) {
	if (typeof pathWithTilde !== 'string') {
		throw new TypeError(`Expected a string, got ${typeof pathWithTilde}`);
	}

	return homeDirectory ? pathWithTilde.replace(/^~(?=$|\/|\\)/, homeDirectory) : pathWithTilde;
}
```

I think I just need to test if it's ok here with `..` in the middle of a path.

No path resolution, just substitution for `~`. So like, the VS Code Terminal API is what's picky about having an absolute path for startup for its `cwd` argument.

andyleejordan: Yeah I think it's fine. `cwd` actually also doesn't need `path.resolve()`...it just needed to be smarter.

andyleejordan:
Hm, that wasn't it @JustinGrote. Maybe it was just on Discord. Anyway, this resolves #4557 now since it no longer writes the `cwd` to the workspace settings. For multi-root workspaces, instead a name can be set (like 'vscode-powershell' for this repo) and then that workspace folder will be used. For a single-root workspace, a relative path will be resolved to its root.

JustinGrote:
I'll run this one through the paces once the VSIX is built, I have a couple multi root workspaces I can test against.

andyleejordan:
@JustinGrote awesome! Please do, I'll have a pre-release out soon. Main thing with multi-root workspaces is that this is only supporting putting the _name_ of the workspace as the `cwd`. No relative paths to it or anything, just a choice. But it's not an absolute path any more!

JustinGrote:
LGTM with multi workspace that I have, just a comment on the UX on the original issue.

PR #4696, opened 2023-08-11
Fix race condition with displaying PowerShell name on icon

andyleejordan:
Refactors the logic a bit so the that Language Status Icon has the correct details when switching sessions. Also delete superfluous (and apparently broken) migration code.


SeeminglyScience (approved):
LGTM!

PR #4703, opened 2023-08-16
Respect file path casing in extension commands

### LINKED ISSUE #2960: $psEditor.Workspace.OpenFile  puts file name into all lower case.
System Details Output
```

### VSCode version: 1.49.0 e790b931385d72cf5669fcefc51cdf65990efa5d x64

### VSCode extensions:      
andyyaldoo.vscode-json@1.5.0
jock.svg@1.3.9
ms-azure-devops.azure-pipelines@1.174.2
ms-azuretools.vscode-azurefunctions@0.24.0
ms-dotnettools.csharp@1.23.2
ms-mssql.mssql@1.9.0
ms-python.python@2020.8.108011
ms-vscode-remote.remote-containers@0.140.1
ms-vscode-remote.remote-ssh@0.51.0
ms-vscode-remote.remote-ssh-edit@0.51.0
ms-vscode-remote.remote-wsl@0.44.5
ms-vscode-remote.vscode-remote-extensionpack@0.20.0
ms-vscode.azure-account@0.9.3
ms-vscode.powershell@2020.6.0
ms-vscode.powershell-preview@2020.7.0


### PSES version: 2.3.0.0

### PowerShell version:

PSVersion                      7.0.3
GitCommitId                    7.0.3
OS                             Microsoft Windows 10.0.18363
```

This also appears with vs-code insiders and the 2020.9.0 extension ... 

The issue is that the $psEditor.Workspace.OpenFile - which is called from various places  - like PsEdit, changes the file name to lower case. This is usually just a cosmetic problem, but I have found a case were pester ran a script using its lower-cased name and the script failed because it tried to do a case sensitive replace on part of its own name.

andyleejordan:
Per https://nodejs.org/en/docs/guides/working-with-different-filesystems we should really just keep the file path we're given, no "normalizing" to be done especially since it is wrong to infer from the OS if the filesystem is case-sensitive or not. Gone are the days where Windows and macOS could be presumed case-insensitive, they now both support and are often found with with case-sensitive filesystems. Plus, there was really no reason to be doing this in the first place. Tested interactively, this works just fine since the VS Code APIs we're using handle case-insensitivity for us.

Resolves #2960.

### DIFF
--- a/src/features/ExtensionCommands.ts
+++ b/src/features/ExtensionCommands.ts
@@ -68,9 +68,11 @@ export const GetEditorContextRequestType =
 export interface IGetEditorContextRequestArguments {
 }
 
+// NOTE: The server at least now expects this response, but it's not used in any
+// way. In the future we could actually communicate an error to the user.
 enum EditorOperationResponse {
-    Unsupported = 0,
     Completed,
+    Failed
 }
 
 export const InsertTextRequestType =
@@ -148,6 +150,7 @@ interface IInvokeRegisteredEditorCommandParameter {
 export class ExtensionCommandsFeature extends LanguageClientConsumer {
     private commands: vscode.Disposable[];
     private handlers: vscode.Disposable[] = [];
+    private statusBarMessages: vscode.Disposable[] = [];
     private extensionCommands: IExtensionCommand[] = [];
 
     constructor(private logger: ILogger) {
@@ -216,9 +219,7 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
 
             this.languageClient.onRequest(
                 NewFileRequestType,
-                // NOTE: The VS Code API does not support naming a file as it's
-                // opened, only when it's saved. Hence the argument is not used.
-                (_filePath) => this.newFile()),
+                (_content) => this.newFile(_content)),
 
             this.languageClient.onRequest(
                 OpenFileRequestType,
@@ -266,6 +267,9 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
         }
         for (const handler of this.handlers) {
             handler.dispose();
+        }
+        for (const statusBarMessage of this.statusBarMessages) {
+            statusBarMessage.dispose();
         }
     }
 
@@ -355,27 +359,35 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
         };
     }
 
-    private async newFile(): Promise<EditorOperationResponse> {
-        const doc = await vscode.workspace.openTextDocument({ content: "" });
+    private async newFile(content: string): Promise<EditorOperationResponse> {
+        const doc = await vscode.workspace.openTextDocument(
+            { language: "powershell", content: content });
         await vscode.window.showTextDocument(doc);
         return EditorOperationResponse.Completed;
     }
 
     private async openFile(openFileDetails: IOpenFileDetails): Promise<EditorOperationResponse> {
         const filePath = await this.resolveFilePathWithCwd(openFileDetails.filePath);
-        const doc = await vscode.workspace.openTextDocument(filePath);
-        await vscode.window.showTextDocument(doc, { preview: openFileDetails.preview });
+        try {
+            const doc = await vscode.workspace.openTextDocument(filePath);
+            await vscode.window.showTextDocument(doc, { preview: openFileDetails.preview });
+        } catch {
+            void this.logger.writeAndShowWarning(`File to open not found: ${filePath}`);
+            return EditorOperationResponse.Failed;
+        }
         return EditorOperationResponse.Completed;
     }
 
     private async closeFile(filePath: string): Promise<EditorOperationResponse> {
         filePath = await this.resolveFilePathWithCwd(filePath);
-        const doc = vscode.workspace.textDocuments.find((x) => x.fileName === filePath);
+        const doc = vscode.workspace.textDocuments.find((x) => x.uri.fsPath === filePath);
         if (doc != undefined && !doc.isClosed) {
             await vscode.window.showTextDocument(doc);
             await vscode.commands.executeCommand("workbench.action.closeActiveEditor");
+            return EditorOperationResponse.Completed;
         }
-        return EditorOperationResponse.Completed;
+        void this.logger.writeAndShowWarning(`File to close not found or already closed: ${filePath}`);
+        return EditorOperationResponse.Failed;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


andyleejordan:
@SeeminglyScience I couldn't find a way to test `CloseFile`...since while it's all plumbed through, I don't see us exposing it in any API.

src/features/ExtensionCommands.ts
```diff
@@ -161,11 +161,8 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
 
             vscode.commands.registerCommand("PowerShell.InvokeRegisteredEditorCommand",
                 async (param: IInvokeRegisteredEditorCommandParameter) => {
-                    if (this.extensionCommands.length === 0) {
-                        return;
-                    }
-
-                    const commandToExecute = this.extensionCommands.find((x) => x.name === param.commandName);
+                    const commandToExecute = this.extensionCommands.find(
```
andyleejordan, line 164: Just removed a pointless length check. It's not undefined so `.find` is just going to return nothing, which is later checked anyway.

src/features/ExtensionCommands.ts
```diff
@@ -364,15 +362,16 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
     }
 
     private async openFile(openFileDetails: IOpenFileDetails): Promise<EditorOperationResponse> {
-        const filePath = await this.normalizeFilePath(openFileDetails.filePath);
+        const filePath = await this.resolveFilePathWithCwd(openFileDetails.filePath);
```
andyleejordan, line 365: Renamed this because we're not doing normalization, we're just resolving to the CWD if we have to.

src/features/ExtensionCommands.ts
```diff
@@ -364,15 +362,16 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
...
         await vscode.window.showTextDocument(doc, { preview: openFileDetails.preview });
         return EditorOperationResponse.Completed;
     }
 
     private async closeFile(filePath: string): Promise<EditorOperationResponse> {
-        if (this.findTextDocument(await this.normalizeFilePath(filePath))) {
-            const doc = await vscode.workspace.openTextDocument(filePath);
+        filePath = await this.resolveFilePathWithCwd(filePath);
+        const doc = vscode.workspace.textDocuments.find((x) => x.fileName === filePath);
+        if (doc != undefined && !doc.isClosed) {
```
andyleejordan, line 374: I can't figure out how to test this function!

src/session.ts
```diff
@@ -444,10 +444,10 @@ export class SessionManager implements Middleware {
...
         // Detect any setting changes that would affect the session.
         if (!this.suppressRestartPrompt && this.sessionStatus === SessionStatus.Running &&
-            (settings.cwd.toLowerCase() !== this.sessionSettings.cwd.toLowerCase()
-                || settings.powerShellDefaultVersion.toLowerCase() !== this.sessionSettings.powerShellDefaultVersion.toLowerCase()
-                || settings.developer.editorServicesLogLevel.toLowerCase() !== this.sessionSettings.developer.editorServicesLogLevel.toLowerCase()
-                || settings.developer.bundledModulesPath.toLowerCase() !== this.sessionSettings.developer.bundledModulesPath.toLowerCase()
+            (settings.cwd !== this.sessionSettings.cwd
+                || settings.powerShellDefaultVersion !== this.sessionSettings.powerShellDefaultVersion
+                || settings.developer.editorServicesLogLevel !== this.sessionSettings.developer.editorServicesLogLevel
+                || settings.developer.bundledModulesPath !== this.sessionSettings.developer.bundledModulesPath
```
andyleejordan, line 450: Similarly we shouldn't be lower casing for comparisons in general. If it changed, it changed.

src/session.ts
```diff
@@ -894,7 +896,8 @@ Type 'help' to get help.
             action: async (): Promise<void> => {
                 await vscode.env.openExternal(
                     vscode.Uri.parse("https://dotnet.microsoft.com/en-us/download/dotnet-framework"));
-            }}]
+            }
+        }]
```
andyleejordan, line 900: Interesting...saving the file re-ran the formatter (which has had dependent packages updated...) so it looks like it now wants to format differently.

src/features/ExtensionCommands.ts
```diff
@@ -384,142 +383,88 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
...
+    // Resolve file path against user's CWD setting
+    private async resolveFilePathWithCwd(filePath: string): Promise<string> {
+        if (!path.isAbsolute(filePath)) {
+            const cwd = await validateCwdSetting(this.logger);
+            return path.resolve(cwd, filePath);
         }
-    }
-
-    private findTextDocument(filePath: string): boolean {
-        // since Windows and macOS are case-insensitive, we need to normalize them differently
```
andyleejordan, line 509: Nope! And we didn't need any of this code anway.

src/features/ExtensionCommands.ts
```diff
@@ -384,142 +383,88 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
...
+     * @param filePath the absolute path to save the document contents to
      */
-    private async saveDocumentContentToAbsolutePath(
-        documentUri: vscode.Uri,
-        destinationAbsolutePath: string): Promise<void> {
-        // Retrieve the text out of the current document
-        const oldDocument = await vscode.workspace.openTextDocument(documentUri);
-
-        // Write it to the new document path
+    private async saveFileAs(doc: vscode.TextDocument, filePath: string): Promise<void> {
```
andyleejordan, line 443: I simplified this to take the doc we already opened.

src/features/ExtensionCommands.ts
```diff
@@ -384,142 +383,88 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
...
+            if (!path.isAbsolute(newFilePath)) {
+                newFilePath = path.join(path.dirname(currentFileUri.fsPath), newFilePath);
             }
-            break; }
-
-        case "untitled": {
+        } else if (currentFileUri.scheme === "untitled") {
             // We need a new name to save an untitled file
-            if (!saveFileDetails.newPath) {
-                // TODO: Create a class handle vscode warnings and errors so we can warn easily
```
andyleejordan, line 425: Meh no we don't just log it anyway it's fine.

src/features/ExtensionCommands.ts
```diff
@@ -384,142 +383,88 @@ export class ExtensionCommandsFeature extends LanguageClientConsumer {
...
             }
 
-            // Make sure we have an absolute path
-            if (path.isAbsolute(saveFileDetails.newPath)) {
-                newFileAbsolutePath = saveFileDetails.newPath;
-            } else {
-                // If not, interpret the path as relative to the current file
-                newFileAbsolutePath = path.join(path.dirname(currentFileUri.fsPath), saveFileDetails.newPath);
+            // Special case where we interpret a path as relative to the current
+            // file, not the CWD!
```
andyleejordan, line 412: I didn't want to change this behavior.

src/features/ExtensionCommands.ts
```diff
@@ -69,9 +68,11 @@ export const GetEditorContextRequestType =
 export interface IGetEditorContextRequestArguments {
 }
 
+// NOTE: The server at least now expects this response, but it's used in any
```
SeeminglyScience, line 71:
```suggestion
// NOTE: The server at least now expects this response, but it's not used in any
```

right?

SeeminglyScience (approved):
LGTM! One minor possible typo correction in a comment

src/features/ExtensionCommands.ts
andyleejordan: Right!

PR #4707, opened 2023-08-20
Move `--sourcemap` to scripts instead of `Invoke-Build`

JustinGrote:
## PR Summary

ESBuild does not currently generate sourcemaps for debugging. This enables that generation and I verified that `vsce` does not include the sourcemaps when packaging.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
   "main": "./out/main.js",
   "scripts": {
     "lint": "eslint . --ext .ts",
-    "build": "esbuild ./src/main.ts --outdir=out --bundle --external:vscode --platform=node",
+    "build": "esbuild ./src/main.ts --outdir=out --sourcemap --bundle --external:vscode --platform=node",
     "build-watch": "npm run build -- --watch",
     "build-test": "tsc --incremental",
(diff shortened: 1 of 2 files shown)


andyleejordan:
Isn't https://github.com/evanw/esbuild/issues/944 an issue?

https://github.com/PowerShell/vscode-powershell/blob/bf548708fa0d8359a844759ddc0546963aac35ab/vscode-powershell.build.ps1#L97-L107

andyleejordan:
The build script adds `--sourcemap` on debug and `--minify` on release.

JustinGrote:
I'm using build watch and not the build script, that's probably why, it happens with "run extension" but not run tests because the file that it uses to run is different for both. Let me think about this tomorrow, I have an idea how to do it better.

andyleejordan:
@JustinGrote I gotchu.

JustinGrote:
@andyleejordan thanks! I was going to just put it in the watch npm script but as long as it doesn't hurt anything in main (it shouldn't since the tsc source maps shouldn't conflict) I'm good with it there, they take an insignificant amount of additional time to generate.

PR #4729, opened 2023-09-06
Add note about installing ESLint globally for VS Code extension

andyleejordan:
And do some documentation clean up while we're at it.


SeeminglyScience (approved):
LGTM!

PR #4780, opened 2023-10-10
Stop hiding Activity Bar in ISE Mode

andyleejordan:
The user can, if they really wish, do this with
`workbench.action.toggleActivityBarVisibility`, but at least in the latest Insiders it does not appear we can set this to `false` any more. Moreover, it's a confusing UI change to make even for the sake of an "ISE-like" experience. Simplest answer seems to be to stop trying to hide it so as to unblock our CI.


SeeminglyScience (approved):
LGTM!

Issue #4843, opened 2023-11-20
Cannot attach debugger to PS 7.4.0 processes


nightroman:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [x] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

Cannot attach debugger to an interactively selected PS 7.4.0 processes.
This operation takes long time, then results in this error

![image](https://github.com/PowerShell/vscode-powershell/assets/927533/546171f3-f963-4f95-9757-b3bd77c03cc7)

### PowerShell Version

```console
PS C:\ROM\SET\VSCode\debug-psf> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.0
PSEdition                      Core
GitCommitId                    7.4.0
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.84.2
1a5daa3a0231a0fbba4f14db7ec463cf99d7768e
x64
```

### Extension Version

```console
ms-vscode.powershell@2023.8.0
I have also tried the latest pre-release, same issue.
```

### Steps to Reproduce

- Select launch profile "PowerShell attach" (below)
- Hit F5
- Select one of PS 7.4.0 choices

Launch profile:

```json
{
  "launch": {
    "version": "0.2.0",
    "configurations": [
      {
        "type": "PowerShell",
        "request": "attach",
        "name": "PowerShell attach",
        "processId": "${command:PickPSHostProcess}",
        "runspaceId": 1
      }
    ]
  }
}
```

nightroman:
Bump...

[...]

Issue #4699, opened 2023-08-14
Powershell extension has not finished starting up yet. Please try again...


TESTER-sec:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

When launching the PowerShell: Command Explorer, a notice appears every single time in the lower right corner and remains there until the user manually closes it...

"Powershell extension has not finished starting up yet. Please try again..."

This is a very long-standing issue that you people have never fixed. It has been reported countless times over the years. So maybe one of you should take us users who take the time to report this recurring stuff ad nauseum seriously and definitely solve this issue permanently once and for all? Hmmm?

If someone does not like my tone, well I'm not going to apologize for it. The team responsible for VSCode PowerShell extension has a very long tradition of dismissing user reports and not fixing stuff. What credibility it once had has long since been wasted due to the long list of things never fixed.

<img width="960" alt="Untitled" src="https://github.com/PowerShell/vscode-powershell/assets/97270395/0b2ca896-ee4a-4786-b36e-db6ebc9b5726">

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.3.6
PSEdition                      Core
GitCommitId                    7.3.6
OS                             Microsoft Windows 10.0.22621
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
Version: 1.81.1 (system setup)
Commit: 6c3e3dba23e8fadc360aed75ce363ba185c49794
Date: 2023-08-09T22:22:42.175Z
Electron: 22.3.18
ElectronBuildId: 22689846
[...]

PR #4888, opened 2024-01-13
Fix up debugger configuration resolvers 

andyleejordan:
Wow this was a mess. Needs https://github.com/PowerShell/PowerShellEditorServices/pull/2130, and I'm a little worried about breaking changes _but_ supplying `"current"` to `processId` is special-cased to still work (well, throw an appropriate warning), since the launch configuration types only generate warnings.

Ok so we removed the notion of IDs (which are integers) being strings because that broke serialization with recent updates. But on inspection we realized the whole point was to be able to say "current" and have the attach configuration attach to the currently running Extension Terminal. Except that this was an unsupported, half-baked scenario (sure, it attached, but it couldn't actually be debugged properly). So now that throws warnings in both the client and server. The defaults for `processId` and `runspaceId` where changed to `null` (again, now they're just integers) which means to query for them. The supplied but half-baked configuration was removed.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4843

### DIFF
--- a/package.json
+++ b/package.json
@@ -486,18 +486,7 @@
             "body": {
               "name": "PowerShell Attach to Host Process",
               "type": "PowerShell",
-              "request": "attach",
-              "runspaceId": 1
-            }
-          },
-          {
-            "label": "PowerShell: Attach Interactive Session Runspace",
-            "description": "Open runspace picker to select runspace to attach debugger to",
-            "body": {
-              "name": "PowerShell Attach Interactive Session Runspace",
-              "type": "PowerShell",
-              "request": "attach",
-              "processId": "current"
+              "request": "attach"
             }
           },
           {
@@ -588,29 +577,27 @@
             "properties": {
               "computerName": {
                 "type": "string",
-                "description": "Optional: The computer name to which a remote session will be established.  Works only on PowerShell 4 and above."
+                "description": "Optional: The computer name to which a remote session will be established.",
+                "default": null
               },
               "processId": {
-                "type": "string",
-                "description": "The process id of the PowerShell host process to attach to.  Works only on PowerShell 5 and above.",
+                "type": "number",
+                "description": "Optional: The ID of the PowerShell host process that should be attached. Will prompt if unspecified.",
                 "default": null
               },
               "runspaceId": {
-                "type": [
-                  "string",
-                  "number"
-                ],
-                "description": "Optional: The ID of the runspace to debug in the attached process.  Defaults to 1.  Works only on PowerShell 5 and above.",
+                "type": "number",
+                "description": "Optional: The ID of the runspace to debug in the attached process. Will prompt if unspecified.",
                 "default": null
               },
               "runspaceName": {
                 "type": "string",
-                "description": "Optional: The Name of the runspace to debug in the attached process.  Works only on PowerShell 5 and above.",
+                "description": "Optional: The name of the runspace to debug in the attached process.",
                 "default": null
               },
               "customPipeName": {
                 "type": "string",
-                "description": "The custom pipe name of the PowerShell host process to attach to.  Works only on PowerShell 6.2 and above.",
+                "description": "The custom pipe name of the PowerShell host process to attach to.",
                 "default": null
               }
             }
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -87,7 +87,6 @@ export const DebugConfigurations: Record<DebugConfig, DebugConfiguration> = {
         name: "PowerShell: Attach to PowerShell Host Process",
         type: "PowerShell",
         request: "attach",
-        runspaceId: 1,
     },
     [DebugConfig.RunPester]: {
         name: "PowerShell: Run Pester Tests",
@@ -95,7 +94,7 @@ export const DebugConfigurations: Record<DebugConfig, DebugConfiguration> = {
         request: "launch",
         script: "Invoke-Pester",
         createTemporaryIntegratedConsole: true,
-        attachDotnetDebugger: true
+        attachDotnetDebugger: true,
     },
     [DebugConfig.ModuleInteractiveSession]: {
         name: "PowerShell: Module Interactive Session",
@@ -109,7 +108,7 @@ export const DebugConfigurations: Record<DebugConfig, DebugConfiguration> = {
         request: "launch",
         script: "Enter command to import your binary module, for example: \"Import-Module -Force ${workspaceFolder}/path/to/module.psd1|dll\"",
         createTemporaryIntegratedConsole: true,
-        attachDotnetDebugger: true
+        attachDotnetDebugger: true,
     },
     [DebugConfig.BinaryModulePester]: {
         name: "PowerShell: Binary Module Pester Tests",
@@ -117,7 +116,7 @@ export const DebugConfigurations: Record<DebugConfig, DebugConfiguration> = {
         request: "launch",
         script: "Invoke-Pester",
         createTemporaryIntegratedConsole: true,
-        attachDotnetDebugger: true
+        attachDotnetDebugger: true,
     }
 };
 
@@ -131,7 +130,7 @@ export class DebugSessionFeature extends LanguageClientConsumer
 
     constructor(context: ExtensionContext, private sessionManager: SessionManager, private logger: ILogger) {
         super();
-        // This "activates" the debug adapter for use with  You can only do this once.
+        // This "activates" the debug adapter. You can only do this once.
         [
             DebugConfigurationProviderTriggerKind.Initial,
             DebugConfigurationProviderTriggerKind.Dynamic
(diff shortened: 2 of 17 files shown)


src/features/DebugSession.ts
```diff
@@ -648,9 +650,10 @@ export class PickPSHostProcessFeature extends LanguageClientConsumer {
             matchOnDescription: true,
             matchOnDetail: true,
         };
+
         const item = await window.showQuickPick(items, options);
 
-        return item ? item.pid : undefined;
```
andyleejordan, line 653: This was causing the bug because `pid` was a number, not a string and something made it no longer "just work" and instead be invalid and crash.

andyleejordan: The smaller fix I first tested was just to `ToString()` it in the `item` above...but honestly we should just fix this the right way.

JustinGrote: Makes sense

src/features/DebugSession.ts
```diff
@@ -480,6 +484,17 @@ export class DebugSessionFeature extends LanguageClientConsumer
...
         }
 
+        // NOTE: We don't support attaching to the Extension Terminal, even
+        // though in the past it looked like we did. The implementation was
+        // half-baked and left things unusable.
+        if (config.processId === "current" || config.processId === await this.sessionManager.getLanguageServerPid()) {
+            // TODO: When (if ever) it's supported, we need to convert 0 and the
+            // old notion of "current" to the actual process ID, like this:
+            // config.processId = await this.sessionManager.getLanguageServerPid();
+            void this.logger.writeAndShowError("Attaching to the PowerShell Extension terminal is not supported. Please use the 'PowerShell: Interactive Session' debug configuration instead.");
```
JustinGrote, line 494: A great UX you provided for probably noone using it at all :)

JustinGrote (approved):
LGTM

src/features/DebugSession.ts
andyleejordan: I sure hope that's the case. It's crazy to me that we've literally been shipping provided debug config that doesn't work, and never did!

src/features/DebugSession.ts
```diff
@@ -617,25 +630,24 @@ export class PickPSHostProcessFeature extends LanguageClientConsumer {
         }
     }
 
-    private async pickPSHostProcess(): Promise<string | undefined> {
+    private async pickPSHostProcess(): Promise<number | undefined> {
```
andyleejordan, line 633: @JustinGrote I went through and plumbed these _all_ to be `number`.

JustinGrote: Looks like build didn't trigger on force push for some demonic reason. Lets try your re-push

JustinGrote:
Still happening to me in a `code --profile-temp` with the extension installed from vsix. I checked the source zip and the JS looks to match the PR
```js
  async pickPSHostProcess() {
    await this.getLanguageClient();
    const items = [];
    const response = await this.languageClient?.sendRequest(GetPSHostProcessesRequestType, {});
    for (const process3 of response ?? []) {
      let windowTitle = "";
      if (process3.mainWindowTitle) {
        windowTitle = `, ${process3.mainWindowTitle}`;
      }
      items.push({
        label: process3.processName,
        description: `PID: ${process3.processId.toString()}${windowTitle}`,
        processId: process3.processId
      });
    }
```

Let me try it in extension debug

package.json
```diff
@@ -588,29 +577,26 @@
...
               "computerName": {
                 "type": "string",
-                "description": "Optional: The computer name to which a remote session will be established.  Works only on PowerShell 4 and above."
+                "description": "Optional: The computer name to which a remote session will be established."
               },
               "processId": {
-                "type": "string",
-                "description": "The process id of the PowerShell host process to attach to.  Works only on PowerShell 5 and above.",
+                "type": "number",
+                "description": "Optional: The ID of the PowerShell host process that should be attached.",
```
andyleejordan, line 584:
```suggestion
                "description": "Optional: The ID of the PowerShell host process that should be attached. Will prompt if not specified.",
```

package.json
```diff
@@ -588,29 +577,26 @@
...
                 "default": null
               },
               "runspaceId": {
-                "type": [
-                  "string",
-                  "number"
-                ],
-                "description": "Optional: The ID of the runspace to debug in the attached process.  Defaults to 1.  Works only on PowerShell 5 and above.",
+                "type": "number",
+                "description": "Optional: The ID of the runspace to debug in the attached process.",
```
andyleejordan, line 589:
```suggestion
                "description": "Optional: The ID of the runspace to debug in the attached process.  Will prompt if not specified.",
```

src/features/DebugSession.ts
```diff
@@ -617,25 +630,24 @@ export class PickPSHostProcessFeature extends LanguageClientConsumer {
...
+        // We need the language client in order to send the request.
+        await this.getLanguageClient();
+
         // Start with the current PowerShell process in the list.
```
andyleejordan, line 637:
```suggestion
        // Start with an empty list!
```

src/features/DebugSession.ts
```diff
@@ -734,28 +745,26 @@ export class PickRunspaceFeature extends LanguageClientConsumer {
...
-    private async pickRunspace(processId: string): Promise<string | undefined> {
+    private async pickRunspace(processId: number): Promise<number | undefined> {
+        // We need the language client in order to send the request.
+        await this.getLanguageClient();
+
         const response = await this.languageClient?.sendRequest(GetRunspaceRequestType, { processId });
         const items: IRunspaceItem[] = [];
         for (const runspace of response ?? []) {
-            // Skip default runspace
-            if ((runspace.id === 1 || runspace.name === "PSAttachRunspace")
```
andyleejordan, line 750: We do need to keep the PSAttachRunspace (it's the bridge, we can't debug it). -- Patrick

andyleejordan (commented):
From Code Review

andyleejordan:
Now also resolves the root of https://github.com/PowerShell/vscode-powershell/issues/4699.

src/features/ISECompatibility.ts
```diff
@@ -25,32 +25,32 @@ export class ISECompatibilityFeature implements vscode.Disposable {
         { path: "powershell.codeFolding", name: "showLastLine", value: false }
     ];
 
-    private _commandRegistrations: vscode.Disposable[] = [];
-    private _iseModeEnabled: boolean;
-    private _originalSettings: Record<string, boolean | string | undefined> = {};
+    private commands: vscode.Disposable[] = [];
```
JustinGrote, line 28: sweet, I'm all for getting rid of _ prefixes

src/languageClientConsumer.ts
```diff
@@ -1,28 +1,66 @@
...
 import { LanguageClient } from "vscode-languageclient/node";
 
 export abstract class LanguageClientConsumer {
+    private static languageClientPromise?: Promise<LanguageClient>;
+    private static getLanguageClientResolve?: (value: LanguageClient) => void;
 
-    private _languageClient: LanguageClient | undefined;
+    // Implementations of this class must override this method to register their
+    // handlers, as its called whenever the client is restarted / replaced.
+    public abstract onLanguageClientSet(languageClient: LanguageClient): void;
```
JustinGrote, line 13: It feels like maybe you wanted an EventEmitter here but this is fine.

src/languageClientConsumer.ts
```diff
@@ -1,28 +1,66 @@
...
 
-    public get languageClient(): LanguageClient | undefined {
-        if (!this._languageClient) {
-            // TODO: Plumb through the logger.
-            void window.showInformationMessage("PowerShell extension has not finished starting up yet. Please try again in a few moments.");
-        }
-        return this._languageClient;
+    // We should have a promise as defined in resetLanguageClient, but if we
+    // don't, create it.
+    public static async getLanguageClient(): Promise<LanguageClient> {
```
JustinGrote, line 32: I feel like this would be more suited to a lazy getter than encapsulates all the other logic and returns a `Promise<LanguageClient>`. That way derived Consumers can just reference the property and the first one to do so triggers a language client startup e.g. `await this.languageClient` or `await this.languageClientPromise` (I prefer the former)

JustinGrote: That said this is totally fine too

JustinGrote: After further review I see this class is basically using languageClient as a singleton. If that's the case then why does everything have to inherit/consume from this class vs. it just be a standalone singleton class that has getLanguageClient as a static async method, and a onLanguageClientStarted EventEmitter (using either the vscode flavor or the nodeJS flavor) that clients can register to? That would enable them to be decoupled from this base class and be whatever they want (or have this functionality put in an interface)

JustinGrote: https://vscode-docs.readthedocs.io/en/stable/extensions/patterns-and-principles/#events
https://refactoring.guru/design-patterns/singleton/typescript/example

src/features/ShowHelp.ts
```diff
@@ -4,6 +4,7 @@
 import vscode = require("vscode");
 import { NotificationType } from "vscode-languageclient";
 import { LanguageClientConsumer } from "../languageClientConsumer";
+import { LanguageClient } from "vscode-languageclient/node";
```
JustinGrote, line 7: I think this was an autocomplete import oopsie

JustinGrote, line 7: Should be `type LanguageClient` to be clear you're only using this for type validation and not actually instantiating it

andyleejordan: TIL

Issue #4894, opened 2024-01-18
Update $env:TF_BUILD to $env:CI after Github Action Switch


JustinGrote:
This Var doesn't exist in Github Actions, it should be $CI
https://docs.github.com/en/actions/learn-github-actions/variables

https://github.com/PowerShell/vscode-powershell/blob/cbe8b7bd3f131e7414f973c38bd3d1294c3382e9/vscode-powershell.build.ps1#L29-L33

Also in PSES
https://github.com/PowerShell/PowerShellEditorServices/blob/70ef45d1655a7e5d37441095e1e5ab37408fe543/PowerShellEditorServices.build.ps1#L107-L108

Probably not causing a failure now but could pull in an unwanted dependency update that breaks the build in difficult to troubleshoot ways.

@andyleejordan please confirm I'm not missing something here.

andyleejordan:
Ah so the only thing "missing" is that the release pipeline, where not bringing in a new dependency matters most, is still ADO with this variable. It should instead be `$env:TF_BUILD -or $env:CI`.

JustinGrote:
There are other vars in there like the build or PR determination that need to be updated to the GH equivalents as well

PR #4916, opened 2024-02-13
Fix use of Azure NPM registry

andyleejordan:
Figured out how to authenticate it because just temporarily adding it doesn't satisfy requirements.


andyleejordan:
Also resolves https://github.com/PowerShell/vscode-powershell/issues/4894 for this repo I think.

SeeminglyScience (approved):
LGTM!

PR #4920, opened 2024-02-15
Update Azure NPM secret names

andyleejordan:
Turns out Dependabot-triggered Actions use Dependabot secrets, not Actions secrets. In the Action, the password must be base64 encoded, but in the Dependabot configuration it must not, so we need three secrets:

* GitHub Action secret: base64 encoded password
* Dependabot secret: base64 encoded password
* Dependabot secret: plain text password

There's no reasonable way to base64 encode the secret on-the-fly.

See: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions


SeeminglyScience (approved):
LGTM!

Issue #4932, opened 2024-02-28
Command Explorer should be off by default to improve startup speed


SydneyhSmith:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Summary

It should be on in ISE mode but otherwise off by default to improve startup

PR #4934, opened 2024-02-28
Don't make the Command Explorer visible by default

andyleejordan:
But turn it on in ISE mode.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4932.


SeeminglyScience (approved):
LGTM!

Issue #261, opened 2016-08-25
"Smart" variable rename


adbertram:
When I rename a variable in one place, I want to ensure all references to that variable are renamed that are in the same scope. Basically, I want to ensure when a rename a variable that anything that might be depending on that name is renamed as well in the same script.  Here's an example:

http://www.screencast.com/t/ggUIoKqSC

Bill-Stewart:
I think you can do this already by 1) selecting the variable, 2) use the command `editor.action.selectHighlights` (default key mapping is <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd>), then 3) type the replacement name.

sgtoj:
@Bill-Stewart You're some what correct. @adbertram is referring to the **Rename Symbol** (F2) command. It will smartly update the symbol (e.g. variable, class, functions, etc.) versus a blind change of all occurrence of a word. 

```
# get some objects to work with
$something    = (Get-Something);
$anotherthing = (Get-AnotherThing);

# set some value from the first object
$someValue1  = $something.SomeValue;
$someValue2  = $something.SomeValue;

# set another one from the other object
$somevalue3  = $anotherthing.SomeValue;
```

With Symbol Rename, the editor will understand that both occurrence `SomeValue` from `$something.SomeValue` are the same but the one from `$anotherthing.SomeValue` is not. So the Symbol Rename command of `SomeValue` property on the `$something` object will only make two changes even though there are at least 3 occurrence of "somevalue" (actually there are 6). 
With Symbol Rename, the editor will understand that both occurrence of `$something.SomeValue` are the same but not `$anotherthing.SomeValue`. Symbol Rename command of `SomeValue` property on the `$something` object will only make two changes even though there are at least 3 occurrence of "somevalue" (actually there are 6).

Bill-Stewart:
Well, I interpreted the OP's request as variables rather than object members. I wouldn't mind this if performance is good. (As noted, we can already change variables pretty easily.)

mattmcnabb:
Plus one on the request from @adbertram. The key here is that the token rename should be scope-aware in PowerShell. We've been doing this in ISESteroids for a while:

![image](https://cloud.githubusercontent.com/assets/6998507/20901627/5975974a-bb01-11e6-870a-ed543d3a68c4.png)

rkeithhill:
[...]

PR #4950, opened 2024-03-25
WIP: Implementing Client side Rename provider to allow variable/function rename

Razmo99:
## PR Summary

This pull request implements the VSCode `RenameProvider` class along with two methods within it `provideRenameEdits` and `prepareRename` which is requested in in #261 

This will allow for the client to make requests to the server side of the extension when pressing F2 to perform refactoring on variables and functions.

This client side code has been modified from the C# VSCode extension and modified to work with the Powershell extension's server side component.

There will be a sister pull request on the PowershellEditorServices repository to implement the server side.

## PR Checklist

- [X] Summarized changes
- [NA] PR has tests


Razmo99:
@microsoft-github-policy-service agree

SydneyhSmith:
Hi @Razmo99 super appreciate the detailed PR! Just want to give you a heads up that the maintainers are all pretty heads down on commitments at the moment so it may be a couple of weeks before we get a chance to review the PR

silverqx:
I would like to ask, what is the state of this PR, does it work actually? I fastly scrolled through the proposed PR and also PowerShellEditorServices#2152, looks like @Razmo99 did a great job about `registerRenameProvider` and worked ~1 year on this.

Would it be possible to review it, git checkout it, test it, and release it in the next version? I lack renaming in PS extension so much, would greatly boost dev. productivity in many cases. I know there always is Search & Replace, but properly implemented rename refactorings are gold in some cases 🙂 To just hit ctrl+shift+r or f2 and rename the symbol under the cursor, to avoid classic search and replace.

I know that the PR is huge, but also the functionality is amazing.

andyleejordan:
I would really like to get this feature in, yes. However, we're currently paused on feature work for a bit while some security work takes priority. Said another way: please expect to see it within the year, but probably not in the next month.

silverqx:
Ok, thx for stating it out, would be nice to make it a little quicker but whatever, at least check-in it out and test it to see if it works and in which state the PR is would be a good idea. I think that having a new feature like this is more interesting for developers than having npm packages at the latest versions every month 😁

JustinGrote:
For what it's worth I'm currently reviewing the PSES part of this to hopefully help reduce the friction when the opportunity arises for a more thorough review.

andyleejordan:
Justin's work will be very helpful and should speed up the process!

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export class RenameSymbolFeature extends LanguageClientConsumer implements RenameProvider {
+    public override onLanguageClientSet(_languageClient: LanguageClient): void {
+        throw new Error("Method not implemented.");
+    }
+    private command: vscode.Disposable;
+
+    constructor() {
+        super();
+        this.command = vscode.commands.registerCommand("PowerShell.RenameSymbol", () => {
```
JustinGrote, line 56: This can be removed, no need to register a command, no other LSPs really do this.

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+interface IPrepareRenameSymbolRequestResponse {
+    message : string
+}
+
+export const RenameSymbolRequestType = new RequestType<IRenameSymbolRequestArguments, IRenameSymbolRequestResponse, void>("powerShell/renameSymbol");
+export const PrepareRenameSymbolRequestType = new RequestType<IPrepareRenameSymbolRequestArguments, IPrepareRenameSymbolRequestResponse, void>("powerShell/PrepareRenameSymbol");
+
+export class RenameSymbolFeature extends LanguageClientConsumer implements RenameProvider {
+    public override onLanguageClientSet(_languageClient: LanguageClient): void {
+        throw new Error("Method not implemented.");
```
JustinGrote, line 50: Was there something that needs to be registered here? Otherwise this is just throwing all the time because I'm pretty sure this gets called on startup.

src/main.ts
```diff
@@ -57,6 +59,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<IPower
     const settings = getSettings();
     logger.writeVerbose(`Loaded settings:\n${JSON.stringify(settings, undefined, 2)}`);
 
+    const RenameSymbol = new RenameSymbolFeature();
+    languageRenameProvider = vscode.languages.registerRenameProvider(documentSelector,RenameSymbol);
```
JustinGrote, line 63: This should be structured under the languageClientConsumers below and implemented as a feature. I know it is a bit obtuse but thats how the extension was originally architected and it enables features to be more pluggable.

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+                return Promise.reject(response.message);
+
+            }
+
+            return {
+                range: wordRange,
+                placeholder: wordText
+            };
+        }catch (error) {
+            return null;
```
JustinGrote, line 128: Probably shouldn't swallow errors, these should at least be surfaced in the logs if not returned to the user outright in an errormessage

JustinGrote: Or a promise rejection

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+        try {
+            const client = await LanguageClientConsumer.getLanguageClient();
+            const response = await client.sendRequest(PrepareRenameSymbolRequestType, req);
+
+            if (!response.message) {
+                return null;
+            }
+            const wordRange = document.getWordRangeAtPosition(position);
+            if (!wordRange) {
+                return Promise.reject("Not a valid location for renaming.");
```
JustinGrote, line 114: Nit: Since this is an async function, why not just throw? should have the same result and is more idiomatic.

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface IRenameSymbolRequestResponse {
+    changes : ModifiedFileResponse[]
+}
+
+interface IPrepareRenameSymbolRequestResponse {
+    message : string
+}
+
+export const RenameSymbolRequestType = new RequestType<IRenameSymbolRequestArguments, IRenameSymbolRequestResponse, void>("powerShell/renameSymbol");
```
JustinGrote, line 45: nit: these probably don't need to be exported, they aren't referenced elsewhere.

src/features/RenameSymbol.ts
```diff
@@ -0,0 +1,132 @@
...
+
+            if (!response.changes.length) {
+                return undefined;
+            }
+
+            const edit = new WorkspaceEdit();
+            response.changes.forEach(change => {
+                const uri = Uri.file(change.fileName);
+
+                change.changes.forEach(change => {
```
JustinGrote, line 85: Style nit: Should probably use "for ... of" where these foreaches are, as they support more flow controls and is what is typically used throughout the codebase.

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

src/features/RenameSymbol.ts
JustinGrote: Fixed, thanks!

JustinGrote:
Here is a build of the latest PSES rename changes. Lots of progress to go but it's here to test out:
[powershell-2024.6.0-renameAlpha.vsix.zip](https://github.com/user-attachments/files/15594148/powershell-2024.6.0-renameAlpha.vsix.zip)

You'll need to rename the .vsix.zip to just .vsix and then [install using VSIX locally with vscode](https://code.visualstudio.com/docs/editor/extension-marketplace#_install-from-a-vsix).

JustinGrote:
Caught here too: 

```pwsh
$NeededTools = @{
    OpenSsl       = 'openssl for macOS'
    PowerShellGet = 'PowerShellGet latest'
    InvokeBuild   = 'InvokeBuild latest'
}

function getMissingTools () {
    $missingTools = @()

    if (needsOpenSsl) {
        $missingTools += $NeededTools.OpenSsl
    }
    if (needsPowerShellGet) {
        $missingTools += $NeededTools.PowerShellGet
    }
    if (needsInvokeBuild) {
        $missingTools += $NeededTools.InvokeBuild
    }

    return $missingTools
}
```

If you try to rename $NeededTools inside the function, it only renames the external reference but not the internal ones. Vice-versa does not work, if we support one we should support the other.

Maybe this should be a configurable option? Or maybe we require SCRIPT or GLOBAL scope references in order for rename of variables referenced outside functions to work?

andyleejordan:
Hi! We discussed this in triage and our main condition on accepting it is that it's an opt-in feature controlled by a setting that ideally has a good explanation of when and how it can fail (due to PowerShell scoping). As long as it's opt-in most of our concerns don't apply.

JustinGrote:
> Hi! We discussed this in triage and our main condition on accepting it is that it's an opt-in feature controlled by a setting that ideally has a good explanation of when and how it can fail (due to PowerShell scoping). As long as it's opt-in most of our concerns don't apply.

Right, I was thinking more specifically for this handling, but I plan to assist it in such a way that it will operate as such:

First time you attempt F2 rename, a popup will appear with a link to Markdown docs we will have in the docs section explaining about PowerShell being a dynamic language, and the limitations and at-your-own-risk aspects. You will then have to click an "Accept", "Accept for Workspace", etc. option that will drop a setting in the appropriate location to suppress the dialog on future operations.

package.json
```diff
@@ -1016,6 +1016,11 @@
           ],
           "default": "off",
           "description": "Traces the communication between VS Code and the PowerShell Editor Services language server. **This setting is only meant for extension developers!**"
+        },
+        "powershell.renameSymbol.shouldGenerateAlias": {
```
JustinGrote, line 1020: Just from a UX perspective, since this is user facing it can probably be simplified to `powershell.rename.createAlias`

JustinGrote:
@razmo99 this is great work, FYI, I will be PTO June 14-28 in Europe so I may be limited in further reviews until then but I will probably show this off at PSConfEU and give you full credit!

Razmo99:
@JustinGrotes, are you back in action from Europe? just wondering what are next steps are to move forward. Do you want me to implement the opt in mechanism on the client side?

JustinGrote:
@Razmo99 thanks for following up and all your hard work. I've had some life circumstances come up that are going to unfortunately minimize my involvement as a volunteer at the moment, I hope I'll be able to get involved again soon, but I think in the meantime as mentioned by @andyleejordan the opt-in triggered on first F2 rename, controlled by a setting is a good next step as well as preparing a markdown document that details the expectations (no multi-file for now, etc.) and the edge cases. He would ultimately be performing/approving the merge so his guidance would be paramount to the merge.

JustinGrote:
I added the disclaimer, at least the base scaffolding of it. Not super-happy about the display format, I would like to do markdown preview, but I would have to have it as a button instead and recycle the info message.

https://github.com/user-attachments/assets/c2928880-3273-48b8-ad24-896eb4e3243c

JustinGrote:
Force pushed a rebase to deal with the silly not-really-merge-conflicts.

JustinGrote:
@Razmo99 since I've had some time to dive into this, is there a particular reason you made a custom PowerShell/prepareRename rather than using the standard TextDocument/prepareRename that the LSP uses? Since we don't get the full documentUri, that helps explain why unsaved/untitled files don't work right.

Razmo99:
I didn't know there was a standard prepearRename provider 😅 I made it the way I did as I wanted a separate endpoint to send rename requests to the language server and handle them differently rather them complicating the main rename symbol endpoint

ALIENQuake:
Great feature, one suggestion, the initial variable name should not contain `$` in order to not require repeating this char when providing new name.

JustinGrote:
> Great feature, one suggestion, the initial variable name should not contain `$` in order to not require repeating this char when providing new name.

There is logic in the latest commits where it doesn't matter if you have the $ or not, it'll get fixed automatically, and it has checking to make sure it's a valid variable name (e.g. no spaces unless theres a {} around the name, etc.)

Razmo99:
> I'm closing this PR in favor of #5056 since the scope has drastically changed. @Razmo99 you should get co-author credit on the PR, if not we'll make sure it ends up in the release notes for sure.

@JustinGrote no problem at all, honestly I'm just happy to see this finally get implemented as a feature of this plugin!

PR #4962, opened 2024-04-04
Add a NuGet.config file to the BinaryModule mock

andyleejordan:
So that Cloud Feed Services is happy.


SeeminglyScience (approved):
LGTM!

PR #4971, opened 2024-04-09
Check if the Terminal Shell Integration setting is changed

andyleejordan:
Because if it is, we'll need to restart the Extension Terminal (unless it was hidden at startup).
Also add a test to ensure we can find VS Code's script (in case its location changes upstream).


SeeminglyScience (approved):
LGTM!

PR #5036, opened 2024-09-09
Add a setting to launch scripts with the Call operator instead of DotSource - Fixes #4327

LucasArona:
Add support to switch the ExecuteMode from DotSource to Call

## PR Summary

This PR adds a new setting to allow the user to configure how scripts are launched, either through the Dot-Source operator (default, current behavior), or through the Call operator.

image

I'll also create a PR on PowerShellEditorServices to implement the server part.

## PR Checklist

Note: Tick the boxes below that apply to this pull request by putting an `x` between the square brackets.
Please mark anything not applicable to this PR `NA`.

- [X] PR has a meaningful title
- [X] Summarized changes
- [NA] PR has tests
- [X] This PR is ready to merge and is not work in progress
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready

PR #5059, opened 2024-10-07
Add setting to use either `.` or `&` when launching script

andyleejordan:
Replaces #5036, thanks @LucasArona. I was trying just to edit your PR and accidentally closed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -462,7 +462,7 @@
               "name": "PowerShell Launch Script",
               "type": "PowerShell",
               "request": "launch",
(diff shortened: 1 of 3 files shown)


andyleejordan:
I've tested this with https://github.com/PowerShell/PowerShellEditorServices/pull/2177 locally and it works like a charm. I'm trying to come up with an appropriate unit test.

SeeminglyScience (approved):
LGTM!

PR #5071, opened 2024-10-29
Refactor: Replace package.json requires with extension context info

JustinGrote:
## PR Summary
The package.json info is now available in the extension context, this avoids a round trip. Note there is no validation on the import of the information (there wasn't before either), but it's very unlikely to surface a null or undefined, or have that affect operation of the extension (famous last words).

## PR Checklist

- [x] Summarized changes
- [NA] PR has tests

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,10 +26,6 @@ import { LogLevel, getSettings } from "./settings";
 import { PowerShellLanguageId } from "./utils";
 import { LanguageClientConsumer } from "./languageClientConsumer";
 
-// The most reliable way to get the name and version of the current extension.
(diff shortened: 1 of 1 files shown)


andyleejordan (approved):
Love it, thanks!

JustinGrote:
Quick note that this API is also available in 1.82 which is what Azure Data Studio is on so that should be fine but I haven't directly tested it.
https://github.com/microsoft/vscode/blob/release/1.82/src/vscode-dts/vscode.d.ts#L7077

andyleejordan:
@JustinGrote we no longer have to support ADS and I bumped our engine to 1.94!

PR #5099, opened 2024-11-20
Improve warning when additional PowerShell isn't found

andyleejordan:
Since we added logic which searches for possible intended permutations of the given additional PowerShell path, we needed to make the warning show only if none of the permutations were found. This was accomplished by suppressing it in the first iterator and then yielding it again after the permutations were exhausted with the warning unsuppressed.

@JustinGrote this bug just _really_ annoyed me...I'm surprised no one reported it yet.

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -239,17 +239,17 @@ export class PowerShellExeFinder {
                 }
 
                 exePath = untildify(exePath);
-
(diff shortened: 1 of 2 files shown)


JustinGrote (approved):
LGTM and tests pass on my machine.

PR #5103, opened 2024-11-26
Implement Extension Setting Categories

JustinGrote:
## PR Summary

We have a lot of settings, so this organizes them into categories using the contributes sections feature of vscode. I applied some subjective "most commonly used" logic to the order of the categories and settings within the categories.

image

I did a check and it doesn't look like I missed any settings, they appear to all be there.
image
(what may not be clear is I switched branches between the two collections of package.json...)

### DIFF
--- a/package.json
+++ b/package.json
@@ -601,437 +601,464 @@
         "initialConfigurations": []
       }
     ],
-    "configuration": {
-      "title": "PowerShell",
-      "properties": {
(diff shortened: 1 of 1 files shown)


andyleejordan (approved):
Thanks, I love it. I used your same PowerShell snippet in your screenshot but ran it through `Compare-Object`, I think you're good! I tested this out and it's great.

PR #5149, opened 2025-02-26
Fix tests and bump packages

andyleejordan:
Fixes the unit test that needs to be bumped due to PowerShell 7.5 releasing. Fixes the OneBranch pipeline to rely on built-in PSResourceGet now that PowerShell 7.4  is (ironically) finally used. Bumps minor versions of packages automatically and carefully/manually bumps major versions.


SeeminglyScience (approved):
LGTM!

PR #5164, opened 2025-03-26
Teach client to find Homebrew installations on Apple Silicon

andyleejordan:
Since `/opt/homebrew` is now used instead of `/usr/local`.

This annoyed me when setting up a new MacBook.


andyleejordan:
Yet again bit by our platform tests sharing macOS and Linux.

SydneyhSmith (approved):

PR #5063, opened 2024-10-10
Migrate to DeployBox for release stage

andyleejordan:
TODO: Switch from PPE to Production. This requires OneBranch allow the AzureCLI task, used as [documented](https://devblogs.microsoft.com/devops/public-preview-of-workload-identity-federation-for-azure-pipelines/) for providing WIF credentials to the `vsce` tool when uploading to the marketplace.

PR #5190, opened 2025-05-15
Overhaul linting, formatting, and testing infrastructure

andyleejordan:
Replaces #5188 with a whole overhaul of our linting, formatting, and testing infrastructure.


SeeminglyScience (approved):
LGTM!

eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
+import eslint from "@eslint/js";
+import eslintConfigPrettier from "eslint-config-prettier/flat";
+import tseslint from "typescript-eslint";
+
+export default tseslint.config(
+    eslint.configs.recommended,
+    tseslint.configs.strictTypeChecked,
+    tseslint.configs.stylisticTypeChecked,
+    eslintConfigPrettier,
```
JustinGrote, line 9: We should probably see how much of this is redundant with the above now.

package.json
```diff
@@ -60,46 +60,41 @@
...
     "vscode-languageclient": "^9.0.1",
     "vscode-languageserver-protocol": "^3.17.5"
   },
   "devDependencies": {
-    "@vscode/vsce": "^3.3.0",
-    "esbuild": "^0.25.1"
+    "@vscode/vsce": "^3.3.2",
+    "esbuild": "^0.25.4"
   },
   "optionalDependencies": {
```
JustinGrote, line 75: We probably should document *somewhere* why optionalDependencies vs devdependencies

JustinGrote (approved):
LGTM! Nice work on the test fixes. Just some future notes.

package.json
andyleejordan: I'm really hoping it changes when I get back to trying #5063

eslint.config.mjs
andyleejordan: To my knowledge none of these are redundant to each other, but it was a bit ago that I read through those docs.

JustinGrote: Sorry, My callout was supposed to be on the {} with the custom rule definitions below it.

JustinGrote: I just went through them, all the custom rule overrides are still valid to https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/recommended-type-checked.ts

PR #5191, opened 2025-05-15
Prettier reformatting

andyleejordan:
I wonder if this will break GitHub.

### DIFF
--- a/.config/tsaoptions.json
+++ b/.config/tsaoptions.json
@@ -2,7 +2,7 @@
   "instanceUrl": "https://msazure.visualstudio.com",
   "projectName": "One",
   "areaPath": "One\\MGMT\\Compute\\Powershell\\Powershell",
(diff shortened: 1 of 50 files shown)


andyleejordan:
@SeeminglyScience @JustinGrote any files we _don't_ want the Prettier formatting applied to?

andyleejordan:
Ok the commit to add the rebase ignore file will need to come post-merge as all of GitHub rebase-merge and squash-merge both result in a new commit, and I don't want to do a merge-merge.

andyleejordan:
Was not worth running it across Markdown.

JustinGrote:
Only thing I can think is for files that will have their own vscode formatters, might end up in a back-and-forth fight between prettier and format-on-save, so we just have to keep an eye out for that in PRs after this.

SeeminglyScience (approved):

PR #5196, opened 2025-05-20
Fix tests on OneBranch

andyleejordan:
This cleans up our sleeps and our unit test settings plus handles the edge case of OneBranch's non-default PowerShell installation, allowing us to cut an entire stage out of the build.


SeeminglyScience (approved):
LGTM!

JustinGrote:
Late but LGTM as well

PR #5208, opened 2025-05-31
Feature: Add sessionName debug setting to allow different PS for temp console

JustinGrote:
## PR Summary

We want to encourage people to use PS7 for their extension terminal, but need to give them an easy way to test and debug PowerShell 5.1 tests. 

This adds a setting to enable the selection of a different PowerShell executable for the temp terminal. It also adds a launch config sample for Windows PowerShell.

https://github.com/user-attachments/assets/8d62c8a4-a5ee-4507-9552-a1cfb54ac463

This has the added benefit of being able to temp console other PowerShell exes, such as a latest PowerShell build.

### DIFF
--- a/src/features/DebugSession.ts
+++ b/src/features/DebugSession.ts
@@ -56,6 +56,7 @@ export enum DebugConfig {
     ModuleInteractiveSession,
     BinaryModule,
     BinaryModulePester,
+    WindowsPowerShell,
 }
 
 /** Make the implicit behavior of undefined and null in the debug api more explicit  */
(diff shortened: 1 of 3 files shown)


src/features/DebugSession.ts
```diff
@@ -440,6 +458,10 @@ export class DebugSessionFeature
             return PREVENT_DEBUG_START_AND_OPEN_DEBUGCONFIG;
         }
 
+        if (config.sessionName) {
+            config.createTemporaryIntegratedConsole = true;
```
andyleejordan, line 462: My only thought is that we tried to move away from "integrated console" to "extension terminal" but I am unsure how much of that name change made it to this particular feature.

andyleejordan (approved):

src/features/DebugSession.ts
JustinGrote: I can probably fix that up, stay tuned.

JustinGrote: So I can either fix it in this commit or we commit this and I submit another PR to rename/alias createTemporaryIntegratedConsole to createTemporaryPowershellSession. Which would you prefer?

JustinGrote: Seeing as this was merged I will address it in separate PR.

andyleejordan: Oh yeah I didn't see it was set to auto complete.

PR #5227, opened 2025-06-29
Un-register extension between steps to prevent cascading test failures

tsmarvin:
## PR Summary
- Add getRegisteredExtensions method to extension client to allow listing all registered extensions.
- Update ensureEditorServicesIsConnected test method to un-register the extension if its already registered to prevent cascading test failures.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # PowerShell Extension Release History
 
+## v2025.2.0
+### Monday, June 30, 2025
+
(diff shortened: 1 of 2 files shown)


tsmarvin:
Sort of. I added this so that it was easier for me to debug the actual root cause issue I was having.

I added it pretty early on in my testing so I'm not 100% confident on why I was seeing the cascading registration errors, but i suspect the PSES internal host not starting left the test waiting for the connection and it skipped over the de-registration.

If you don't think it's necessary I won't be offended, just figured id submit the PR as it helped me personally when debugging. It seemed like it was the expected state for the start of several tests related to the method I added it to.

PR #5228, opened 2025-06-29
Reduce scope of regex in PowerShellExeFinder

tsmarvin:
## PR Summary
- Update integer checking regex to match implied expectations from related comments.
  - While reading through the comments associated with the IntRegex value it is implied that we're only looking for single digit integer folders. This update formally matches that implied expectation.

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -78,7 +78,7 @@ export function getPlatformDetails(): IPlatformDetails {
  */
 export class PowerShellExeFinder {
     // This is required, since parseInt("7-preview") will return 7.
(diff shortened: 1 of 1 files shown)


andyleejordan:
Hi, is there a known bug that this code change addresses?

andyleejordan:
Al righty, without a reproducible bug I'm going to close this. All code changes introduce risk, so without something identifiable as "fixed" it's not worth introducing such changes. Thanks though!

PR #5249, opened 2025-07-28
Add notifyOnAttach option

jborean93:
## PR Summary
Adds the new `notifyOnAttach` option for an attach request. This option will have PSES create a new event with the source identifier `PSES.Attached` that allows the attached script to wait for the attach event.

Should wait until https://github.com/PowerShell/PowerShellEditorServices/pull/2250 is merged (or rejected).

## PR Checklist
- [x] Summarized changes
- [NA] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -610,6 +610,11 @@
                 "type": "string",
                 "description": "The custom pipe name of the PowerShell host process to attach to.",
                 "default": null
+              },
+              "notifyOnAttach": {
+                "type": "boolean",
+                "description": "Creates the event 'PSES.Attached' when the debugger attaches to the PowerShell host process. This is useful when creating a script that waits for the debugger to attach.",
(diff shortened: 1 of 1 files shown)


JustinGrote:
@jborean93 sorry, yes I am aware, this comment should have been on the actual implementation in PSES, sorry.

jborean93:
I'm not sure if PSES could do it dynamically. Maybe a source generator to generate the record from the `package.json` but IIRC source generators don't like IO work but I could be misremembering.

JustinGrote:
@jborean93 this PR is fine, my point was I was wondering if we could do away with the setting *altogether* in PSES and instead detect when the notify is needed based on the other launch options. If not that's totally fine, having it as a manual specification is not a problem, just thinking about user ergonomics.

Issue #5381, opened 2026-01-21
PowerShell Custom Attribute Types Unresolved on Initial File Open


mdaneri:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [x] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [x] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [x] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

## Summary

When opening a PowerShell script that uses **custom attributes defined in a .NET assembly**, the VS Code PowerShell extension reports diagnostics such as:

> Cannot find the type for custom attribute 'OpenApiSchemaComponent'. Make sure that the assembly that contains this type is loaded.

This happens **on initial file open only**. The error **disappears immediately after any edit** (even a single key press), indicating that a re-analysis fixes the issue.

This occurs **even when `using assembly` with an absolute path is present**, and even though the type is fully resolvable in the PowerShell terminal session.

---

## Environment

- OS: Windows
- VS Code: (fill in)
- PowerShell VS Code Extension: (fill in)
- PowerShell Version: (fill in, e.g. 7.x)
- Module: Kestrun
- Attribute Assembly: `Kestrun.Annotations.dll`

---

### Terminal Session (Correct)

In the integrated PowerShell terminal:

```powershell
Get-Module Kestrun
'OpenApiSchemaComponent' -as [type]
```

Result:

- Module is loaded.
- Attribute type resolves correctly from `Kestrun.Annotations.dll`.

### Editor Diagnostics (Incorrect)

On initial open, VS Code reports missing attribute types **despite**:

- Module being available
- Assembly existing
- Attribute being public and valid
- Runtime resolution succeeding

---

## Attempted Mitigations (All Failed)

- `RequiredAssemblies` in module manifest
- `using module Kestrun`
- `using assembly "<absolute path to Kestrun.Annotations.dll>"`
- Explicit `Import-Module Kestrun`
[...]

PR #5402, opened 2026-02-17
Refresh diagnostics for open PowerShell docs

mdaneri:
## PR Summary
- refresh analysis for already-open PowerShell documents once the language client reaches Running
- replay textDocument/didClose + textDocument/didOpen for open file and untitled PowerShell docs
- avoid requiring a manual edit to clear false custom-attribute type diagnostics on initial open

## Related Issue
Fixes PowerShell/vscode-powershell#5381

## Testing
- verified extension builds with npm run compile
- validated issue repro: initial false diagnostics clear without typing after session startup

## PR Checklist

- [x] Summarized changes
- [ ] PR has tests


andyleejordan (changes requested):
If I'm reading this right, you're essentially sending mock close and open notifications for PowerShell docs after the LSP server starts. Which makes sense to work around the issue, but since that means we've identified the root case (these notifications are being sent too early) can we investigate if there are VS Code / LSP Client APIs to use to delay sending those notifications until we're ready? What I'm saying is, I suspect there's a more correct way to fix this.

PR #5450, opened 2026-04-04
Fix incorrect parser errors from unloaded assemblies

### LINKED ISSUE #5381: PowerShell Custom Attribute Types Unresolved on Initial File Open

andyleejordan:
An attempt to fix #5381 where we believe the cause is that the didOpen()/didChange() notifications are being sent before PowerShell has totally finished loading.

Since these run the PowerShell parser on the OmniSharp thread pool, they essentially race the analysis of the files by PSScriptAnalyzer on the PowerShell runspace pool. So when they beat it, the return errors. But when they're beaten by it, the PSSA analysis has caused the assemblies (and so custom attributes) to be loaded, no longer erroring.

I posited we could gate the notifications instead of duplicating them like in #5402, and if the `Middleware` works as suspected by me (and Claude) this should fix it. Morever, we now also use a proper `Promise` instead of a while loop around a sleep to wait for the LSP server to be running.

While all of this could just be an AI hallucination...it seems right.

@mdaneri can you please test this against your scenario?

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -126,6 +126,10 @@ export class SessionManager implements Middleware {
     private versionDetails: IPowerShellVersionDetails | undefined;
     private traceLogLevelHandler?: vscode.Disposable;
 
+    // Promise-based gate resolved when the session reaches Running status.
+    // Used by waitUntilStarted() and the didOpen()/didChange() notifications.
+    private started = Promise.withResolvers<undefined>();
+
     constructor(
         private extensionContext: vscode.ExtensionContext,
         private sessionSettings: Settings,
@@ -295,6 +299,7 @@ export class SessionManager implements Middleware {
                 `Started PowerShell v${this.versionDetails.version}.`,
             );
             this.setSessionRunningStatus(); // Yay, we made it!
+            this.started.resolve(undefined); // Release didOpen()/didChange() notifications and waitUntilStarted() gate
 
             await this.writePidIfInDevMode(this.languageServerProcess);
 
@@ -328,6 +333,8 @@ export class SessionManager implements Middleware {
         }
 
         this.languageClient = undefined;
+        this.started.resolve(undefined);
+        this.started = Promise.withResolvers<undefined>();
 
         // Stop and dispose the PowerShell process(es).
         this.debugSessionProcess?.dispose();
@@ -497,9 +504,27 @@ export class SessionManager implements Middleware {
     }
 
     public async waitUntilStarted(): Promise<void> {
-        while (this.sessionStatus !== SessionStatus.Running) {
-            await utils.sleep(200);
-        }
+        await this.started.promise;
+    }
+
+    // Middleware hooks to delay document sync notifications until the server
+    // is fully initialized. This prevents stale parser diagnostics (e.g.
+    // unresolved custom attribute types) that would otherwise appear because
+    // textDocument/didOpen is sent before the server's type resolution is ready.
+    public async didOpen(
+        document: vscode.TextDocument,
+        next: (document: vscode.TextDocument) => Promise<void>,
+    ): Promise<void> {
+        await this.started.promise;
+        return next(document);
+    }
+
+    public async didChange(
+        event: vscode.TextDocumentChangeEvent,
+        next: (event: vscode.TextDocumentChangeEvent) => Promise<void>,
+    ): Promise<void> {
+        await this.started.promise;
+        return next(event);
     }
 
     // TODO: Is this used by the magic of "Middleware" in the client library?


SeeminglyScience (approved):
LGTM!

mdaneri:
Thanks @andyleejordan

I tested this and it works almost as expected 👍

If I load the DLL first and then open the file, everything works correctly.
If I open the file first and then load the DLL, the parser continues to show errors.

In the second case, the errors don’t clear automatically, but I found a couple of workarounds:

Make any edit to the file (even a trivial change) → this forces a refresh and fixes the errors
Or:
Close the file
Restart VS Code
Load the DLL
Reopen the file

It looks like the parser/analysis state isn’t being refreshed after the DLL is loaded.

andyleejordan:
@mdaneri That makes sense, but I can't think of a way to detect the DLL being loaded in order to trigger any sort automatic re-parsing.

mdaneri:
@andyleejordan  I agree I don’t think DLL loading is a workable solution. However, the simple closing a tab with the file "in error" and reopening  should force the parser. Unfortunately, it is not working today.

andyleejordan:
Hm, can you take a look? Since you seemed to have it working by re-firing the "on did open" notification, which I thought was going to work the same here.

JustinGrote:
@andyleejordan I think there's something here, I noticed no `didOpen` fires anymore for any document.

image


https://github.com/PowerShell/vscode-powershell/blob/68654f0f114eecfef7bb12c070db70206d8ab78a/src/session.ts#L493-L499

When I breakpoint, it gets to the await on the this.started.promise, but if I step into, it never gets to the return.

didChange and didClose still fire, but didOpen never does.

JustinGrote:
I think you have a `this` binding problem
image

PR #5463, opened 2026-04-09
Fix new suppress terminal setting

andyleejordan:
Behaviorally it was working, but it wasn't showing up in the settings. We also hadn't mapped it into our (pointless IMHO) `Settings` class, or put it under the "correct" section of `integratedConsole`.

To be honest, it's open work to migrate away from `getSettings()` and to just using the VS Code APIs directly, but we shouldn't do it now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1047,7 +1047,7 @@
             "default": false,
             "markdownDescription": "Do not show the startup banner in the PowerShell Extension Terminal."
           },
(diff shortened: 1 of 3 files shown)


andyleejordan:
Updates #5380. Surprised Copilot didn't point out the issue.

SeeminglyScience (approved):

PR #5484, opened 2026-04-30
Add new `additionalPowerShellLocations` option

### LINKED ISSUE #5233: Need a better (cross-platform friendly) way to configure additional PowerShell paths
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.

### Context

- I use PowerShell across operating systems
- I customized additional PowerShell executables for vscode-powershell using `powerShellAdditionalExePaths`:

  ```jsonc
  "powershell.powerShellAdditionalExePaths": {
    "PowerShell 7.5 (macOS)": "/opt/homebrew/bin/pwsh",
    "PowerShell 7 (Windows)": "C:/<path>/<to>/pwsh.exe"
  },
  "powershell.powerShellDefaultVersion": "PowerShell 7 (Windows)",
  ```
- I'm syncing and reusing my vscode settings across operating systems

### The problem

Every time I work with PowerShell files I have to bear the annoying warning of addtional executable I've configured for the other platform not being found.

Image

### Proposed Design

Introduce a new cross-platform/syncing firendly way to customize additional executables

Proposed interface:

```typescript
type OS = "Windows" | "macOS" | "Linux"

interface AddtionalPowerShellLocation {
  name: String
  path: String
  os: OS
  default: Boolean | undefined
}

type Locations = AddtionalPowerShellLocation[]
```

To not break[^1] existing `powerShellAdditionalExePaths`, it'd be a new `additionalPowerShellLocations` key for holding these locations:

```jsonc
"powershell.additionalPowerShellLocations": [
  {
    "name": "PowerShell 7",
    "path": "/opt/powershell/pwsh",
    "os": "macOS",
    "default": true
  },
  {
    "name": "PowerShell 7",
    "path": "C:/PowerShell/stable/pwsh.exe",
    "os": "Windows",
    "default": true
  },
  {
    "name": "PowerShell Nightly",
    "path": "C:/PowerShell/nightly/pwsh.exe",
    "os": "Windows"
  },
  {
    "name": "PowerShell Nightly",
    "path": "/opt/powershell-nightly/pwsh",
    "os": "macOS"
  },
  {
    "name": "PowerShell 7",
    "path": "/opt/powershell7/pwsh",
    "os": "Linux"
  }
]
```

Then `vscode-powershell` can filter out items that are not for the target platform, and launch the first additonal PowerShell executable it successfully resolved. Nothing changes for system fallbacks in this proposal.

`powerShellAdditionalExePaths` and `powerShellDefaultVersion` then could be marked as deprecated.

The proposed design might not be optimal, so please comment.

[^1]: It was a breaking change when `powerShellAdditionalExePaths` was introduced https://github.com/PowerShell/vscode-powershell/issues/4167

chawyehsu:
## PR Summary

Implement and close #5233

This PR attempts to add a better (cross-platform friendly) way to configure additional PowerShell paths.

## Details

This implementaion is a little different from what I've proposed last year in #5233, the primary difference is instead of a `default?: boolean` field, a `weight?: number` is used to rank entries and select the best matching entry. Since we cannot restrict the `default: true` per platform to only one in the VS Code extension configuration, using weighted sorting allows us to achieve a similar purpose. Furthermore, the more specific `platform` field involving `arch` is used to specify the target host instead of `os`.

## Backward Compatibility? Yes

While the old `powerShellAdditionalExePaths` and `powerShellDefaultVersion` stay unchanged and can still be used to define additional PowerShell paths, the new `additionalPowerShellLocations` is added to be mutually exclusive, which means when `additionalPowerShellLocations` is defined(non-empty), old options are ignored. This keeps settings non-conflicting: users pick one path (old or new), not both. Intuitive, simple and easy to understand.

Old options could be marked deprecated after a transition period.

## PR Checklist

- [x] Summarized changes
- [x] PR has tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -968,6 +968,46 @@
               "type": "string"
             }
           },
+          "powershell.additionalPowerShellLocations": {
+            "type": "array",
+            "default": [],
+            "markdownDescription": "Specifies a list of named PowerShell executables tied to a specific platform. When this setting is non-empty, `#powershell.powerShellAdditionalExePaths#` and `#powershell.powerShellDefaultVersion#` are ignored. Entries are ranked by `weight` (highest first), and entries with the same `weight` keep definition order.",
+            "items": {
+              "type": "object",
+              "required": [
+                "name",
+                "path",
+                "platform"
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "name": {
+                  "type": "string",
+                  "markdownDescription": "The display name shown in the Session Menu."
+                },
+                "path": {
+                  "type": "string",
+                  "markdownDescription": "The absolute path to the PowerShell executable or containing folder."
+                },
+                "platform": {
+                  "type": "string",
+                  "enum": [
+                    "windows-x64",
+                    "windows-arm64",
+                    "macos-x64",
+                    "macos-arm64",
+                    "linux-x64",
+                    "linux-aarch64"
+                  ],
+                  "markdownDescription": "The target host platform for this PowerShell location."
+                },
+                "weight": {
+                  "type": "number",
+                  "markdownDescription": "Ranking for this location on its platform. Higher values are preferred; equal values preserve definition order."
+                }
+              }
+            }
+          },
           "powershell.cwd": {
             "type": "string",
             "default": "",
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -34,6 +34,21 @@ export enum OperatingSystem {
     Linux,
 }
 
+export type SupportedPlatform =
+    | "windows-x64"
+    | "windows-arm64"
+    | "macos-x64"
+    | "macos-arm64"
+    | "linux-x64"
+    | "linux-aarch64";
+
+export interface IAdditionalPowerShellLocation {
+    readonly name: string;
+    readonly path: string;
+    readonly platform: SupportedPlatform;
+    readonly weight?: number;
+}
+
 export interface IPlatformDetails {
     operatingSystem: OperatingSystem;
     isOS64Bit: boolean;
@@ -44,6 +59,46 @@ export interface IPowerShellExeDetails {
     readonly displayName: string;
     readonly exePath: string;
     readonly supportsProperArguments: boolean;
+}
+
+export function getSupportedPlatform(
+    platformDetails: IPlatformDetails,
+    processArchitecture: NodeJS.Architecture = process.arch,
+): SupportedPlatform | undefined {
+    switch (platformDetails.operatingSystem) {
+        case OperatingSystem.Windows:
+            switch (processArchitecture) {
+                case "x64":
+                    return "windows-x64";
+                case "arm64":
+                    return "windows-arm64";
+                default:
+                    return undefined;
+            }
+
+        case OperatingSystem.MacOS:
+            switch (processArchitecture) {
+                case "x64":
+                    return "macos-x64";
+                case "arm64":
+                    return "macos-arm64";
+                default:
+                    return undefined;
+            }
+
+        case OperatingSystem.Linux:
+            switch (processArchitecture) {
+                case "x64":
+                    return "linux-x64";
+                case "arm64":
+                    return "linux-aarch64";
+                default:
+                    return undefined;
+            }
+
+        case OperatingSystem.Unknown:
+            return undefined;
+    }
 }
 
 export function getPlatformDetails(): IPlatformDetails {
@@ -91,6 +146,8 @@ export class PowerShellExeFinder {
         // Additional configured PowerShells
         private additionalPowerShellExes: Record<string, string>,
         private logger?: ILogger,
+        private additionalPowerShellLocations: IAdditionalPowerShellLocation[] = [],
+        private processArchitecture: NodeJS.Architecture = process.arch,
     ) {}
 
     /**
@@ -158,7 +215,11 @@ export class PowerShellExeFinder {
 
         // Also show any additionally configured PowerShells
         // These may be duplicates of the default installations, but given a different name.
-        for await (const additionalPwsh of this.enumerateAdditionalPowerShellInstallations()) {
+        const configuredPowerShells =
+            this.additionalPowerShellLocations.length > 0
+                ? this.enumerateAdditionalPowerShellLocations()
+                : this.enumerateAdditionalPowerShellInstallations();
+        for await (const additionalPwsh of configuredPowerShells) {
             if (await additionalPwsh.exists()) {
                 yield additionalPwsh;
             } else if (!additionalPwsh.suppressWarning) {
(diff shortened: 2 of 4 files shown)


chawyehsu:
By any chance can I get some reviews? @JustinGrote @andyleejordan

andyleejordan (commented):
Just stopping by with a decision question: would there be a good way to handle the "default" cases of just windows/macos/linux (without architecture)? My thought is, and this is really an open question: are most people going to want or need to specify the architecture, especially as x64 is kinda on its way out (I mean look at macOS, you'd be hardpressed to find anything but Apple Silicon now). I'm thinking perhaps handle those as defaults, with architecture specific versions as options too.

On the note of these settings getting complicated, do we have a simpler way to handle "weight" that doesn't require the user to specify? I would think the ordered list of options would be implicit weights. Perhaps what this should be is three settings: Windows paths, macOS paths, Linux paths. Infer the priority from the order of paths in the list. Have an optional field on the subproperty (the required field being the path) that is the architecture. Eh?

@JustinGrote could you weigh in on this? Really just spitballing how to make this as clean of a user facing settings interface as possible.

JustinGrote:
@andyleejordan @chawyehsu [terminal profiles separate by OS](https://code.visualstudio.com/docs/terminal/profiles).

I don't like the idea of hardcoding supported platforms. I don't see why this could just be a straight list of paths, and we simply test each path and if it doesn't exist, move on to the next. We would log at warning level paths that weren't found, but we wouldn't bubble up a notification box unless PowerShell couldn't be found at all.


## My thoughts on the best design.
1. If PowerShell Default Version is specified, search AdditionalExePaths that match that name, then autodiscover for that particular name, and if all that fails, typical discovery path. Possibly have a "strict" toggle that will show an error if that specific PowerShell Name cannot be found.

So for example, this config:

Name              | Value
PowerShell        | C:\MyCustomPs
PowerShell        | /home/my/customps
PowerShellPreview | /home/my/preps

If nothing is specified for PowerShell default version, search all 3 of these paths first, in order, then autodiscover.

If 'PowerShell' is specified for default version:
1. Search the first two paths first
2. Autodiscover for "PowerShell" name (as opposed to winps or store)
3. If the new "Strict PowerShell Default Version" is checked, throw an error saying not found at this point.
4. If it is not checked, try remaining additionalexepaths in order, then autodiscover other powershells

We could do some basic path exclusions right off the bat based on Windows/NotWindows path styles, that way you can have a single consolidated list of all the paths you want to try that would be syncable cross platform.

This doesn't require two different config paths and keeps things simple.

@chawyehsu is there a scenario that the above would not address for you? Overrides can be done via profiles or remote development machine specific overrides like if you want preview to be default in a particular container, etc.

andyleejordan:
> I don't like the idea of hardcoding supported platforms. I don't see why this could just be a straight list of paths, and we simply test each path and if it doesn't exist, move on to the next. We would log at warning level paths that weren't found, but we wouldn't bubble up a notification box unless PowerShell couldn't be found at all.

To my knowledge this is exactly how it works now (with the warnings able to be turned off because they're noisy).

JustinGrote:
@andyleejordan I'll have to dive in but the nuance of how I mentioned how the priorities should work along with the "strict" checkbox is the net-new stuff, but the core UX would mostly stay the same.

chawyehsu:
> My thought is, and this is really an open question: are most people going to want or need to specify the architecture, especially as x64 is kinda on its way out (I mean look at macOS, you'd be hardpressed to find anything but Apple Silicon now). I'm thinking perhaps handle those as defaults, with architecture specific versions as options too.

@andyleejordan When I first opened this PR, I considered only checking the OS without taking the architecture into account. For example, on Windows x64 you can still use the x86 version of PowerShell, and macOS is moving toward ARM-only anyway. However, I later changed the implementation to also check the architecture, mainly because of Linux: on Linux aarch64, you can't use the x64 build. The current `SupportedPlatform` logic is therefore based entirely on PowerShell's official support matrix. This is the closest match to PowerShell's release model that I could come up with. If there's a simpler approach, I'd be happy to go with it.

> do we have a simpler way to handle "weight" that doesn't require the user to specify? I would think the ordered list of options would be implicit weights. Perhaps what this should be is three settings: Windows paths, macOS paths, Linux paths.

I did consider this method, but the merits of such implicit conventions are entirely subjective. One issue I find hard to agree with in this priority-by-ordering approach is that changing the default PowerShell via the bottom-right selector would end up reordering the entire configuration, causing much larger diffs in the settings file. In contrast, with the `weight` approach, the diff is limited to changes in the weight values only. And if entries for all platforms are mixed into a single ordering, it would become extremely messy.

> [terminal profiles separate by OS](https://code.visualstudio.com/docs/terminal/profiles).

@JustinGrote I believe this link is irrelevant to the issue this PR seeks to address. Terminal profiles only affect the shell of the VS Code built-in terminal, while this discussion concerns the content within vscode-powershell that affects PowerShell selection and thus controls editor services such as LSP, if I understand correctly...

> Name | Value
> PowerShell | C:\MyCustomPs
> PowerShell | /home/my/customps
> PowerShellPreview | /home/my/preps
> 
> If nothing is specified for PowerShell default version, search all 3 of these paths first, in order, then autodiscover.

This is actually the existing configuration approach. However, it does not solve the issue mentioned in #5233. In case it's unclear, the problem in #5233 can actually be divided into two parts.

The first is the annoying warning, which can be resolved with the setting proposed by `powershell.suppressAdditionalExeNotFoundWarning` (thanks for pointing that out). The second is that the current `powerShellAdditionalExePaths` + `powerShellDefaultVersion` configuration model cannot specify different default PowerShell versions for different operating systems.

Here’s the scenario:

You can only set a single default version through `powerShellDefaultVersion`, and that value applies to all operating systems. For example, you cannot set `PowerShell (Custom on Windows)` as the default on Windows while using `PowerShell (Custom on macOS)` on macOS. The only option is to use something like `PowerShell (Custom)` as the universal default. But then you can no longer distinguish platforms by name, and you end up having to configure `powerShellAdditionalExePaths` like this:



```json
"powershell.powerShellAdditionalExePaths": {
  "PowerShell (Custom)": "/opt/path/to/linux/bin/pwsh",
  "PowerShell (Custom)": "/opt/path/to/macos/bin/pwsh",
  "PowerShell (Custom)": "D:/PowerShell-custom/pwsh.exe"
}
```
_(similar to what you mentioned above)_

At first glance, this seems like it should work, i.e. VS Code would discover and use a different `PowerShell (Custom)` executable on each operating system. But it does not. The reason is that `powerShellAdditionalExePaths` is parsed as a hashmap. When multiple entries share the same key `PowerShell (Custom)`, only the last one is retained and the others are overwritten.



As a result, on macOS/Linux, vscode-powershell ends up trying to discover `D:/PowerShell-custom/pwsh.exe`, fails, and then reports the following error:



I don't think you can solve this problem without getting rid of the current `powerShellDefaultVersion` configuration.

chawyehsu:
@JustinGrote @andyleejordan Any further thoughts on the situation?

JustinGrote (commented):
Sorry for the delay. 

Correct, my thought would be to allow that setting to be parsed not as unique keys but instead more like an array, and correlate the results, so you could effectively specify precedence for the same item.

chawyehsu:
Ok, so now we can introduce a simplified version:

```ts
interface AddtionalPowerShellLocation {
  name: String
  path: String
}

type Locations = AddtionalPowerShellLocation[]
```

```json
"powershell.additionalPowerShellLocations": [
  {
    "name": "PowerShell 7 (Linux)",
    "path": "/opt/powershell/pwsh"
  },
  {
    "name": "PowerShell 7",
    "path": "C:/PowerShell/stable/pwsh.exe"
  },
  {
    "name": "PowerShell Nightly",
    "path": "C:/PowerShell/nightly/pwsh.exe"
  },
  {
    "name": "PowerShell Nightly (macOS)",
    "path": "/opt/powershell-nightly/pwsh"
  },
  {
    "name": "PowerShell 7",
    "path": "/opt/powershell7/pwsh"
  },
]
```

1. Based on the interfaces above, the resolver will evaluate candidates in order. Values appearing earlier have higher priority. This also comes with an implicit "default value" semantic: if a value is intended to be the default, it should be defined first. However, this is not guaranteed, as it remains the user's responsibility to ensure that the value is valid.

2. Two filters will be applied to the candidates:
   - **filter on path**: If the path contains `.exe` suffix and is on a non-Windows platform, it will be filtered out in advance. This is straightforward, to reduce file I/O of testing path.
   - **filter on name**: If the name contains OS strings (e.g., "Windows", "Linux", "macOS", case-insensitive) that do not match the current platform, it will be filtered out in advance. This gives users the ability to set values ​​according to the platform, while abandoning hard-coded platform settings.

If we all agree with the above interface, then I will update the implementation.

JustinGrote:
Looks reasonable to me, except instead of strings in name (which is fragile) I'd prefer an *optional* OS property that ties to an enum, and the filter can occur that way.

I would get @andyleejordan or @SeeminglyScience opinion, as tho I'm a contributor, I don't have ability to directly approve PRs.

chawyehsu:
> I'd prefer an optional OS property that ties to an enum, and the filter can occur that way.

This is based on your earlier comment about not wanting to hardcode os platform values. It's feasible to use optional os field.

While filtering by name is admittedly somewhat fragile, I don't think it has a significant downside in practice. It keeps the design simpler and aligns with the current discussed model.

I would like to get feedback from @andyleejordan and @SeeminglyScience. I've been using PowerShell with custom paths on multiple platforms, and really frustrated by the current settings with cross-platform issues. Anyways @JustinGrote Thanks for reviews!

Issue #5031, opened 2024-08-21
PowerShell extension doesn't start on slow machines


SetTrend:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [_open and closed_ issues](https://github.com/PowerShell/vscode-powershell/issues?q=is%3Aissue) to ensure it has not already been reported.
- [X] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [X] I am sure this issue is with the _extension itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance.
- [X] I have verified that I am using the latest version of Visual Studio Code and the PowerShell extension.
- [X] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

Whenever I start VS Code for the first time in a Windows session, PowerShell extension starts with a warning.

Moreover, closing the terminal session results in the below mentioned error.

Plus, the PowerShell extension banner message is looking odd:

![PowerShell extension banner](https://github.com/user-attachments/assets/c2002d3d-f764-4690-9fa9-ded2557c877a)

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.4.4
PSEdition                      Core
GitCommitId                    7.4.4
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0.}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visual Studio Code Version

```console
1.92.1
x64
```

### Extension Version

```console
ms-vscode.powershell@2024.2.2
```

### Steps to Reproduce

1. Launch a Windows session on a slow machine
2. Right-click a PowerShell file and select "Open in Visual Studio Code" context menu item.
3. Immediately open the terminal window using the keyboard shortcut <CTRL>+<Ö>.

### Logs

```
21.8.2024 16:17:03 [NORMAL] - Visual Studio Code: v1.92.1 64-bit on Windows 64-bit

21.8.2024 16:17:03 [NORMAL] - PowerShell Extension: v2024.2.2

[...]

PR #5523, opened 2026-06-16
Handle version-details timeout and stalled connects during startup

### LINKED ISSUE #5521: Startup version-details timeout throws instead of failing gracefully; `connectFunc` has no connect timeout
## Summary

While investigating #5031, two latent startup-path bugs in `src/session.ts` surfaced. Neither caused #5031, but both are real gaps in how the extension handles a stalled/slow PSES connection.

## 1. `getVersionDetails()` timeout throws out of `start()`

`getVersionDetails()` (around `src/session.ts:1073-1077`) sets a 60s timeout that only cancels the token:

```ts
const timeout = new vscode.CancellationTokenSource();
setTimeout(() => timeout.cancel(), 60 * 1000);
const versionDetails = await this.languageClient?.sendRequest(
    PowerShellVersionRequestType,
    timeout.token,
);
```

When the token is cancelled, `sendRequest(...)` **rejects**, so the `await` throws straight out of `start()`. The intended "Unable to get version details!" handling (around `src/session.ts:281-285`) is therefore unreachable — the failure surfaces as an unhandled rejection / generic error instead of the dedicated message.

## 2. `connectFunc` has no error/timeout handler

The socket/pipe connect logic in `connectFunc` (around `src/session.ts:851-861`) has no error or timeout handler. If the transport never connects, there's nothing to time it out or report a useful failure.

## Suggested direction

- Wrap the version-details request so a timeout produces the intended "Unable to get version details!" failure path rather than throwing.
- Add an error/timeout handler to `connectFunc` so a stalled connect fails with an actionable message.

## Context

Found during the #5031 investigation (fast-process-exit, not a slow start). Filing separately so it's tracked on its own; the exit-code diagnostics improvement for #5031 is being handled in its own change.

<sub>🤖 Drafted by GitHub Copilot at @andyleejordan's request.</sub>

andyleejordan:
## Summary

Fixes the two latent startup-path bugs in `src/session.ts` reported in #5521 (surfaced while investigating #5031; neither caused it).

### 1. `getVersionDetails()` timeout threw out of `start()`

The 60s timeout only cancelled the token, so when it fired `sendRequest(...)` rejected and the `await` threw straight out of `start()` — the intended "Unable to get version details!" handling was never reached. It now catches the rejection, logs it via `ILogger`, and returns `undefined` so the existing `if (versionDetails === undefined)` failure branch runs. Also fixes the original timer leak: the `setTimeout` handle is captured and `clearTimeout`'d, and the `CancellationTokenSource` is `dispose()`d in `finally`.

### 2. `connectFunc` had no error/timeout handler

A transport that never connected just hung with nothing actionable. Adds an `error` handler and a 60s connect timeout that `destroy()`s the socket and rejects with a clear, logged message naming the pipe. Both paths `clearTimeout` so a healthy but idle connection is never torn down. Works whether the transport is a named pipe or a socket.

## Tests

Adds `describe("SessionManager.getVersionDetails")` in `test/core/session.test.ts` covering the reject-to-`undefined` path and the normal fast path. `connectFunc` is a private closure over a live `net` socket with no test seam, so it's left uncovered rather than adding socket-mocking infra the repo deliberately avoids.

## Verification

- `npm run lint`, `npm run format`, `npm run compile` (esbuild) all pass.
- `npx tsc --noEmit` shows only the known pre-existing, unrelated `TelemetryReporter` namespace error at `session.ts:136`.
- The `modules/` PSES symlink wasn't present in the dev worktree, so the `@vscode/test-cli` integration harness wasn't run locally; the new tests are unit-style and mirror the existing pattern already in `test/core/session.test.ts`.

Closes #5521.

🤖 Drafted by GitHub Copilot (Claude Opus 4.8) at @andyleejordan's request.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -280,7 +280,7 @@ export class SessionManager implements Middleware {
             this.versionDetails = await this.getVersionDetails();
             if (this.versionDetails === undefined) {
                 void this.setSessionFailedOpenBug(
-                    "Unable to get version details!",
+                    "PowerShell started but didn't report its version in time, so the session couldn't finish starting up!",
                 );
                 return;
             }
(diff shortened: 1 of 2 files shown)


JustinGrote (approved):
Tests pass on my machine. not sure if anyone ever hit this deadlock but a good find.

PR #5525, opened 2026-06-16
Add `interactive-extension-testing` skill for the dev workspace

andyleejordan:
Adds a project skill documenting how to set up the multi-root `pwsh-extension-dev` VS Code workspace for interactive (F5) testing of the extension against a locally-built PowerShellEditorServices, including the symlink setup a worktree or fresh checkout needs.

It references existing docs (`.github/copilot-instructions.md`, `pwsh-extension-dev.code-workspace`, and the relevant tasks in `vscode-powershell.build.ps1`) as the single sources of truth rather than duplicating them.

Docs-only; no source changes.

### DIFF
--- a/.github/skills/interactive-extension-testing/SKILL.md
+++ b/.github/skills/interactive-extension-testing/SKILL.md
@@ -1,0 +1,88 @@
+---
+name: interactive-extension-testing
+description: >
+    Guide for setting up the multi-root PowerShell VS Code extension/PSES dev
+    workspace. Use when asked to open the dev workspace, interactively test the
+    extension, set up PSES for F5 debugging, or open pwsh-extension-dev workspace.
+---
+
+# Interactive Extension Testing
+
+Read [.github/copilot-instructions.md](../../../.github/copilot-instructions.md)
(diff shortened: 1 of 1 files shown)


JustinGrote (approved):
@andyleejordan
Low risk so I am approving, but does this actually work? I don't think copilot can directly interact or drive a vscode window so how will it test once it is open?

JustinGrote:
@andyleejordan you'll need @SeeminglyScience to approve this one since it's not directly copilot authored.

andyleejordan:
@justingrote it does work! This is in the new GitHub Copilot desktop app. I had to have it figure out how to map two concurrent Git worktrees into a build so I could do the manual dev-test loop, but it did and it was great.

JustinGrote:
@andyleejordan so does it just get you to the point of manually testing it, or can it actually drive the UI to verify by itself? That's wild if the latter.

andyleejordan:
It's the former, but it was tricky since it's not working out of my local checkout, so I can't just open the workspace in Code and have it work. But I can ask it to open it for me. I think Claude Code can do the latter but I am not authorized to use that...hopefully the Copilot team gets to it.

